Issue #84, opened 2020-01-27
Specify Grid Model in Scenario List


rouille:
- [x] Add a new `grid model` column in the ***ScenarioList*** file
- [x] Test changes in the container

PR #337, opened 2020-11-18
Specify grid model when creating scenario

rouille:
### Purpose
Set grid model during the building phase of a scenario (#84)

### What the code is doing
A `set_grid_model` method in the `Create` class allows to set the grid model. If not called the `grid_model` key of the `scenario_info` dictionary is set to `usa_tamu`. When the `create scenario` method of the `Create` class is called the `grid_model` field (located between `engine` and `runtime`) in the ***ScenarioList** file will be set.

### Testing
We will create/launch/extract a dummy scenario

### Where to look
* The `powersimdata.scenario.create` module where the `set_grid_model` method can be called and the `scenario_info` dictionary is filled.
* The `powersimdata.scenario.execute` module where the `Grid` class is instantiated using the `grid_model` key of the `scenario_info` dictionary.

### Time estimate
15min. It mirrors what we did to set the engine

### Note
We will have before testing to add a new `grid_model` column to the ***ScenarioList** file and set `usa_tamu` as value for all existing scenario

### DIFF
--- a/powersimdata/input/grid.py
+++ b/powersimdata/input/grid.py
@@ -11,7 +11,9 @@ class Grid(object):
     """Grid
 
     :param str/list interconnect: interconnect name(s).
-    :param str source: model used to build the network.
+    :param str source: should be either the name of a grid model within the
+        '*powersimdata/network*' directory or a filename for a '*.mat*' file, which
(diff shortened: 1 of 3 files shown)


rouille:
I close this since we will tackle this in #344

PR #374, opened 2021-01-21
test: remove index column from test assertion

jon-hagg:
### Purpose
Fix failing integration test - id column is now an index, so isn't in the `.columns` attribute. Code changes were introduced in #365

### What the code is doing
Updated the expected list of column headers

### Testing
Ran `pytest -m "not db"` locally

### Time estimate
2 min


rouille:
Should we test that `execute_table.index.name` and `scenario_table.index.name` equal `id`?

rouille (approved):
Good catch

jon-hagg:
Found via @dmuldrew's integration test branch

PR #375, opened 2021-01-22
fix: change caught Exception for get_load_shed

danielolsen:
### Purpose
Restore previous functionality when calling `scenario.state.get_load_shed()` on a scenario with no load shed.

### What the code is doing
Change which Exception triggers the construction of an empty load_shed data frame.

### Testing
Can be verified by calling `Scenario(1861).state.get_load_shed()` before and after.

### Time estimate
2 minutes.

### DIFF
--- a/powersimdata/scenario/analyze.py
+++ b/powersimdata/scenario/analyze.py
@@ -216,7 +216,8 @@ class Analyze(State):
             load_shed = construct_load_shed(self._scenario_info, grid, infeasibilities)
 
             filename = scenario_id + "_LOAD_SHED.pkl"
-            filepath = os.path.join(server_setup.LOCAL_DIR, filename)
+            output_dir = server_setup.OUTPUT_DIR
+            filepath = os.path.join(server_setup.LOCAL_DIR, output_dir, filename)
             with open(filepath, "wb") as f:
                 pickle.dump(load_shed, f)
 
--- a/powersimdata/scenario/analyze.py
+++ b/powersimdata/scenario/analyze.py
@@ -209,7 +209,7 @@ class Analyze(State):
             # It's either on the server or in our local ScenarioData folder
             output_data = OutputData(data_loc=self.data_loc)
             load_shed = output_data.get_data(scenario_id, "LOAD_SHED")
-        except FileNotFoundError:
+        except OSError:
             # The scenario was run without load_shed, and we must construct it
             grid = self.get_grid()
             infeasibilities = self._parse_infeasibilities()


powersimdata/scenario/analyze.py
```diff
@@ -209,7 +209,7 @@ def get_load_shed(self):
             # It's either on the server or in our local ScenarioData folder
             output_data = OutputData(data_loc=self.data_loc)
             load_shed = output_data.get_data(scenario_id, "LOAD_SHED")
-        except FileNotFoundError:
+        except OSError:
```
rouille, line 212: Is this the error raised by the `check_file_exists` method of the `DataAccess` class?

danielolsen: Exactly.

rouille (approved):
Thanks

danielolsen:
One more change: we want to store the `load_shed.pkl` files that we create in the **output/** folder now, not the root of the local dir.

powersimdata/scenario/analyze.py
```diff
@@ -209,14 +209,15 @@ def get_load_shed(self):
...
             # The scenario was run without load_shed, and we must construct it
             grid = self.get_grid()
             infeasibilities = self._parse_infeasibilities()
             load_shed = construct_load_shed(self._scenario_info, grid, infeasibilities)
 
             filename = scenario_id + "_LOAD_SHED.pkl"
-            filepath = os.path.join(server_setup.LOCAL_DIR, filename)
+            folder = server_setup.OUTPUT_DIR
+            filepath = os.path.join(server_setup.LOCAL_DIR, folder, filename)
```
rouille, line 220: Indeed.

Can we rename `folder` to `output_dir` or do `filepath = os.path.join(server_setup.LOCAL_DIR, server_setup.OUTPUT_DIR, filename)`

BainanXia:
@danielolsen Could you remind me why we decide to create an all-zero file if there is no load shedding in a scenario when call `get_load_shed()`?

danielolsen:
> 
> 
> @danielolsen Could you remind me why we decide to create an all-zero file if there is no load shedding in a scenario when call `get_load_shed()`?

@BainanXia the thought was that if the `construct_load_shed` function takes a while, and we expect to be trying access the `load_shed` repeatedly, then it is easier to cache the result and unpickle it rather than reconstruct it every time. But I could also see the argument of saving space. Maybe we get the best of both worlds by having an optional flag `cache_constructed` which defaults to false but saves the files if true?

That reminds me, I noticed the other day that I have written three different functions to get the bus demand (part of construct_load_shed), in three different ways!
- `powersimdata.input.input_data.get_bus_demand`
- `powersimdata.scenario.helpers.calculate_bus_demand`
- `postreise.analyze.transmission.congestion.map_demand_to_buses`

Please no one let me write any more! I had intended to fix this eventually by picking the fastest method, standardizing to that one in powersimdata, and importing it from PostREISE, but it slipped behind all the other priorities, maybe now is the time.

BainanXia:
> > @danielolsen Could you remind me why we decide to create an all-zero file if there is no load shedding in a scenario when call `get_load_shed()`?
> 
> @BainanXia the thought was that if the `construct_load_shed` function takes a while, and we expect to be trying access the `load_shed` repeatedly, then it is easier to cache the result and unpickle it rather than reconstruct it every time. But I could also see the argument of saving space. Maybe we get the best of both worlds by having an optional flag `cache_constructed` which defaults to false but saves the files if true?
> 
> That reminds me, I noticed the other day that I have written three different functions to get the bus demand (part of construct_load_shed), in three different ways!
> 
> * `powersimdata.input.input_data.get_bus_demand`
> * `powersimdata.scenario.helpers.calculate_bus_demand`
> * `postreise.analyze.transmission.congestion.map_demand_to_buses`
> 
> Please no one let me write any more! I had intended to fix this eventually by picking the fastest method, standardizing to that one in powersimdata, and importing it from PostREISE, but it slipped behind all the other priorities, maybe now is the time.

That's another point that I haven't thought of. Thanks for bringing that up. I understand the idea of saving the load shedding info in a `pkl` file for faster + repetitive access, however, unlike storage, if we do `scenario.state.get_stoarge_pg()` with a no-storage scenario, it simply pops the error of not finding the file (we can definitely catch this exception and print a more user friendly message), then why for load_shed, we would like to create an empty file for it?

rouille:
> > @danielolsen Could you remind me why we decide to create an all-zero file if there is no load shedding in a scenario when call `get_load_shed()`?
> 
> @BainanXia the thought was that if the `construct_load_shed` function takes a while, and we expect to be trying access the `load_shed` repeatedly, then it is easier to cache the result and unpickle it rather than reconstruct it every time. But I could also see the argument of saving space. Maybe we get the best of both worlds by having an optional flag `cache_constructed` which defaults to false but saves the files if true?
> 
> That reminds me, I noticed the other day that I have written three different functions to get the bus demand (part of construct_load_shed), in three different ways!
> 
>     * `powersimdata.input.input_data.get_bus_demand`
> 
>     * `powersimdata.scenario.helpers.calculate_bus_demand`
> 
>     * `postreise.analyze.transmission.congestion.map_demand_to_buses`
> 
> 
> Please no one let me write any more! I had intended to fix this eventually by picking the fastest method, standardizing to that one in powersimdata, and importing it from PostREISE, but it slipped behind all the other priorities, maybe now is the time.

That would be nice if you can do that, i.e., select the fastest method and import the function everywhere it is needed. Then depending on how fast the method is we can either just return the data frame or return the data frame and save the file with the keyword you mentioned.

jon-hagg (approved):

BainanXia:
> > > @danielolsen Could you remind me why we decide to create an all-zero file if there is no load shedding in a scenario when call `get_load_shed()`?
> > 
> > 
> > @BainanXia the thought was that if the `construct_load_shed` function takes a while, and we expect to be trying access the `load_shed` repeatedly, then it is easier to cache the result and unpickle it rather than reconstruct it every time. But I could also see the argument of saving space. Maybe we get the best of both worlds by having an optional flag `cache_constructed` which defaults to false but saves the files if true?
> > That reminds me, I noticed the other day that I have written three different functions to get the bus demand (part of construct_load_shed), in three different ways!
> > ```
> > * `powersimdata.input.input_data.get_bus_demand`
> > 
> > * `powersimdata.scenario.helpers.calculate_bus_demand`
> > 
> > * `postreise.analyze.transmission.congestion.map_demand_to_buses`
> > ```
> > 
> > 
> > Please no one let me write any more! I had intended to fix this eventually by picking the fastest method, standardizing to that one in powersimdata, and importing it from PostREISE, but it slipped behind all the other priorities, maybe now is the time.
> 
> That would be nice if you can do that, i.e., select the fastest method and import the function everywhere it is needed. Then depending on how fast the method is we can either just return the data frame or return the data frame and save the file with the keyword you mentioned.

Ah, this answers my question as well. Now I remember why we were doing this. Unlike `storage_pg`, which is stored/extracted directly in `REISE.jl`, we need to construct `load_shed` first before we can get the info. Then I'm wondering whether we could check if `infeasibilities` is None before `construct_load_shed`, i.e. if there is no infeasibilities, we just print out the message instead of creating a all-zero `pkl` file.

danielolsen:
> Ah, this answers my question as well. Now I remember why we were doing this. Unlike `storage_pg`, which is stored/extracted directly in `REISE.jl`, we need to construct `load_shed` first before we can get the info. Then I'm wondering whether we could check if `infeasibilities` is None before `construct_load_shed`, i.e. if there is no infeasibilities, we just print out the message instead of creating an all-zero `pkl` file.

More of my original thinking:
- A given scenario can be known to either have storage or not at before trying to access the results. We could also return an empty DataFrame here if we wanted (rather than an Exception), but the construction of a blank dataframe is very fast already so there's no need to cache this.
- By contrast, a given scenario cannot be known to have load shedding or not until it is run. So we may want to be able to call `get_load_shed` on every scenario, as part of automated checking, and in this case I figured it was best if the return type was always a DataFrame, rather than a DataFrame or None, since anything automated would probably need extra logic to parse the None into something useful (e.g. not call `load_shed.sum()` when `load_shed` is `None`).

BainanXia:
> > Ah, this answers my question as well. Now I remember why we were doing this. Unlike `storage_pg`, which is stored/extracted directly in `REISE.jl`, we need to construct `load_shed` first before we can get the info. Then I'm wondering whether we could check if `infeasibilities` is None before `construct_load_shed`, i.e. if there is no infeasibilities, we just print out the message instead of creating an all-zero `pkl` file.
> 
> More of my original thinking:
> 
> * A given scenario can be known to either have storage or not at before trying to access the results. We could also return an empty DataFrame here if we wanted (rather than an Exception), but the construction of a blank dataframe is very fast already so there's no need to cache this.
> * By contrast, a given scenario cannot be known to have load shedding or not until it is run. So we may want to be able to call `get_load_shed` on every scenario, as part of automated checking, and in this case I figured it was best if the return type was always a DataFrame, rather than a DataFrame or None, since anything automated would probably need extra logic to parse the None into something useful (e.g. not call `load_shed.sum()` when `load_shed` is `None`). But I could be convinced otherwise.

Let's keep it in the current form then.

BainanXia (approved):
This is good to go.

PR #376, opened 2021-01-22
DataAccess implementation for shared volume architecture

jon-hagg:
### Purpose
Add remaining pieces to enable expected functionality within the containerized setup (or any setup where reise.jl and powersimdata have access to a shared data volume). 

### What the code is doing
* Implemented `LocalDataAccess` subclass - logic is simplified via the symlink in the dockerfile
* Use the factory in `context.py` to create the correct instance
* Add template for scenario/execute list in docker image
* Fixes for issues found during test workflow

### Testing
Built the containers and run with docker-compose. Copy-pasted the instructions from the readme for launching a simulation. ~~Everything succeeds, up til the point where we invoke gurobi - so there is more testing to be done, but I think it's at a reasonable point, and shouldn't expect major changes here.~~

Update: I'm able to launch a simulation now 

### Time estimate
20 min


powersimdata/data_access/context.py
```diff
@@ -8,7 +12,7 @@ class Context:
     @staticmethod
     def get_data_access(data_loc=None):
         """Return a data access instance appropriate for the current
-        environment. Currently only supports internal client server mode.
+        environment.
 
         :param str data_loc: pass "disk" if using for backups otherwise leave
```
rouille, line 17: should it read `pass "disk" if using backups otherwise ...`

powersimdata/scenario/execute.py
```diff
@@ -290,7 +291,7 @@ def create_folder(self):
         :raises IOError: if folder cannot be created.
         """
         print("--> Creating temporary folder on server for simulation inputs")
-        command = "mkdir %s" % self.TMP_DIR
+        command = "mkdir -p %s" % self.TMP_DIR
```
rouille, line 294: Just curious, why do you need the `-p` option here

jon-hagg: Couple reasons - create missing parent directories, and makes it re-runnable. Initially I got an error during `prepare_simulation_input` due to the `tmp` folder not existing yet. So this fixes that, and has the nice side effect of not failing when called twice.

powersimdata/data_access/context.py
```diff
@@ -18,4 +22,8 @@ def get_data_access(data_loc=None):
             root = server_setup.BACKUP_DATA_ROOT_DIR
         else:
             root = server_setup.DATA_ROOT_DIR
-        return SSHDataAccess(root)
+
+        mode = get_deployment_mode()
+        if mode == DeploymentMode.Server:
+            return SSHDataAccess(root)
+        return LocalDataAccess(root)
```
rouille, line 29: Just being curious, what is the advantage of having a class with 2 static attributes over a dictionary (`deployment_mode = {"Server": "SERVER", "Container": "CONTAINER"}`)?

powersimdata/data_access/context.py
jon-hagg: Should it? I wasn't sure on the wording, can change if that's more accurate

powersimdata/data_access/context.py
jon-hagg: Not sure there is an advantage either way. I think the class works better with tab completion and seems slightly more readable. Another option would be an enum.

powersimdata/data_access/context.py
rouille: Not sure, you are the native speaker! What about: pass "disk" if data are located on a mounted disk whose location is in specified in :mod:\`powersimdata.utility.server_setup\` ...

jon-hagg: Haha but you are more familiar with the backup feature :) The latest wording sounds good though

powersimdata/scenario/execute.py
dmuldrew: Yep, I had to deal with errors due to the missing `tmp` folder in my ssh server build as well

powersimdata/data_access/data_access.py
```diff
@@ -93,6 +93,57 @@ def close(self):
...
+
+        proc = Popen(
+            command,
+            shell=True,
+            executable="/bin/bash",
+            stdout=PIPE,
+            stderr=PIPE,
+            text=True,
+        )
+        return wrap(None), wrap(proc.stdout), wrap(proc.stderr)
```
rouille, line 144: Not sure I understand the need for the first element of the returned tuple. Do you redirect the standard streams to dev/null?

powersimdata/data_access/data_access.py
```diff
@@ -93,6 +93,57 @@ def close(self):
...
+    def __init__(self, root=None):
+        self.root = root if root else server_setup.DATA_ROOT_DIR
+
+    def copy_from(self, file_name, from_dir):
+        """Copy a file from data store to userspace.
+
+        :param str file_name: file name to copy.
+        :param str from_dir: data store directory to copy file from.
+        """
+        pass
```
rouille, line 108: Do we need to overwrite the method here. It raises a `NotImplementedError` in the parent class, is that problematic?

powersimdata/data_access/data_access.py
```diff
@@ -93,6 +93,57 @@ def close(self):
...
+        :param str to_dir: data store directory to copy file to.
+        :param str change_name_to: new name for file when copied to data store.
+        """
+        self.check_filename(file_name)
+        src = posixpath.join(server_setup.LOCAL_DIR, file_name)
+        file_name = file_name if change_name_to is None else change_name_to
+        dest = posixpath.join(self.root, to_dir, file_name)
+        self.check_file_exists(dest, should_exist=False)
+        self.copy(src, dest)
+        self.remove(src)
```
rouille, line 123: Should it be called `move_to`. I guess this would apply to `DartaAccess` and `SSDataAccess`

powersimdata/data_access/data_access.py
jon-hagg: The implementation in `SSHDataAccess` uses `exec_command` from paramiko which returns the standard stream tuple with 3 elements - we have to match that here. We don't ever use the stdin component so I omitted that. The `wrap` is mostly for the `stderr` part which, when successful, will be `None` and cause an error when we call `.readlines()` - compared to paramiko which gives an empty stream. To fix that we just read from `/dev/null` here.

powersimdata/data_access/data_access.py
jon-hagg: Yeah, that makes more sense.

powersimdata/data_access/data_access.py
jon-hagg: This does in fact override the behavior - in this case to do nothing, since the directories are symlinked already.

powersimdata/data_access/data_access.py
rouille: You can keep `copy_to` without the remove if we need it in the future and define `move_to` that will call `copy_to` + the `remove` method

jon-hagg: We also call `remove` in the ssh implementation, so we only need `move_to`, right? In which case the change would just be renaming all the usages of `copy_to`.

rouille: Yeah, right now we only move files.

rouille (approved):
Looks good. Thanks.

PR #378, opened 2021-01-25
Update scenario status

rouille:
### Purpose
Update scenario status before launching a simulation.

### What the code is doing
Update the `_scenario_status` attribute of the `Execute`.

### Where to look
The `_update_scenario_status` method is invoked in the `_check_if_ready` method which, in turn, is invoked in the `launch_simulation` method

### Time estimate
2 min

### DIFF
--- a/powersimdata/scenario/execute.py
+++ b/powersimdata/scenario/execute.py
@@ -169,6 +169,7 @@ class Execute(State):
 
         :raises ValueError: if status is invalid
         """
+        self._update_scenario_status()
         valid_status = ["prepared", "failed", "finished"]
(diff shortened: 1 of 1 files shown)


danielolsen:
Looks reasonable, I'll test with the next scenario I create and report back.

jon-hagg (approved):
I ended up testing the same fix in #376

BainanXia (approved):
It works! I tested it with scenario 1995 just now. Without `print_scenario_status`, we can still launch scenario now.

danielolsen (approved):
Also works with scenario number 1996 :)

PR #382, opened 2021-01-28
Load and transform US normalized profiles

rouille:
### Purpose
Use the newly constructed profiles:
* ***demand_vJan2021.csv***
* ***hydro_vJan2021.csv***
* ***solar_vJan2021.csv***
* ***wind_vJan2021.csv***
The power output in these profiles are now normalized (independent of the `Pmax` of the generator) and enclose, for each fuel type, all the generator in the US grid.

### What the code is doing
* Modify the `InputData` class to fetch the new profile located in **raw/GRID_MODEL*** where ***GRID_MODEL*** indicates the grid model, e.g., `usa_tamu`.
* Add a new dictionary in the `powersimdata.network.usa_tamu.constants.zones` to map interconnect to load zone id
* Refactor the `TransformProfile` class and the associated tests

### Testing
run `pytest -v powersimdata/input/tests/test_transform_profile.py`. All the new profiles will be downloaded from the server the first time you run the tests. 

### Where to look
The core of the changes is located in `powersimdata/input/transform_profile`

### Time estimate
30min

### DIFF
--- a/powersimdata/input/input_data.py
+++ b/powersimdata/input/input_data.py
@@ -1,9 +1,9 @@
 import os
+import posixpath
 
 import pandas as pd
 
 from powersimdata.data_access.context import Context
-from powersimdata.scenario.helpers import interconnect2name
 from powersimdata.utility import server_setup
 from powersimdata.utility.helpers import MemoryCache, cache_key
 
@@ -59,10 +59,11 @@ class InputData(object):
         ext = self.file_extension[field_name]
 
         if field_name in ["demand", "hydro", "solar", "wind"]:
-            interconnect = interconnect2name(scenario_info["interconnect"].split("_"))
             version = scenario_info["base_" + field_name]
-            file_name = interconnect + "_" + field_name + "_" + version + "." + ext
-            from_dir = server_setup.BASE_PROFILE_DIR
+            file_name = field_name + "_" + version + "." + ext
+            from_dir = posixpath.join(
+                server_setup.BASE_PROFILE_DIR, scenario_info["grid_model"]
+            )
         else:
             file_name = scenario_info["id"] + "_" + field_name + "." + ext
             from_dir = server_setup.INPUT_DIR
--- a/powersimdata/input/tests/test_transform_profile.py
+++ b/powersimdata/input/tests/test_transform_profile.py
@@ -1,5 +1,6 @@
 import numpy as np
 import pytest
+from numpy.testing import assert_almost_equal
 
 from powersimdata.input.change_table import ChangeTable
 from powersimdata.input.grid import Grid
@@ -9,32 +10,34 @@ from powersimdata.input.transform_profile import TransformProfile
 
 interconnect = ["Western"]
 param = {
-    "Western": {
-        "demand": "v4",
-        "hydro": "v2",
-        "solar": "v4.1",
-        "wind": "v5.2",
-        "n_zone_to_scale": 6,
-        "n_plant_to_scale": 50,
-        "n_plant_to_add": 100,
-    }
+    "demand": "vJan2021",
+    "hydro": "vJan2021",
+    "solar": "vJan2021",
+    "wind": "vJan2021",
+    "n_zone_to_scale": 6,
+    "n_plant_to_scale": 50,
+    "n_plant_to_add": 100,
+}
+profile_type = {
+    "wind": {"wind", "wind_offshore"},
+    "solar": {"solar"},
+    "hydro": {"hydro"},
+    "demand": {"demand"},
 }
 
 
 def get_zone_with_resource(base_grid, resource):
-    zone = list(
-        base_grid.plant.groupby("type").get_group(resource)["zone_name"].unique()
-    )
+    zone = base_grid.plant.groupby("type").get_group(resource)["zone_name"].unique()
     return zone
 
 
 def get_plant_with_resource(base_grid, resource):
-    plant_id = list(base_grid.plant.groupby("type").get_group(resource).index)
+    plant_id = base_grid.plant.groupby("type").get_group(resource).index
     return plant_id
 
 
 def get_change_table_for_zone_scaling(base_grid, resource):
-    n_zone = param["_".join(interconnect)]["n_zone_to_scale"]
+    n_zone = param["n_zone_to_scale"]
     zones = get_zone_with_resource(base_grid, resource)
 
     ct = ChangeTable(base_grid)
@@ -52,7 +55,7 @@ def get_change_table_for_zone_scaling(base_grid, resource):
 
 
 def get_change_table_for_id_scaling(base_grid, resource):
-    n_plant = param["_".join(interconnect)]["n_plant_to_scale"]
+    n_plant = param["n_plant_to_scale"]
     plants = get_plant_with_resource(base_grid, resource)
 
     ct = ChangeTable(base_grid)
@@ -70,7 +73,7 @@ def get_change_table_for_id_scaling(base_grid, resource):
 
 
 def get_change_table_for_new_plant_addition(base_grid, resource):
-    n_plant = param["_".join(interconnect)]["n_plant_to_add"]
+    n_plant = param["n_plant_to_add"]
     new_plant_bus_id = np.random.choice(
         base_grid.bus.index, size=n_plant, replace=False
     )
@@ -85,9 +88,12 @@ def get_change_table_for_new_plant_addition(base_grid, resource):
     return ct.ct
 
 
-def _check_plants_are_scaled(ct, base_grid, base_profile_resource, resource):
-    profile_info = base_profile_resource[0]
-    base_profile = base_profile_resource[1]
+def _check_plants_are_scaled(ct, base_grid, profile_info, raw_profile, resource):
+    plant_id_type = get_plant_with_resource(base_grid, resource)
+
+    base_profile = (
+        raw_profile[plant_id_type] * base_grid.plant.loc[plant_id_type, "Pmax"]
+    )
 
     tg = TransformGrid(base_grid, ct)
     transformed_grid = tg.get_grid()
@@ -115,12 +121,12 @@ def _check_plants_are_scaled(ct, base_grid, base_profile_resource, resource):
                 scaling_factor_plant.append(f)
 
     assert not base_profile.equals(transformed_profile)
-    assert transformed_profile[scaled_plant_id].equals(
-        base_profile[scaled_plant_id].multiply(scaling_factor_plant, axis=1)
+    assert_almost_equal(
+        transformed_profile[scaled_plant_id].values,
+        base_profile[scaled_plant_id].multiply(scaling_factor_plant, axis=1).values,
     )
 
-    all_plant_id = base_grid.plant.groupby("type").get_group(resource).index.tolist()
-    unscaled_plant_id = set(all_plant_id) - set(scaled_plant_id)
+    unscaled_plant_id = set(plant_id_type) - set(scaled_plant_id)
     if unscaled_plant_id:
         assert transformed_profile[unscaled_plant_id].equals(
             base_profile[unscaled_plant_id]
@@ -128,10 +134,14 @@ def _check_plants_are_scaled(ct, base_grid, base_profile_resource, resource):
     return transformed_profile
 
 
-def _check_new_plants_are_added(ct, base_grid, base_profile_resource, resource):
-    n_plant = param["_".join(interconnect)]["n_plant_to_add"]
-    profile_info = base_profile_resource[0]
-    base_profile = base_profile_resource[1]
+def _check_new_plants_are_added(ct, base_grid, profile_info, raw_profile, resource):
+    n_plant = param["n_plant_to_add"]
+    plant_id_type = (
+        base_grid.plant.isin(profile_type[resource]).query("type == True").index
+    )
+    base_profile = (
+        raw_profile[plant_id_type] * base_grid.plant.loc[plant_id_type, "Pmax"]
+    )
 
     tg = TransformGrid(base_grid, ct)
     transformed_grid = tg.get_grid()
@@ -149,39 +159,36 @@ def _check_new_plants_are_added(ct, base_grid, base_profile_resource, resource):
     return transformed_profile.drop(base_profile.columns, axis=1)
 
 
-def _check_new_plants_are_not_scaled(base_grid, base_profile_resource, resource):
+def _check_new_plants_are_not_scaled(base_grid, profile_info, raw_profile, resource):
     ct_zone = get_change_table_for_zone_scaling(base_grid, resource)
     ct_id = get_change_table_for_id_scaling(base_grid, resource)
     ct_new = get_change_table_for_new_plant_addition(base_grid, resource)
     ct = {**ct_zone, **ct_id[resource], **ct_new}
     # profile of new plants
     new_profile = _check_new_plants_are_added(
-        ct_new, base_grid, base_profile_resource, resource
+        ct_new, base_grid, profile_info, raw_profile, resource
     )
     # transformed profile
     scaled_profile = _check_plants_are_scaled(
-        ct, base_grid, base_profile_resource, resource
+        ct, base_grid, profile_info, raw_profile, resource
     )
     # check that the profiles of new plants in the scaled profile are not scaled
     assert new_profile.equals(scaled_profile[new_profile.columns])
 
 
 def _check_profile_of_new_plants_are_produced_correctly(
-    base_grid, base_profile_resource, resource
+    base_grid, profile_info, raw_profile, resource
 ):
     ct_new = get_change_table_for_new_plant_addition(base_grid, resource)
     new_profile = _check_new_plants_are_added(
-        ct_new, base_grid, base_profile_resource, resource
+        ct_new, base_grid, profile_info, raw_profile, resource
     )
     neighbor_id = [d["plant_id_neighbor"] for d in ct_new["new_plant"]]
-    neighbor_pmax = base_grid.plant.loc[neighbor_id].Pmax.to_list()
     new_plant_pmax = [d["Pmax"] for d in ct_new["new_plant"]]
 
     for i, c in enumerate(new_profile.columns):
-        neighbor_profile = base_profile_resource[1][neighbor_id[i]]
-        assert new_profile[c].equals(
-            neighbor_profile.multiply(new_plant_pmax[i] / neighbor_pmax[i])
-        )
+        neighbor_profile = raw_profile[neighbor_id[i]]
+        assert new_profile[c].equals(neighbor_profile.multiply(new_plant_pmax[i]))
 
 
 @pytest.fixture(scope="module")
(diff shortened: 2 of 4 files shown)


rouille:
Loading the full profiles, especially hydro, is slow. We could speed up the code by using the `usecols` parameters of `pandas.read_csv` to load only the profiles of the interconnect in the `InputData` class. There, we have access to `scenario_info` and we could create a `Grid` object for the given interconnect to retrieve the plant id for the fuel type. Then we the list of plant id would be be passed to `usecols`.

danielolsen:
> Loading the full profiles, especially hydro, is slow. We could speed up the code by using the `usecols` parameters of `pandas.read_csv` to load only the profiles of the interconnect in the `InputData` class. There, we have access to `scenario_info` and we could create a `Grid` object for the given interconnect to retrieve the plant id for the fuel type. Then we the list of plant id would be be passed to `usecols`.

Is that faster? I think loading a grid right now can be kind of slow, with the gencost tables taking the longest time. Maybe we want to change the construction of those tables to be lazy, for cases like this where we don't need them.

rouille:
> > Loading the full profiles, especially hydro, is slow. We could speed up the code by using the `usecols` parameters of `pandas.read_csv` to load only the profiles of the interconnect in the `InputData` class. There, we have access to `scenario_info` and we could create a `Grid` object for the given interconnect to retrieve the plant id for the fuel type. Then we the list of plant id would be be passed to `usecols`.
> 
> Is that faster? I think loading a grid right now can be kind of slow, with the gencost tables taking the longest time. Maybe we want to change the construction of those tables to be lazy, for cases like this where we don't need them.

Good point. I wrote the following script:
```
import pandas as pd
import time

start = time.time()
hydro = pd.read_csv("/Users/brdo/ScenarioData/raw/usa_tamu/hydro_vJan2021.csv", index_col=0)
print(time.time() - start)

start = time.time()
from powersimdata.input.grid import Grid
grid = Grid("Texas")
plant_id = grid.plant.groupby("type").get_group("hydro").index
hydro = pd.read_csv("/Users/brdo/ScenarioData/raw/usa_tamu/hydro_vJan2021.csv", index_col=0, usecols=list(map(str, plant_id)))
print(time.time() - start)
```
and ran it for different interconnect. Here is what I get:
### Texas
```
4.015028953552246
Reading bus.csv
Reading plant.csv
Reading gencost.csv
Reading branch.csv
Reading dcline.csv
Reading sub.csv
Reading bus2sub.csv
Reading zone.csv
1.9824612140655518
```
### Western
```
3.9911980628967285
Reading bus.csv
Reading plant.csv
Reading gencost.csv
Reading branch.csv
Reading dcline.csv
Reading sub.csv
Reading bus2sub.csv
Reading zone.csv
2.5363688468933105
```
### Eastern
```
4.023339033126831
Reading bus.csv
Reading plant.csv
Reading gencost.csv
Reading branch.csv
Reading dcline.csv
Reading sub.csv
Reading bus2sub.csv
Reading zone.csv
4.254112958908081
```
### USA
```
3.955845832824707
Reading bus.csv
Reading plant.csv
Reading gencost.csv
Reading branch.csv
Reading dcline.csv
Reading sub.csv
Reading bus2sub.csv
Reading zone.csv
4.973045110702515
```
So it is faster for Texas and Western, similar for Eastern and slower for USA

powersimdata/input/tests/test_transform_profile.py
```diff
@@ -237,10 +244,10 @@ def base_demand():
 @pytest.mark.integration
 @pytest.mark.ssh
 def test_demand_is_scaled(base_grid, base_demand):
-    n_zone = param["_".join(interconnect)]["n_zone_to_scale"]
+    n_zone = param["n_zone_to_scale"]
     profile_info = base_demand[0]
-    base_profile = base_demand[1]
-
+    base_profile = base_demand[1][base_grid.id2zone.keys()]
+    print(base_profile)
```
danielolsen, line 250: Did you forget to remove this print?

rouille: Yep

powersimdata/input/transform_profile.py
```diff
@@ -1,111 +1,98 @@
...
         :param dict scenario_info: scenario information.
         :param powersimdata.input.grid.Grid grid: a Grid object.
         :param dict ct: change table.
         """
         self._input_data = InputData()
         self.scenario_info = scenario_info
-        self.grid = copy.deepcopy(grid)
+
         self.ct = copy.deepcopy(ct)
+        self.grid = copy.deepcopy(grid)
```
jon-hagg, line 21: Should we mention in the docstring that the grid passed in is already transformed? I was just realizing that we have that assumption built in here when dealing with new plants

rouille: yes we should. Unfortunately, we cannot checked the type since base grid and transformed grid are both `Grid` object. It should be mentioned in the docstring at the very least.

powersimdata/input/transform_profile.py
```diff
@@ -1,111 +1,98 @@
...
+        if self.n_new_clean_plant > 0:
+            new_profile = self._add_plant_profile(profile, resource)
+            return scaled_profile.join(new_profile)
         else:
-            if self.n_new_plant > 0:
-                power_output = self._add_plant_profile(power_output, resource)
-            if resource in self.ct.keys():
-                power_output = self._scale_plant_profile(power_output, resource)
-            return power_output
+            return scaled_profile
```
BainanXia, line 66: We are not dealing with `wind_offshore` here, are we?

rouille: Ah. The `plant_id` list is wrong for wind. I need to include `wind_offsore`. Thanks!

powersimdata/input/transform_profile.py
jon-hagg: I suppose we could change the parameter type to `TransformGrid` and call `get_grid`, but yeah for now just updating the docstring sounds good.

rouille: I thought about it. It is not optimal in the `Analyze` class because the grid is retrieve from the `mat` file and hence does not need to be transformed. We could get instantiate a `TransformGrid` object with an empty `dict` for the change table.

If we want to do that I would wait that we have the refactor of the `create` module first because the `TransformProfile` class is instantiated there.

powersimdata/input/input_data.py
```diff
@@ -56,10 +56,11 @@ def get_data(self, scenario_info, field_name):
...
 
         if field_name in ["demand", "hydro", "solar", "wind"]:
-            interconnect = interconnect2name(scenario_info["interconnect"].split("_"))
             version = scenario_info["base_" + field_name]
-            file_name = interconnect + "_" + field_name + "_" + version + "." + ext
-            from_dir = server_setup.BASE_PROFILE_DIR
+            file_name = field_name + "_" + version + "." + ext
+            from_dir = posixpath.join(
+                server_setup.BASE_PROFILE_DIR, scenario_info["grid_model"]
+            )
```
danielolsen, line 63: If someone ever tries to run the 'server' on a Windows machine, this could cause trouble, but I think that will be rare.

powersimdata/input/transform_profile.py
```diff
@@ -1,111 +1,98 @@
...
+
+        :return: (*tuple*) -- first element is the total number of new plant and second
+            element is the total number of new clean plant (*hydro*, *solar*,
+            *onshore wind* and *offshore wind*).
+        """
+        n_plant = [0, 0]
+        if "new_plant" in self.ct.keys():
+            for p in self.ct["new_plant"]:
+                n_plant[0] += 1
+                if p["type"] in sum(self.scale_keys.values(), []):
```
danielolsen, line 42: Do we refactor `self.scale_keys` to be able to support this line? If so, we could do the same thing based on the current structure using something like `if p["type"] in set().union(*self.scale_keys.values()):`. I think they are equivalent.

powersimdata/input/input_data.py
rouille: @jon-hagg, did you create a class that creates path smartly. I remember something in that vein

powersimdata/input/transform_profile.py
rouille: yeah, it was for the `sum`. Do you prefer to stick to `set`?

danielolsen: I think a set is more correct when the order doesn't matter and there can be no duplicates, but it doesn't really matter here. Mostly I wanted to understand the meaning of the change.

powersimdata/input/transform_profile.py
```diff
@@ -1,111 +1,98 @@
...
-            specified type with plant identification number  as columns and
-            UTC timestamp as index.
+        :return: (*pandas.DataFrame*) -- power output for generators of specified type
+            with plant identification number as columns and UTC timestamp as indices.
         """
-        power_output = self._input_data.get_data(self.scenario_info, resource)
-        if not bool(self.ct):
-            return power_output
+        plant_id = (
+            self.grid.plant.iloc[: len(self.grid.plant) - self.n_new_plant]
```
danielolsen, line 54: Can this line be simplified to just `self.grid.plant.iloc[:-self.n_new_plant]`?

powersimdata/input/transform_profile.py
```diff
@@ -1,111 +1,98 @@
...
+        return new_profile
 
     def _get_demand_profile(self):
-        """Returns scaled demand profile.
+        """Return scaled demand profile.
 
         :return: (*pandas.DataFrame*) -- data frame of demand.
         """
-        demand = self._input_data.get_data(self.scenario_info, "demand")
+        zone_id = interconnect2id[self.scenario_info["interconnect"]]
```
danielolsen, line 94: Is this equivalent to `zone_id = sorted(self.grid.bus.zone_id.unique())`?

powersimdata/input/tests/test_transform_profile.py
```diff
@@ -34,7 +33,7 @@ def get_plant_with_resource(base_grid, resource):
 
 
 def get_change_table_for_zone_scaling(base_grid, resource):
-    n_zone = param["_".join(interconnect)]["n_zone_to_scale"]
+    n_zone = param["n_zone_to_scale"]
```
danielolsen, line 36: Nice

powersimdata/input/tests/test_transform_profile.py
```diff
@@ -86,8 +85,13 @@ def get_change_table_for_new_plant_addition(base_grid, resource):
...
 
 def _check_plants_are_scaled(ct, base_grid, base_profile_resource, resource):
+    plant_id_type = base_grid.plant.groupby("type").get_group(resource).index
+
     profile_info = base_profile_resource[0]
-    base_profile = base_profile_resource[1]
+    base_profile = (
+        base_profile_resource[1][plant_id_type]
+        * base_grid.plant.loc[plant_id_type, "Pmax"]
+    )
```
danielolsen, line 94: I see what's going on here with `base_profile_resource`, but it's a little confusing. What do you think about changing the params for `_check_plants_are_scaled` to (ct, base_grid, **profile_info, base_profile**, resource), and then calling this function with e.g. `_check_plants_are_scaled(ct, base_grid, *base_solar, "solar")`? Then `base_solar` unpacks automatically and the call signature for this check is a bit clearer.

I think we can also simplify `_ = _check_plants_are_scaled(...)` to just `_check_plants_are_scaled(...)`.

powersimdata/input/transform_profile.py
rouille: I tried but If I remember correctly, I got an error when `self.n_new_plant` equals 0 with `self.grid.plant.iloc[:-self.n_new_plant]`

powersimdata/input/transform_profile.py
rouille: Yes, indeed.

powersimdata/input/tests/test_transform_profile.py
rouille: The other option would be to scale the profiles (not demand) in their respective fixture so you always get the normalized profile scaled by the Pmax of the plants

powersimdata/input/transform_profile.py
danielolsen: Aha, that makes sense.

powersimdata/input/tests/test_transform_profile.py
danielolsen: My comment was less about where you do the profile scaling, more about not wanting the `base_profile_resource` parameter to be a tuple that is immediately unpacked, when it would be clearer to have two parameters instead.

powersimdata/input/input_data.py
jon-hagg: I did something, but not smart enough for this case. Anyway, don't think it's an issue - either we will host the server, or users can run the linux container, which still works on windows.

powersimdata/input/transform_profile.py
rouille: I have updated the docstring

powersimdata/input/tests/test_transform_profile.py
rouille: Done

powersimdata/input/transform_profile.py
rouille: Fixed. Tests have been refactored accordingly.

powersimdata/input/transform_profile.py
rouille: Done

powersimdata/input/transform_profile.py
rouille: Done

powersimdata/input/tests/test_transform_profile.py
rouille: Done

powersimdata/input/tests/test_transform_profile.py
```diff
@@ -191,56 +203,63 @@ def base_grid():
...
+# @pytest.fixture(scope="module")
+# def raw_demand():
+#     input_data = InputData()
+#     profile_info = {
+#         "interconnect": "_".join(interconnect),
+#         "grid_model": "usa_tamu",
+#         "base_demand": param["demand"],
+#     }
+#     profile = input_data.get_data(profile_info, "demand")
+#     return profile_info, profile
```
BainanXia, line 247: Is there a reason we keep this here?

powersimdata/input/transform_profile.py
```diff
@@ -1,111 +1,109 @@
...
+            with plant identification number as columns and UTC timestamp as indices.
         """
-        power_output = self._input_data.get_data(self.scenario_info, resource)
-        if not bool(self.ct):
-            return power_output
+        plant_id = (
+            self.grid.plant.iloc[: len(self.grid.plant) - self.n_new_plant]
+            .isin(self.scale_keys[resource])
+            .query("type == True")
+            .index
```
BainanXia, line 57: This is nice.

powersimdata/input/tests/test_transform_profile.py
rouille: Nope. I just commented it before moving it but did not erase it afterwards. Thanks.

rouille: Done

powersimdata/network/usa_tamu/constants/zones.py
```diff
@@ -85,7 +85,11 @@
     | interconnect2loadzone["Western"]
     | interconnect2loadzone["Texas"]
 )
-
+# Map interconnect to load zone id
+interconnect2id = {
+    k: sorted([zone_df.reset_index().set_index("zone_name").loc[l].zone_id for l in v])
```
BainanXia, line 90: Is it equivalent to `sorted([zone_df[zone_df.zone_name == l].index for l in v])`?

rouille: I simplified it further:
```
interconnect2id = {
    k: set(zone_df.isin(v).query("zone_name == True").index)
    for k, v in interconnect2loadzone.items()
}
```

rouille: By the way, I ended up not using this `dict` since @danielolsen came up with a different way to obtain the mapping in the `TransformProfile` class saving an import. I leave it since it may come up handy in the future.

BainanXia: Nice.

powersimdata/input/tests/test_transform_profile.py
```diff
@@ -85,9 +88,16 @@ def get_change_table_for_new_plant_addition(base_grid, resource):
     return ct.ct
 
 
-def _check_plants_are_scaled(ct, base_grid, base_profile_resource, resource):
-    profile_info = base_profile_resource[0]
-    base_profile = base_profile_resource[1]
+def _check_plants_are_scaled(ct, base_grid, profile_info, raw_profile, resource):
+    plant_id_type = (
+        base_grid.plant.isin(profile_type[resource]).query("type == True").index
+    )
```
danielolsen, line 94: Nice, I will have to try to remember this.

powersimdata/input/tests/test_transform_profile.py
```diff
@@ -85,9 +88,16 @@ def get_change_table_for_new_plant_addition(base_grid, resource):
...
+
+    plant_id_type = get_plant_with_resource(base_grid, resource)
```
danielolsen, line 96: Does l.96 replace l. 92-94?

powersimdata/input/transform_profile.py
```diff
@@ -1,111 +1,109 @@
...
-        )
-        self.base_plant = Grid(self.grid.interconnect).plant
+        self.n_new_plant, self.n_new_clean_plant = self._get_number_of_new_plant()
+
+    def _get_number_of_new_plant(self):
+        """Return the total number of new plant and new clean plant.
+
+        :return: (*tuple*) -- first element is the total number of new plant and second
+            element is the total number of new clean plant (*hydro*, *solar*,
+            *onshore wind* and *offshore wind*).
```
danielolsen, line 36: `new clean plant` -> `new plants with profiles`. Geothermal and nuclear are clean too but don't have profiles.

powersimdata/input/tests/test_transform_profile.py
rouille: Yes. I forgot to delete the first assignment

powersimdata/network/usa_tamu/constants/zones.py
```diff
@@ -85,7 +85,11 @@
     | interconnect2loadzone["Western"]
     | interconnect2loadzone["Texas"]
 )
-
+# Map interconnect to load zone id
+interconnect2id = {
+    k: set(zone_df.isin(v).query("zone_name == True").index)
+    for k, v in interconnect2loadzone.items()
+}
```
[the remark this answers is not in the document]
BainanXia: I thought this is something new we implement in this PR, no?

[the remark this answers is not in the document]
danielolsen: I think we had been using this new dict, but refactored our way around needing it.

powersimdata/input/tests/test_transform_profile.py
```diff
@@ -115,23 +125,28 @@ def _check_plants_are_scaled(ct, base_grid, base_profile_resource, resource):
...
 
-def _check_new_plants_are_added(ct, base_grid, base_profile_resource, resource):
-    n_plant = param["_".join(interconnect)]["n_plant_to_add"]
-    profile_info = base_profile_resource[0]
-    base_profile = base_profile_resource[1]
+def _check_new_plants_are_added(ct, base_grid, profile_info, raw_profile, resource):
+    n_plant = param["n_plant_to_add"]
+    plant_id_type = (
+        base_grid.plant.isin(profile_type[resource]).query("type == True").index
+    )
```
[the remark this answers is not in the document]
rouille: No. In this case we need to query both the onshore and offshore wind profiles

powersimdata/network/usa_tamu/constants/zones.py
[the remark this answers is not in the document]
rouille: yeah we don't use it anymore. I would say it does not hurt to keep it. It is not impossible that we will need this mapping one day.

powersimdata/input/tests/test_transform_profile.py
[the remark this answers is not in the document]
danielolsen: Can we refactor `get_plant_with_resource` to also be able to take an iterable of resources in addition to just a single str? Something like:
```
resources = {resource} if isinstance(resource, str) else resource
base_grid.plant.isin(resources).query("type == True").index
```

Maybe it's not worth it just to enable this one line in `_check_new_plants_are_added` though. Up to you.

powersimdata/input/transform_profile.py
rouille: Done

powersimdata/input/tests/test_transform_profile.py
rouille: Done

powersimdata/input/tests/test_transform_profile.py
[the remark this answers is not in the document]
BainanXia: @danielolsen we have that in PostREISE `postreise/analyze/helpers.py/get_plant_id_for_resources_in_area` btw.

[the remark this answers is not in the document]
rouille: yeah but we cannot import from 

> @danielolsen we have that in PostREISE `postreise/analyze/helpers.py/get_plant_id_for_resources_in_area` btw.

yeah but we cannot import from **PostREISE**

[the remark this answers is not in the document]
rouille: I am not satisfied by the way the tests are built. The problem is that the `get_profile` method of the `TransformProfile` class return the onshore and offshore wind profiles for `resource=wind`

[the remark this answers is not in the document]
danielolsen: My mistake from many months ago continues to haunt us.

danielolsen (approved):
This looks good, and should help us refactor a lot of old code.

powersimdata/input/tests/test_transform_profile.py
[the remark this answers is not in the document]
rouille: I seriously spent a couple of hours after Bainan pointed out that we were not returning the offshore wind profiles trying to refactor the tests to query the right columns of the profile in the tests. To summarize, the problem I ran into is that we query onshore and offshore wind profiles in the `TransformProfile` calss but we scale/add them separately in the `ChangeTable` class and this complicates the testing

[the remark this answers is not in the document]
BainanXia: @danielolsen We made that decision together and now this is essentially everywhere, thus can't regret.

BainanXia (approved):
I think this is good to go. Thanks.

powersimdata/input/tests/test_transform_profile.py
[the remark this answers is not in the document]
danielolsen: I regret anyway.

PR #383, opened 2021-01-28
feat: let user specify more parameters for storage

### LINKED ISSUE #380: Allow user to specify more parameters for energy storage
The most important one here for our short-term work is duration, but while we're at it we could also allow specification for others (e.g. efficiencies). These are currently hardcoded in `powersimdata.network.usa_tamu.usa_tamu_model.TAMU._set_storage()`, and we would want to extend the functionality of `powersimdata.input.change_table.ChangeTable.add_storage_capacity()`.

danielolsen:
### Purpose
Let user specify more parameters for storage. Closes #380.

### What the code is doing
- In **abstract_grid.py**: we add parameters `LossFactor`, `terminal_min`, and `terminal_max`.
- In **change_table.py**: we refactor `add_storage_capacity` to follow the pattern set in `add_plant` etc.
- In **grid.py**: ~we add the `Lossfactor` parameter to the ignored keys for `grid.storage` in the grid equality check.~ we automatically populate the ignored keys for grid.storage in the grid equality check from the new **storage.py**.
- In **transform_grid.py**: we interpret the new change table schema.
- We add a new file **storage.py** to handle `usa_tamu` storage constants.
- In **usa_tamu_model.py**: we get our storage default values from the new **storage.py** file.
- We update **test_change_table.py** and **test_transform_grid.py** accordingly

### Testing
Unit tests are updated to the new format and now pass.

### Time estimate
15-30 minutes. New tests to ensure that the new parameters are set properly are still incoming, but review can get started on everything else.

### DIFF
--- a/powersimdata/input/change_table.py
+++ b/powersimdata/input/change_table.py
@@ -462,31 +462,58 @@ class ChangeTable(object):
         """
         scale_congested_mesh_branches(self, ref_scenario, **kwargs)
 
-    def add_storage_capacity(self, bus_id):
+    def add_storage_capacity(self, info):
         """Sets storage parameters in change table.
 
-        :param dict bus_id: key(s) for the id of bus(es), value(s) is (are)
-            capacity of the energy storage system in MW.
-        :raises TypeError: if bus_id is not a dict.
-        :raises ValueError: if bus_id contains any bus ids not present in the grid,
-            or any non-positive values are given.
+        :param list info: each entry is a dictionary. The dictionary gathers
+            the information needed to create a new storage device.
+        :raises TypeError: if info is not a list.
+        :raises ValueError: if any of the new storages to be added have bad values.
         """
-        if not isinstance(bus_id, dict):
-            raise TypeError("bus_id must be a dict")
+        if not isinstance(info, list):
+            raise TypeError("Argument enclosing new storage(s) must be a list")
+
+        info = copy.deepcopy(info)
+        new_storages = []
+        required = {"bus_id", "capacity"}
+        optional = {
+            "duration",
+            "min_stor",
+            "max_stor",
+            "energy_value",
+            "InEff",
+            "OutEff",
+            "LossFactor",
+            "terminal_min",
+            "terminal_max",
+        }
         anticipated_bus = self._get_new_bus()
-        diff = set(bus_id.keys()).difference(set(anticipated_bus.index))
-        if len(diff) != 0:
-            raise ValueError(f"No bus with the following id: {', '.join(diff)}")
-        for k, v in bus_id.items():
-            if not isinstance(v, (float, int)):
-                raise ValueError(f"values must be numeric, bad type for bus {k}")
-            if v <= 0:
-                raise ValueError(f"values must be positive, bad value for bus {k}")
+        for i, storage in enumerate(info):
+            self._check_entry_keys(storage, i, "storage", required, None, optional)
+            if storage["bus_id"] not in anticipated_bus.index:
+                raise ValueError(
+                    f"No bus id {storage['bus_id']} available for {ordinal(i)} storage"
+                )
+            for o in optional:
+                if o not in storage:
+                    storage[o] = self.grid.storage[o]
+            for k, v in storage.items():
+                if not isinstance(v, (int, float)):
+                    err_msg = f"values must be numeric, bad type for {ordinal(i)} {k}"
+                    raise ValueError(err_msg)
+                if v < 0:
+                    raise ValueError(
+                        f"values must be non-negative, bad value for {ordinal(i)} {k}"
+                    )
+            for k in {"min_stor", "max_stor", "InEff", "OutEff", "LossFactor"}:
+                if storage[k] > 1:
+                    raise ValueError(
+                        f"value for {k} must be <=1, bad value for {ordinal(i)} storage"
+                    )
+            new_storages.append(storage)
         if "storage" not in self.ct:
-            self.ct["storage"] = {}
-        if "bus_id" not in self.ct["storage"]:
-            self.ct["storage"]["bus_id"] = {}
-        self.ct["storage"]["bus_id"].update(bus_id)
+            self.ct["storage"] = []
+        self.ct["storage"] += new_storages
 
     def add_dcline(self, info):
         """Adds HVDC line(s).
--- a/powersimdata/input/transform_grid.py
+++ b/powersimdata/input/transform_grid.py
@@ -356,7 +356,7 @@ class TransformGrid(object):
 
     def _add_storage(self):
         """Adds storage to the grid."""
-        first_storage_id = self.grid.plant.shape[0] + 1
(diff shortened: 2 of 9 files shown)


danielolsen:
In addition to adding the LossFactor parameter, I've also added two more parameters, `terminal_min` and `terminal_max`, which constraint the amount of energy that the storage unit must end the interval with. Previously, this had been hardcoded to be between 50-100%; now the default values are 0-100% but the user can specify anything they wish.

rouille:
It reminds me something ...

powersimdata/input/change_table.py
```diff
@@ -462,31 +462,57 @@ def scale_congested_mesh_branches(self, ref_scenario, **kwargs):
...
+                raise ValueError(
+                    f"No bus id {storage['bus_id']} available for {ordinal(i)} storage"
+                )
+            for o in optional:
+                if o not in storage:
+                    storage[o] = self.grid.storage[o]
+            for k, v in storage.items():
+                if not isinstance(v, (int, float)):
+                    err_msg = f"values must be numeric, bad type for {ordinal(i)} {k}"
+                    raise ValueError(err_msg)
```
rouille, line 503: I know we already had this discussion in the past but should not we raise a `TypeError` here instead?

danielolsen: Haha. The input is the correct dict type, but the value of the dict that we are passing contains something of the wrong type. Which is the 'right' error?

rouille: I know ...

powersimdata/input/transform_grid.py
```diff
@@ -356,29 +356,33 @@ def _add_gencost(self):
...
+        for i, entry in enumerate(self.ct["storage"]):
+            storage_id = first_storage_id + i
+            self._add_storage_unit(entry)
             self._add_storage_gencost()
             self._add_storage_genfuel()
-            self._add_storage_data(storage_id, value)
+            self._add_storage_data(storage_id, entry)
+        # Ensure that int columns remain int after the appends convert them to float
+        storage = self.grid.storage
+        storage["gen"] = storage["gen"].astype({"bus_id": "int", "status": "int"})
```
rouille, line 368: should we move this line into `_add_storage_unit`?

powersimdata/input/transform_grid.py
```diff
@@ -356,29 +356,33 @@ def _add_gencost(self):
...
+        storage["StorageData"] = storage["StorageData"].astype({"UnitIdx": "int"})
```
rouille, line 369: and this one into `_add_storage_data`

powersimdata/input/transform_grid.py
danielolsen: Yeah, I think that makes more sense

powersimdata/input/change_table.py
```diff
@@ -462,31 +462,57 @@ def scale_congested_mesh_branches(self, ref_scenario, **kwargs):
...
+                if v < 0:
+                    raise ValueError(
+                        f"values must be non-negative, bad value for {ordinal(i)} {k}"
+                    )
+            for k in {"min_stor", "max_stor", "InEff", "OutEff", "LossFactor"}:
+                if storage[k] > 1:
+                    err_msg = f"value for {k} must be <=1, bad value for {ordinal(i)}"
```
jon-hagg, line 510: Should this message have a `{k}` at the end? Looks like it would end up something like "...bad value for 3rd"

danielolsen: I see what you mean. We could do "3rd storage", if you think that would be clearer.

jon-hagg: Yeah, however you want to phrase it is fine, I was mainly checking if it was a typo ("3rd storage" makes sense to me though).

powersimdata/input/transform_grid.py
danielolsen: Done.

powersimdata/input/transform_grid.py
danielolsen: Done.

powersimdata/input/change_table.py
danielolsen: Done.

rouille (approved):
Thanks for the improvement.

powersimdata/input/change_table.py
```diff
@@ -462,31 +462,58 @@ def scale_congested_mesh_branches(self, ref_scenario, **kwargs):
...
-                raise ValueError(f"values must be positive, bad value for bus {k}")
+        for i, storage in enumerate(info):
+            self._check_entry_keys(storage, i, "storage", required, None, optional)
+            if storage["bus_id"] not in anticipated_bus.index:
+                raise ValueError(
+                    f"No bus id {storage['bus_id']} available for {ordinal(i)} storage"
+                )
+            for o in optional:
+                if o not in storage:
+                    storage[o] = self.grid.storage[o]
```
BainanXia, line 499: We should keep in mind that this won't work if we add storage in the base grid (although very unlikely).

BainanXia (approved):
Thanks for the improvements.

powersimdata/input/change_table.py
danielolsen: What do you mean?

BainanXia: If we have storage devices in the base grid, it will be listed in grid.storage right? And in that case, grid.storage will be no longer default values, right?

danielolsen: Storage devices that are part of the base grid will have entries in `grid.storage["gen"]`, `grid.storage["StorageData"]`, etc., but I believe `grid.storage` would still have top-level keys that give some default parameters for construction of new entries (like we have when we load `grid.mat` files that have storage).

BainanXia: I guess I'm trying to understand the following situation: if we have a storage device with all the optional parameters set, then they are listed in `grid.storage["StorageData"]` and we still have default values stored in the top level dictionary of `grid.storage`, right?

danielolsen: Exactly. You should be able to play around with a scenario in Create state and see exactly that.

powersimdata/input/transform_grid.py
```diff
@@ -356,32 +356,33 @@ def _add_gencost(self):
 
     def _add_storage(self):
         """Adds storage to the grid."""
-        storage_id = self.grid.plant.shape[0]
-        for bus_id, value in self.ct["storage"]["bus_id"].items():
-            storage_id += 1
-            self._add_storage_unit(bus_id, value)
+        first_storage_id = self.grid.plant.index.max() + 1
```
rouille, line 359: I guess we could rename `first_storage_id` to `storage_id` and then do `storage_id += i` in the loop

danielolsen: For some reason I didn't get the email about this message until after the merge, sorry.

rouille: No worries. Very minor.

PR #384, opened 2021-01-29
fix: correct bug with repeated substation in TransformGrid._add_bus

danielolsen:
### Purpose
Fix an Exception that's thrown when you add two new buses with the same coordinates.

### What the code is doing
In `TransformGrid` the `latlon2sub` variable holds a dictionary of `{(lat, lon): list_of_substation_ids}`. When we add a bus at a new location, we create a new substation, and we want to add its id to this lookup dict, but we had been adding it as an int, rather than a list containing that int.

**test_transform_grid** is updated to check for this bug.

### Testing
The unit test is updated to check for this bug, and after the fix all tests pass again.

### Time estimate
2 minutes.

### DIFF
--- a/powersimdata/input/tests/test_transform_grid.py
+++ b/powersimdata/input/tests/test_transform_grid.py
@@ -506,13 +506,15 @@ def test_add_bus(ct):
     prev_max_bus = grid.bus.index.max()
     prev_num_subs = len(grid.sub.index)
     ct.ct["new_bus"] = [
-        # These two are buses at new locations
+        # These three are buses at new locations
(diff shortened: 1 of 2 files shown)


BainanXia (approved):
Good catch!

danielolsen:
> 
> 
> Good catch!

I caught it the hard way 😩

rouille:
Where is the Exception raised when you add two buses at the same location?

danielolsen:
> Where is the Exception raised when you add two buses at the same location?

`sub_id = latlon2sub[(lat, lon)][0]` within `TransformGrid._add_bus`. We get `IndexError: invalid index to scalar variable.` when we try to add the second new bus, since we try to `[0]` into an int.

PR #389, opened 2021-02-05
ci: use black profile for isort

jon-hagg:
### Purpose
We can take advantage of isort's built in "profile" for black compatibility. This includes a line length option which we were msising, along with a couple others I'm not sure about. [Source](https://github.com/PyCQA/isort/blob/develop/isort/profiles.py) from their repo is easy to read.

### What the code is doing
Added isort section in tox.ini

### Testing
Ran `tox -e format` locally, and ci passes.

### Time estimate
5 min

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN ln -s /mnt/bes/pcm $HOME/ScenarioData
 
 COPY powersimdata/utility/templates /mnt/bes/pcm/
 
-WORKDIR /app
+WORKDIR /PowerSimData
 COPY Pipfile .
 COPY Pipfile.lock .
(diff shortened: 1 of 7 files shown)


BainanXia (approved):
Thanks.

Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN ln -s /mnt/bes/pcm $HOME/ScenarioData
 
 COPY powersimdata/utility/templates /mnt/bes/pcm/
 
-WORKDIR /app
+WORKDIR /PowerSimData
```
jon-hagg, line 9: This is unrelated, but wanted to keep it. Has no effect except in docker images based on this one.

PR #391, opened 2021-02-08
fix: calculate investment cost for nuclear generators

danielolsen:
### Purpose
Allow investment costs to be calculated for nuclear generators.

### What the code is doing
The functional change is in **const.py**: the set `gen_inv_cost_techdetails_to_keep` defines the only entries to keep from the NREL ATB data, and none of the existing ones cover nuclear. The category `"*"` covers nuclear and nuclear only, so there should be no side effects.

The changes in **test_investment_costs.py** are primarily to create a test for nuclear (the changes to `mock_plant` and `test_calculate_gen_inv_costs_2030`), and secondarily to clean up the code to make it a bit more readable.

### Testing
The first commit adds the test, which will fail. The second commit implements the fix and the tests pass.

### Time estimate
5 minutes.

### DIFF
--- a/powersimdata/design/investment/const.py
+++ b/powersimdata/design/investment/const.py
@@ -79,6 +79,7 @@ gen_inv_cost_techdetails_to_keep = {
     "LTRG1",  # Single tech for wind
     "4Hr Battery Storage",  # Single tech for storage
     "Seattle",  # Single tech for solar
+    "*",  # Single tech for nuclear
 }
 
 regional_multiplier_gen_translation = {
--- a/powersimdata/design/investment/tests/test_investment_costs.py
+++ b/powersimdata/design/investment/tests/test_investment_costs.py
@@ -39,10 +39,10 @@ mock_branch["to_lon"] = [
 ]
 
 mock_plant = {
-    "plant_id": ["A", "B", "C", "D", "E", "F", "G"],
-    "bus_id": [2010228, 2010228, 2021106, 2010319, 2010319, 2010319, 2010320],
-    "type": ["solar", "coal", "wind", "solar", "solar", "ng", "wind"],
-    "Pmax": [15, 30, 10, 12, 8, 20, 15],
+    "plant_id": ["A", "B", "C", "D", "E", "F", "G", "H"],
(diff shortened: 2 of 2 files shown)


powersimdata/design/investment/tests/test_investment_costs.py
```diff
@@ -97,18 +97,20 @@ def test_calculate_ac_inv_costs(mock_grid):
...
-        2020
-    )  # terminals
+    expected_dc_cost = (
+        # lines
+        10 * 679.1799258421203 * 457.1428571 * calculate_inflation(2015)
+        # terminals
+        + 135e3 * 10 * 2 * calculate_inflation(2020)
+    )
     dc_cost = _calculate_dc_inv_costs(mock_grid)
     assert dc_cost == pytest.approx(expected_dc_cost)
```
rouille, line 107: There are no changes here. Just reorganizing the lines, correct?

danielolsen: Correct.

powersimdata/design/investment/const.py
```diff
@@ -79,6 +79,7 @@
     "LTRG1",  # Single tech for wind
     "4Hr Battery Storage",  # Single tech for storage
     "Seattle",  # Single tech for solar
+    "*",  # Single tech for nuclear
```
rouille, line 82: That is a pretty bad choice for accessing nuclear information.

rouille (approved):
Looks good

powersimdata/design/investment/const.py
danielolsen: To be clear, this is not wildcard matching. The CSV file literally has an entry `*`.

rouille: Yeah, it is terrible. I guess there is a documentation but anything else could be more explanatory.

danielolsen: There is definitely room for improvement. Currently, we're somewhat-arbitrarily selecting one `TechDetail` to consider for each technology, hard-coded. It would be nice to give the user more flexibility to select different ones, when they are available, or maybe even to define their own.

PR #394, opened 2021-02-12
data: set Pmin for solar and wind gens to 0

danielolsen:
### Purpose
Set Pmin for all solar & wind plants to 0. We do this automatically in our optimization, but our data still shows non-zero Pmins for some solar & wind gens, and we should update this in case people use the data in any other way besides with the current version of REISE.jl.

```python
>>> from powersimdata.input.grid import Grid
>>> grid = Grid('USA')
Reading bus.csv
Reading plant.csv
Reading gencost.csv
Reading branch.csv
Reading dcline.csv
Reading sub.csv
Reading bus2sub.csv
Reading zone.csv
>>> grid.plant.groupby('type').sum().Pmin
type
coal              84956.595090
dfo               10151.991371
geothermal            0.000000
hydro             29889.259935
ng               163825.174150
nuclear           26988.915047
other                 0.000000
solar              6279.596845
wind              22190.243671
wind_offshore         0.000000
Name: Pmin, dtype: float64
```

### What the code is doing
No code, only data changes, which were achieved by:
```python
>>> import pandas as pd
>>> df = pd.read_csv("plant.csv", dtype=str, index_col=0)
>>> df.loc[df.type.isin(["solar", "wind"]), "Pmin"] = "0"
>>> df.to_csv("plant.csv")
```
Reading as `dtype=str` prevents any changes due to floating-point cruft in the other columns.

### Testing
With a copy of the original **plant.csv** as **plant_backup.csv**, we see that there are no changes besides the `Pmin` for solar & wind gens:
```python
>>> import pandas as pd
>>> df_orig = pd.read_csv("plant_backup.csv", index_col=0)
>>> df_new = pd.read_csv("plant.csv", index_col=0)
>>> df_orig[set(df_orig.columns) - {"Pmin"}].equals(df_new[set(df_new.columns) - {"Pmin"}])
True
>>> df_orig.loc[~df_orig.type.isin(["solar", "wind"])].equals(df_new.loc[~df_new.type.isin(["solar", "wind"])])
True
```

### Time estimate
5 minutes.


BainanXia (approved):
Good catch. I think that inherits from the original tamu model, either the case file or the aux file.

rouille (approved):

PR #401, opened 2021-02-19
feat: add temporal curtailment calculation

danielolsen:
### Purpose
Add functionality to pre-calculate curtailment caused by temporal mismatch between demand, inflexible generation, and curtailable profile resources.

### What the code is doing

First, profiles for hydro, solar, and wind are summed by type, with the exclusion of any plant IDs which the user specifies do not behave like all the rest (e.g., maybe some solar plants are behind-the-meter and are not curtailable, and we would want to see how this impact the rest of the wind plants).

Then, a time-series of inflexible generation is built-up, first based on on per-generator-type logic for plants without plant-level overrides (default values mimic what REISE.jl does currently), then based on the behavior of plant-level overrides.

Finally, the summed profile of curtailable resources is compared against the demand minus firm generation to get a time-series of the maximum possible use of these curtailable resources (assuming no energy storage) and this is normalized based on the total available energy from the curtailable resources.

### Testing
Unit tests still need to be added, but integration tests behave as expected (matching previously-done manual calculations). 

### Usage Example/Visuals
First, let's see the minimum solar/wind curtailment with default values:
```python
>>> from powersimdata.design.generation.curtailment import temporal_curtailment
>>> from powersimdata.scenario.scenario import Scenario
>>> scenario = Scenario(2505)
[...truncated...]
>>> temporal_curtailment(scenario)
[...truncated...]
0.25601475590030276
```

What if the nuclear and geothermal plants are running at 100%? This sort of calculation is useful when calculating total potential clean energy.
```python
>>> temporal_curtailment(scenario, pmin_by_type={'nuclear': 1, 'geothermal': 1})
[...truncated...]
0.2594608282619938
```
As expected, this would lead to higher curtailment of wind and solar. What if most nuclear and geothermal plants are running at 100%, but there are a few nuclear plants with integrated storage, so we assume that these plants can go down to 0 when necessary?
```python
>>> temporal_curtailment(scenario, pmin_by_type={'nuclear': 1, 'geothermal': 1}, pmin_by_id={p: 0 for p in range(14026, 14032)})
[...truncated...]
0.24669058923694587
```
As expected, this would lead to lower curtailment.

### Time estimate
15 minutes.

### DIFF
--- a/powersimdata/design/generation/curtailment.py
+++ b/powersimdata/design/generation/curtailment.py
@@ -1,0 +1,127 @@
+import pandas as pd
+
+from powersimdata.scenario.scenario import Scenario
+
+default_pmin_dict = {
+    "coal": None,
+    "dfo": 0,
+    "geothermal": 0.95,
+    "hydro": None,
+    "ng": 0,
+    "nuclear": 0.95,
+    "other": 0,
+    "solar": 0,
+    "wind": 0,
+    "wind_offshore": 0,
+}
+profile_methods = {
+    "hydro": "get_hydro",
+    "solar": "get_solar",
+    "wind": "get_wind",
+    "wind_offshore": "get_wind",
+}
+
+
+def temporal_curtailment(
+    scenario, pmin_by_type=None, pmin_by_id=None, curtailable={"solar", "wind"}
+):
+    """Calculate the minimum share of potential renewable energy that will be curtailed
+        due to supply/demand mismatch, assuming no storage is present.
+
+    :param powersimdata.scenario.scenario.Scenario scenario: scenario instance.
+    :param dict/pandas.Series pmin_by_type: Mapping of types to Pmin assumptions. Values
+        between 0 and 1 (inclusive) are treated as shares of Pmax, and None values
+        either maintain given Pmin values (for dispatchable resources) or track profiles
+        (for profile resources, e.g. hydro).
+    :param dict/pandas.Series pmin_by_id: Mapping of IDs to Pmin assumptions, as an
+        override to the default behavior for that plant type. Values between 0 and 1
+        (inclusive) are treated as shares of Pmax, and None values either maintain given
+        Pmin values (for dispatchable resources) or track profiles
+        (for profile resources, e.g. hydro).
+    :param iterable curtailable: resource types which can be curtailed.
+    :return: (*float*) -- share of curtailable resources that will be curtailed.
+    :raises TypeError: if inputs do not mach specified type.
+    :raises ValueError: if any entries in curtailable or keys in pmin_by_type are not
+        types in the grid, any keys in pmin_by_id are not plant IDs in the Grid,
+        or any values in pmin_by_type/pmin_by_id are not in the range [0, 1] or None.
+    """
+    if not isinstance(scenario, Scenario):
+        raise TypeError("scenario must be a Scenario")
+    if pmin_by_type is None:
+        pmin_by_type = {}
+    if pmin_by_id is None:
+        pmin_by_id = {}
+    check_dicts = {"pmin_by_id": pmin_by_id, "pmin_by_type": pmin_by_type}
+    for name, d in check_dicts.items():
+        if not isinstance(d, (dict, pd.Series)):
+            raise TypeError(f"{name} must be a dict or pandas Series")
+        # Access values via appropriate method whether d is a dict or a pandas Series
+        values = d.values() if isinstance(d, dict) else d.values
+        if not all([v is None or 0 <= v <= 1 for v in values]):
+            err_msg = f"all entries in {name} must be None or in the range [0, 1]"
+            raise ValueError(err_msg)
+    plant = scenario.state.get_grid().plant
+    valid_types = plant["type"].unique()
+    if not set(pmin_by_type.keys()) <= set(valid_types):
+        raise ValueError("Got invalid plant type as a key to pmin_by_type")
+    if not set(pmin_by_id.keys()) <= set(plant.index):
+        raise ValueError("Got invalid plant id as a key to pmin_by_id")
+    try:
+        if not set(curtailable) <= set(valid_types):
+            raise ValueError("Got invalid plant type within curtailable")
+    except TypeError:
+        raise TypeError("curtailable must be an iterable")
+
+    # Get profiles, filter out plant-level overrides, then sum
+    all_profiles = pd.concat(
+        [getattr(scenario.state, m)() for m in set(profile_methods.values())], axis=1
+    )
+    plant_id_mask = ~plant.index.isin(pmin_by_id.keys())
+    base_plant_ids_by_type = plant.loc[plant_id_mask].groupby("type").groups
+    valid_profile_types = set(base_plant_ids_by_type) & set(profile_methods)
+    plant_ids_for_summed_profiles = set().union(
+        *[set(base_plant_ids_by_type[g]) for g in valid_profile_types]
+    )
+    summed_profiles = (
+        all_profiles[plant_ids_for_summed_profiles].groupby(plant["type"], axis=1).sum()
+    )
+
+    # Build up a series of firm generation
+    summed_demand = scenario.state.get_demand().sum(axis=1)
+    firm_generation = pd.Series(0, index=summed_demand.index)
+    # Add plants without plant-level overrides ('base' plants)
+    pmin_dict = {**default_pmin_dict, **pmin_by_type}
+    # Don't iterate over plant types not present in this grid
+    pmin_dict = {k: pmin_dict[k] for k in base_plant_ids_by_type}
+    for resource, pmin in pmin_dict.items():
+        if (resource in curtailable) or (pmin == 0):
+            continue
+        if pmin is None:
+            if resource in profile_methods:
+                firm_generation += summed_profiles[resource]
+            else:
+                summed_pmin = plant.Pmin.loc[base_plant_ids_by_type[resource]].sum()
+                firm_generation += pd.Series(summed_pmin, index=summed_demand.index)
+        else:
+            summed_pmin = pmin * plant.Pmax.loc[base_plant_ids_by_type[resource]].sum()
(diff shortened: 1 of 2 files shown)


powersimdata/design/generation/curtailment.py
```diff
@@ -0,0 +1,124 @@
...
+    "solar": 0,
+    "wind": 0,
+    "wind_offshore": 0,
+}
+profile_methods = {
+    "hydro": "get_hydro",
+    "solar": "get_solar",
+    "wind": "get_wind",
+    "wind_offshore": "get_wind",
+}
```
rouille, line 22: Should we write a `get_onshore_wind` and `get_offshore_wind` methods in the `Analyze` class that would relieve our brain once for all?

powersimdata/design/generation/curtailment.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def temporal_curtailment(
+    scenario, pmin_by_type=None, pmin_by_id=None, curtailable={"solar", "wind"}
+):
+    """Calculate the minimum share of potential renewable energy that will be curtailed
+        due to supply/demand mismatch, assuming no storage is present.
+
+    :param powersimdata.scenario.scenario.Scenario scenario: scenario instance.
+    :param dict/pandas.Series pmin_by_type: Mapping of types to Pmin assumptions. Values
```
rouille, line 32: it seems that this parameter will always be a `dict`

powersimdata/design/generation/curtailment.py
```diff
@@ -0,0 +1,124 @@
...
+        between 0 and 1 (inclusive) are treated as shares of Pmax, and None values
+        either maintain given Pmin values (for dispatchable resources) or track profiles
+        (for profile resources, e.g. hydro).
+    :param dict/pandas.Series pmin_by_id: Mapping of IDs to Pmin assumptions, as an
```
rouille, line 36: idem

powersimdata/design/generation/curtailment.py
```diff
@@ -0,0 +1,124 @@
...
+    "dfo": 0,
+    "geothermal": 0.95,
+    "hydro": None,
+    "ng": 0,
+    "nuclear": 0.95,
+    "other": 0,
+    "solar": 0,
+    "wind": 0,
+    "wind_offshore": 0,
+}
```
rouille, line 16: I believe 0.95 of `pmax` is standard for nuclear and geothermal. I would have imagined that `geothermal` would be dispatchable.

powersimdata/design/generation/curtailment.py
```diff
@@ -0,0 +1,124 @@
...
+    :raises ValueError: if any entries in curtailable or keys in pmin_by_type are not
+        types in the grid, any keys in pmin_by_id are not plant IDs in the Grid,
+        or any values in pmin_by_type/pmin_by_id are not in the range [0, 1] or None.
+    """
+    if not isinstance(scenario, Scenario):
+        raise TypeError("scenario must be a Scenario")
+    if pmin_by_type is None:
+        pmin_by_type = {}
+    if pmin_by_id is None:
+        pmin_by_id = {}
```
rouille, line 53: Should we set the default value to an empty `dict` instead of `None` since it is overwritten above

powersimdata/design/generation/curtailment.py
```diff
@@ -0,0 +1,124 @@
...
+    check_dicts = {"pmin_by_id": pmin_by_id, "pmin_by_type": pmin_by_type}
+    for name, d in check_dicts.items():
+        if not isinstance(d, (dict, pd.Series)):
+            raise TypeError(f"{name} must be a dict or pandas Series")
+        if not all([v is None or 0 <= v <= 1 for v in d.values()]):
```
rouille, line 58: It seems that `d` can either be a `dict` or a `pd.Series`. So if it is a `pd.Series`, `d.values()` won't work. No?

powersimdata/design/generation/curtailment.py
danielolsen: We need this to work for Create as well.

powersimdata/design/generation/curtailment.py
BainanXia: I think @jon-hagg suggested we should avoid using mutables as default values of a function.

powersimdata/design/generation/curtailment.py
danielolsen: I just encoded what we do right now in REISE.jl. Nuclear and geothermal are set to 95%, and everything else besides coal is set to 0.

powersimdata/design/generation/curtailment.py
rouille: Indeed. This thing is hunting me.

powersimdata/design/generation/curtailment.py
danielolsen: Good point, since we don't modify the `pmin_by_dict` or `pmin_by_type` parameters, it should be safe to use an object as the default value.

rouille: Ah. Even if it will be altered only when it is empty?

powersimdata/design/generation/curtailment.py
danielolsen: You're right, I had originally thought that we can be flexible on inputs because a Series has an equivalent `items()` method, but it does not have an equivalent `values()` method.

powersimdata/design/generation/curtailment.py
danielolsen: It seems safe here, but maybe we want to maintain the discipline to never do this so that we don't slip up when it matters.

powersimdata/design/generation/curtailment.py
rouille: I guess we will test all that when you have time to do write the test module

danielolsen: I added a small change that should fix this issue.

powersimdata/design/generation/curtailment.py
```diff
@@ -0,0 +1,124 @@
...
+    valid_types = plant["type"].unique()
+    if not set(pmin_by_type.keys()) <= set(valid_types):
+        raise ValueError("Got invalid plant type as a key to pmin_by_type")
+    if not set(pmin_by_id.keys()) <= set(plant.index):
+        raise ValueError("Got invalid plant id as a key to pmin_by_id")
+    try:
+        if not set(pmin_by_type.keys()) <= set(valid_types):
+            raise ValueError("Got invalid plant type within curtailable")
+    except TypeError:
+        raise TypeError("curtailable must be an iterable")
```
rouille, line 73: you lost me here. The `try` does not do anything different than line 63 where we check for plant type. I don't see how we test for curtailable in the `try`

danielolsen: Good catch, I made a mistake in my copy/paste. `pmin_by_type.keys()` should be `curtailable`, and then we will get a TypeError on the `set()` call if curtailable is not an iterable. This seems to be the most pythonic way to check for an iterable.

danielolsen: Fixed.

powersimdata/design/generation/curtailment.py
```diff
@@ -0,0 +1,126 @@
...
+
+    # Get profiles, filter out plant-level overrides, then sum
+    all_profiles = pd.concat(
+        [getattr(scenario.state, m)() for m in set(profile_methods.values())], axis=1
+    )
+    plant_id_mask = ~plant.index.isin(pmin_by_id.keys())
+    base_plant_ids_by_type = plant.loc[plant_id_mask].groupby("type").groups
+    plant_ids_for_summed_profiles = set().union(
+        *[set(base_plant_ids_by_type[g]) for g in profile_methods]
+    )
```
rouille, line 83: We can do this here:
```
import itertools
plant_ids_for_summed_profiles = set(itertools.chain.from_iterable([base_plant_ids_by_type[g] for g in profile_methods]))
```
Not sure it is better

rouille:
It looks good. I believe we should wait for the function to be tested before merging.

powersimdata/design/generation/curtailment.py
rouille: The fact that the values of the `pmin_by_type` dictionary can be a `dict` or `pandas.Series` should be mentioned in the description.

danielolsen: `pmin_by_type` can be either a dict or a Series, the values should all be float either way.

rouille: Sorry I got confused with `check_dicts` that encapsulate `pmin_by_type` and `pmin_by_id` in a `dict`

powersimdata/design/generation/curtailment.py
```diff
@@ -0,0 +1,126 @@
...
+    summed_profiles = (
+        all_profiles[plant_ids_for_summed_profiles]
+        .groupby(plant.loc[plant_ids_for_summed_profiles, "type"], axis=1)
```
BainanXia, line 86: I think we can simply do `all_profiles[plant_ids_for_summed_profiles].groupby(plant.type, axis=1).sum()` here.

powersimdata/design/generation/curtailment.py
danielolsen: I would prefer to avoid an extra import.

powersimdata/design/generation/curtailment.py
danielolsen: You're right. I thought there might be a problem since `plant.type` is a different length than `all_profiles[plant_ids_for_summed_profiles]`, but it seems to work since `plant.type` has an index where the appropriate entries can be looked up.

danielolsen: Changed.

danielolsen:
Tests are added. They helped catch a couple of bugs for grids that don't have all the resources in the `default_pmin_dict` (e.g. geothermal) and `profile_methods` (e.g. `"wind_offshore"`).

rouille (approved):
Thanks

PR #402, opened 2021-02-19
ci: use pre defined action to trigger workflow

jon-hagg:
### Purpose
Use the action defined in our actions [repo](https://github.com/Breakthrough-Energy/actions).

### What the code is doing
The same thing it was before

### Testing
Will be tested by merging

### Time estimate
3 min


rouille (approved):
i was waiting for this one. I guess PreREISE and PostREISE are next

PR #404, opened 2021-02-23
feat: add optional solver parameter to launch_simulation

jon-hagg:
### Purpose
Corresponding changes for the new solver option in REISE.jl. This allows a user to specify either gurobi or glpk when launching a simulation, which closes the loop Daniel started by allowing multiple solvers. 

### What the code is doing
Expose the solver parameter which defaults to gurobi. This is passed to both the server and container "launch" methods for consistency.

### Testing
Built the container and tested the solver argument in "container mode". This didn't invoke the `_launch_on_server` overload but it's essentially copy paste (same logic is used by the [api](https://github.com/Breakthrough-Energy/REISE.jl/blob/develop/pyreisejl/utility/app.py#L38) when launched via http)

### Time estimate
5 min

### DIFF
--- a/powersimdata/scenario/execute.py
+++ b/powersimdata/scenario/execute.py
@@ -173,11 +173,13 @@ class Execute(State):
                 f"Status must be one of {valid_status}, but got status={self._scenario_status}"
             )
 
-    def _launch_on_server(self, threads=None, extract_data=True):
+    def _launch_on_server(self, threads=None, solver=None, extract_data=True):
         """Launch simulation on server, via ssh.
 
         :param int/None threads: the number of threads to be used. This defaults to None,
             where None means auto.
+        :param str solver: the solver used for optimization. This defaults to
+            None, which translates to gurobi
         :param bool extract_data: whether the results of the simulation engine should
             automatically extracted after the simulation has run. This defaults to True.
         :raises TypeError: if extract_data is not a boolean
@@ -189,6 +191,9 @@ class Execute(State):
             # Use the -t flag as defined in call.py in REISE.jl
             extra_args.append("--threads " + str(threads))
 
+        if solver:
+            extra_args.append("--solver", solver)
+
         if not isinstance(extract_data, bool):
             raise TypeError("extract_data must be a boolean: 'True' or 'False'")
         if extract_data:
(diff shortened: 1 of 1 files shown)


powersimdata/scenario/execute.py
```diff
@@ -189,36 +191,43 @@ def _launch_on_server(self, threads=None, extract_data=True):
             # Use the -t flag as defined in call.py in REISE.jl
             extra_args.append("--threads " + str(threads))
 
+        if solver:
+            extra_args.append("--solver", solver)
```
rouille, line 195: Since we test the type of `extract_data` should we do the same with `solver` and `threads`?

jon-hagg: We check `threads` just before calling this, since it's also passed to `_launch_in_container`. I can add the check for `solver` though

powersimdata/scenario/execute.py
```diff
@@ -189,54 +191,70 @@ def _launch_on_server(self, threads=None, extract_data=True):
...
         :param int/None threads: the number of threads to be used. This defaults to None,
             where None means auto.
         :param bool extract_data: whether the results of the simulation engine should
             automatically extracted after the simulation has run. This defaults to True.
+        :param str solver: the solver used for optimization. This defaults to
+            None, which translates to gurobi
         :raises TypeError: if threads is not an int
-        :raises ValueError: if threads is not a positive value
+        :raises ValueError: if threads is not a positive value or invalid
+            solver provided
```
rouille, line 245: You can remove all the exceptions in the docstring

powersimdata/scenario/execute.py
```diff
@@ -189,54 +191,70 @@ def _launch_on_server(self, threads=None, extract_data=True):
...
-        resp = requests.post(url, params={"threads": threads})
+        resp = requests.post(url, params={"threads": threads, "solver": solver})
         if resp.status_code != 200:
             print(
                 f"Failed to launch simulation: status={resp.status_code}. See response for details"
             )
         return resp
 
-    def launch_simulation(self, threads=None, extract_data=True):
+    def _check_threads(self, threads):
```
rouille, line 222: missing docstring

powersimdata/scenario/execute.py
```diff
@@ -189,54 +191,70 @@ def _launch_on_server(self, threads=None, extract_data=True):
...
+        if threads:
+            if not isinstance(threads, int):
+                raise TypeError("threads must be an int")
+            if threads < 1:
+                raise ValueError("threads must be a positive value")
+
+    def _check_solver(self, solver):
```
rouille, line 229: missing docstring

powersimdata/scenario/execute.py
jon-hagg: Yeah I know usually we limit the `raises` to the local exceptions, but felt like in this case it's clearer to keep it as if the `_check_*` methods are just an extension of this one (and it's unlikely the docstring on those would be useful, since I'd expect a user to do `help(scenario.state.launch_simulation)`. Kinda gray area though, what do you think? I can stick with the pattern if that's preferable.

rouille: Not sure. If you do help(scenario.state.launch_simulation) you will see the expected type of the parameters through `:param str solver:` so you are not completely in the dark.

jon-hagg: That's true. Anyway, I updated it to be consistent since I think that's enough of a reason on its own.

rouille (approved):
Looks good. Thanks.

PR #405, opened 2021-02-24
Modify schema of scenario list

rouille:

### Purpose
Update the schema of the scenario list (csv and sql)

### What the code is doing
Add an extra field `grid_model`

### Testing

### Where to look
Look for an extra field named `grid_model` in the scenario list template.

### Time estimate
2min

### DIFF
--- a/powersimdata/data_access/sql/schema.sql
+++ b/powersimdata/data_access/sql/schema.sql
@@ -9,6 +9,7 @@ CREATE TABLE IF NOT EXISTS scenario_list(
   plan text not null,
   name text not null,
   state text not null,
+  grid_model text not null,
   interconnect text not null,
   base_demand text not null,
   base_hydro text not null,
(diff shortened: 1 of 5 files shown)


jon-hagg (approved):

powersimdata/data_access/sql/schema.sql
```diff
@@ -20,5 +20,6 @@ CREATE TABLE IF NOT EXISTS scenario_list(
   interval text not null,
   engine text not null,
   runtime text,
-  infeasibilities text
+  infeasibilities text,
+  grid_model text
```
jon-hagg, line 24: Might want this to be `not null` unless we have a default, but it's fine for now

rouille: Done

powersimdata/utility/templates/ScenarioList.csv
```diff
@@ -1 +1 @@
-id,plan,name,state,interconnect,base_demand,base_hydro,base_solar,base_wind,change_table,start_date,end_date,interval,engine,runtime,infeasibilities
+id,plan,name,state,interconnect,base_demand,base_hydro,base_solar,base_wind,change_table,start_date,end_date,interval,engine,runtime,infeasibilities,grid_model
```
danielolsen, line 1: Should the new `grid_model` column live next to the `interconnect` column? I think we factored out all the hard-coded column numbers, so changing the position of the other columns shouldn't matter.

rouille: As you wish. I will have to switch the order in the Create class.

rouille: Done

danielolsen (approved):

PR #407, opened 2021-02-25
fix: correct bug with carried-over gencosts for new plants

danielolsen:
### Purpose
Fix the bug, where new renewable generators added after a thermal generator inherit the thermal generator's cost curve coefficients.

### What the code is doing
The first commit reproduces the code in a test, the second commit fixes it by ensuring that no values are carried over.

### Time estimate
1 minute.

### DIFF
--- a/powersimdata/input/tests/test_transform_grid.py
+++ b/powersimdata/input/tests/test_transform_grid.py
@@ -446,6 +446,7 @@ def test_add_gen_add_entries_in_gencost_data_frame(ct):
             "c1": 50,
             "c2": 0.5,
         },
+        {"type": "solar", "bus_id": 2050363, "Pmax": 15},
     ]
     ct.add_plant(new_plant)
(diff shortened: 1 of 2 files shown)


BainanXia (approved):
Good catch!

rouille:
Should we move in the `for` loop the other too (`new_branch`, `new_plant`, etc.)?

danielolsen:
> Should we move in the `for` loop the other too (`new_branch`, `new_plant`, etc.)? I mean we need to set the value of the columns to 0 for all of them in principle.

The problem that is specific to gencost is the `if` block which updates those entries only sometimes. It should not be a problem for the other loops which do not have this feature, but it is probably good practice to move them all, to avoid this issue potentially arising later.

danielolsen:
The analogous line has been moved inside the loops in the other add methods.

rouille (approved):
Makes sense

PR #409, opened 2021-03-09
Refactor mock objects

rouille:

### Purpose
Add grid model information in mock objects

### What the code is doing
The scenario info dictionary is updated in the `MockScenario` class and the `Grid` class can be instantiated with an optional `model` parameter (default to `usa_tamu`)

### Testing
All tests pass

### Where to look
`MockGrid` and `MockScenario` classes

### Time estimate
5 min


BainanXia (approved):
Looks good.

PR #411, opened 2021-03-11
fix/refactor: return grid model for ScenarioGrid, reduce repeated get_grid_model calls

danielolsen:
### Purpose
Return grid model for ScenarioGrid, reduce repeated get_grid_model calls

### What the code is doing
The first commit in **grid.py** fixes a bug where the grid model is determined but not returned.
The next commits in **clean_capacity_scaling.py** and **upgrade.py** move the `Grid.get_grid_model()` call out of a loop, since every call to `get_grid_model()` for a `ScenarioGrid` requires a fresh download of the ScenarioList to complete the `ScenarioListManager.get_scenario` call.

### Time estimate
5 minutes.

### DIFF
--- a/powersimdata/design/transmission/upgrade.py
+++ b/powersimdata/design/transmission/upgrade.py
@@ -175,8 +175,9 @@ def get_branches_by_area(grid, area_names, method="either"):
 
     branch = grid.branch
     selected_branches = set()
+    grid_model = grid.get_grid_model()
     for a in area_names:
-        load_zone_names = area_to_loadzone(grid.get_grid_model(), a)
+        load_zone_names = area_to_loadzone(grid_model, a)
         to_bus_in_area = branch.to_zone_name.isin(load_zone_names)
         from_bus_in_area = branch.from_zone_name.isin(load_zone_names)
         if method in ("internal", "either"):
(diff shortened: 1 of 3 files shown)


powersimdata/input/grid.py
```diff
@@ -69,7 +69,7 @@ def get_grid_model(self):
         :return: (*str*).
         """
         if os.path.isfile(self.data_loc):
-            _get_grid_model_from_scenario_list(self.data_loc)
+            return _get_grid_model_from_scenario_list(self.data_loc)
```
rouille, line 72: indeed

powersimdata/design/transmission/upgrade.py
```diff
@@ -175,8 +175,9 @@ def get_branches_by_area(grid, area_names, method="either"):
 
     branch = grid.branch
     selected_branches = set()
+    grid_model = grid.get_grid_model()
     for a in area_names:
-        load_zone_names = area_to_loadzone(grid.get_grid_model(), a)
+        load_zone_names = area_to_loadzone(grid_model, a)
```
rouille, line 180: good catch

rouille (approved):
Thanks for debugging

PR #416, opened 2021-03-13
Manage scenario/execute list using pandas

jon-hagg:
### Purpose
Supports local usage on windows by limiting the assumption of a unix environment to the `SSHDataAccess` class (in the context of scenario/execute list only; there is more work to do for full windows support)

### What the code is doing
Implements the following pattern for scenario/execute list operations. For local environments: just modify and query using pandas. For the client/server environment:
* calculate `sha1sum` prior to download and save in a local variable
* download the file and modify as needed
* upload the file
* using `flock` to synchronize access, check if the current `sha1sum` of the file is the same as before, and if so, write the changes and create a backup (the `-b` option on `mv` command). 

The client/server operation basically extends the local variant by adding the integrity check, so we get some code reuse out of it. Other changes are mostly supporting functionality, with a few minor improvements thrown in.

### Testing
* Added unit tests (originally in plug, but realized they could go here). These test against an empty csv with headers, copied from the `powersimdata/utility/templates` for each test. 
* Tested in the standalone container setup by running a simulation from scratch
* Tested in the client/server container setup by creating and preparing a scenario (it's not set up to run one yet), as well as setting a breakpoint during the modification and editing the file in the server container to hit the error state

### Usage Example/Visuals
I could add some usage output if it's useful

### Time estimate
30 min

### DIFF
--- a/powersimdata/data_access/execute_list.py
+++ b/powersimdata/data_access/execute_list.py
@@ -135,7 +135,7 @@ class ExecuteListManager(CsvStore):
         """
         table = self.get_execute_table()
         scenario_id = int(scenario_info["id"])
-        table.drop(scenario_id)
+        table.drop(scenario_id, inplace=True)
 
         print("--> Deleting entry in execute table on server")
         return table
--- a/powersimdata/data_access/execute_list.py
+++ b/powersimdata/data_access/execute_list.py
@@ -1,3 +1,4 @@
+import os
 import posixpath
 
 from powersimdata.data_access.csv_store import CsvStore
@@ -50,15 +51,15 @@ class ExecuteTable(SqlStore):
             ),
         )
 
-    def update_execute_list(self, status, scenario_info):
+    def set_status(self, scenario_id, status):
         """Updates status of scenario in execute list
 
+        :param int scenario_id: the scenario id
         :param str status: execution status.
-        :param collections.OrderedDict scenario_info: entry to update
         """
         self.cur.execute(
             "UPDATE execute_list SET status = %s WHERE id = %s",
-            (status, scenario_info["id"]),
+            (status, scenario_id),
         )
 
     def delete_entry(self, scenario_info):
@@ -85,6 +86,13 @@ class ExecuteListManager(CsvStore):
             server_setup.DATA_ROOT_DIR, self._EXECUTE_LIST
         )
 
+    def _save_file(self, table):
+        """Save to local directory
+
+        :param pandas.DataFrame table: the execute list data frame
+        """
+        table.to_csv(os.path.join(server_setup.LOCAL_DIR, self._EXECUTE_LIST))
+
     def get_execute_table(self):
         """Returns execute table from server if possible, otherwise read local
         copy. Updates the local copy upon successful server connection.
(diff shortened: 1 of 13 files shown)


powersimdata/data_access/scenario_list.py
```diff
@@ -148,34 +138,31 @@ def err_message(text):
...
-            scenario_info["id"],
-            entry,
-        )
-        command = "awk %s %s %s" % (options, program, self._server_path)
-
-        err_message = "Failed to add entry in %s on server" % self._SCENARIO_LIST
-        _ = self._execute_and_check_err(command, err_message)
-
-    def delete_entry(self, scenario_info):
+        table = self.get_scenario_table()
```
jon-hagg, line 148: This seemed like it should be easier to do in pandas, but it's the only way I got it to work. Let me know if you have suggestions.

powersimdata/data_access/execute_list.py
```diff
@@ -107,39 +105,36 @@ def get_status(self, scenario_id):
...
-
-    def delete_entry(self, scenario_info):
+        table = self.get_execute_table()
+        table.loc[int(scenario_id), "status"] = status
+
+        print(f"-->  Setting status={status} in execute table on server")
+        return table
+
+    @verify_hash
+    def delete_entry(self, scenario_id):
```
jon-hagg, line 130: I noticed we don't actually have code calling the delete method (ditto for the scenario list), but figured it might be useful to keep?

rouille: It should be called in the `powersimdata.scenario.delete` module

jon-hagg: Well that makes sense. Not sure how I missed it

rouille: `grep -R "delete_entry" powersimdata`

rouille:
We probably want to include a test for the `Delete` class (see README for an example) in the standalone container setup.

powersimdata/data_access/tests/test_execute_csv.py
```diff
@@ -1,10 +1,16 @@
...
+import shutil
+from collections import OrderedDict
+
 import pandas as pd
 import pytest
 from numpy.testing import assert_array_equal
 from pandas.testing import assert_frame_equal
 
-from powersimdata.data_access.data_access import SSHDataAccess
+import powersimdata.utility.templates as templates
```
danielolsen, line 10: `from powersimdata.utility import templates`?

powersimdata/data_access/tests/test_scenario_csv.py
```diff
@@ -1,10 +1,16 @@
...
+import shutil
+from collections import OrderedDict
+
 import pandas as pd
 import pytest
 from numpy.testing import assert_array_equal
 from pandas.testing import assert_frame_equal
 
-from powersimdata.data_access.data_access import SSHDataAccess
+import powersimdata.utility.templates as templates
```
danielolsen, line 10: `from powersimdata.utility import templates`?

danielolsen:
I'll run a test on the server this afternoon.

danielolsen:
I get the following traceback trying to call `Create.create_scenario()`:
```python
>>> scenario.state.create_scenario()
CREATING SCENARIO: test | pandas

Transferring ScenarioList.csv from server
100%|#######################################| 542k/542k [00:00<00:00, 3.65Mb/s]
Transferring ScenarioList.csv from server
100%|#######################################| 542k/542k [00:00<00:00, 4.24Mb/s]
--> Adding entry in ScenarioList.csv on server
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 205, in create_scenario
    self._scenario_list_manager.add_entry(self._scenario_info)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\data_access\csv_store.py", line 20, in wrapper
    self.commit(table, checksum)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\data_access\csv_store.py", line 76, in commit
    self.data_access.push(self._FILE_NAME, checksum)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\data_access\data_access.py", line 349, in push
    self.move_to(file_name, change_name_to=backup, preserve=True)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\data_access\data_access.py", line 297, in move_to
    self._check_file_exists(to_path, should_exist=False)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\data_access\data_access.py", line 68, in _check_file_exists
    raise OSError(f"{filepath} {msg} on server")
OSError: /mnt/bes/pcm/ScenarioList.csv.bak already exists on server
```

jon-hagg:
I think if we remove the file it should work - the approach here is to upload the file as `FILE.csv.bak` and if the checksum matches, rename it to `FILE.csv` and create `FILE.csv~` as the backup. So  the `.bak` file should only be temporary, unless there is a conflict which has to be resolved manually.

danielolsen:
After @jon-hagg's tweaks, I've successfully created, launched, extracted, and deleted a scenario on the server.

rouille (approved):
Thanks!

PR #419, opened 2021-03-18
Download profiles from blob storage

jon-hagg:
### Purpose
Use blob storage as the source for profiles, downloading as needed. 

### What the code is doing
Added a [version.json](https://bescienceswebsite.blob.core.windows.net/profiles/version.json) file to support listing available versions. The profiles are in the `raw/usa_tamu` folder in the blob container to mimic the structure elsewhere. Similar to before, we download if it doesn't exist locally. 

### Testing
Deleted local copies of the profiles and ran a simulation using plug, which shows the files being downloaded (during the call to `prepare_simulation_input`). 

### Usage Example/Visuals
Example query (no change in usage, other than switching to static method)
```
In [18]: InputData.get_profile_version("usa_tamu", "solar")
Out[18]: ['vJan2021']
```

Snippet of files being downloaded
```
In [7]: scenario.state.prepare_simulation_input()
---------------------------
PREPARING SIMULATION INPUTS
---------------------------
--> Creating temporary folder on server for simulation inputs
--> Loading demand
demand_vJan2021.csv not found in /root/ScenarioData/ on local machine
--> Downloading demand_vJan2021.csv from blob storage.
100%|████████████████████████████████████████████████████████████████████████████| 5.15M/5.15M [00:00<00:00, 6.17MB/s]
--> Done!
Multiply demand in Far West (#301) by 1.09
/usr/local/lib/python3.8/site-packages/pandas/core/indexing.py:1843: SettingWithCopyWarning:
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  self.obj[item_labels[indexer[info_axis]]] = value
Multiply demand in North (#302) by 1.09
Multiply demand in West (#303) by 1.09
Multiply demand in South (#304) by 1.09
Multiply demand in North Central (#305) by 1.09
Multiply demand in South Central (#306) by 1.09
Multiply demand in Coast (#307) by 1.09
Multiply demand in East (#308) by 1.09
Writing scaled demand profile in /root/ScenarioData/ on local machine
--> Moving file /root/ScenarioData/1_demand.csv to /mnt/bes/pcm/tmp/scenario_1/demand.csv
--> Deleting original copy
--> Loading hydro
hydro_vJan2021.csv not found in /root/ScenarioData/ on local machine
--> Downloading hydro_vJan2021.csv from blob storage.
100%|██████████████████████████████████████████████████████████████████████████████| 222M/222M [00:40<00:00, 5.74MB/s]
--> Done!
```

### Time estimate
20 mins

### DIFF
--- a/powersimdata/data_access/data_access.py
+++ b/powersimdata/data_access/data_access.py
@@ -1,4 +1,3 @@
-import json
 import operator
 import os
 import posixpath
@@ -203,18 +202,9 @@ class LocalDataAccess(DataAccess):
         :param str kind: *'demand'*, *'hydro'*, *'solar'* or *'wind'*.
         :return: (*list*) -- available profile version.
         """
-        blob_versions = super().get_profile_version(grid_model, kind)
-        version_file = os.path.join(server_setup.LOCAL_DIR, "version.json")
-        if not os.path.exists(version_file):
-            return blob_versions
-        with open(version_file) as f:
-            version = json.load(f)
-            return list(
-                set(
-                    blob_versions
-                    + ProfileHelper.parse_version(grid_model, kind, version)
-                )
-            )
+        blob_version = super().get_profile_version(grid_model, kind)
+        local_version = ProfileHelper.get_profile_version_local(grid_model, kind)
+        return list(set(blob_version + local_version))
 
 
 class SSHDataAccess(DataAccess):
--- a/powersimdata/data_access/data_access.py
+++ b/powersimdata/data_access/data_access.py
@@ -117,6 +117,12 @@ class DataAccess:
         raise NotImplementedError
 
     def get_profile_version(self, grid_model, kind):
+        """Returns available raw profile from blob storage
+
+        :param str grid_model: grid model.
+        :param str kind: *'demand'*, *'hydro'*, *'solar'* or *'wind'*.
+        :return: (*list*) -- available profile version.
+        """
         return ProfileHelper.get_profile_version_cloud(grid_model, kind)
 
     def close(self):
--- a/powersimdata/data_access/data_access.py
+++ b/powersimdata/data_access/data_access.py
@@ -117,7 +117,7 @@ class DataAccess:
         raise NotImplementedError
 
     def get_profile_version(self, grid_model, kind):
-        return ProfileHelper.get_profile_version(grid_model, kind)
+        return ProfileHelper.get_profile_version_cloud(grid_model, kind)
 
     def close(self):
         """Perform any necessary cleanup for the object."""
--- a/powersimdata/data_access/data_access.py
+++ b/powersimdata/data_access/data_access.py
@@ -1,3 +1,4 @@
+import json
 import operator
 import os
 import posixpath
@@ -7,6 +8,7 @@ from subprocess import PIPE, Popen
 import paramiko
 from tqdm import tqdm
 
+from powersimdata.data_access.profile_helper import ProfileHelper
 from powersimdata.utility import server_setup
 from powersimdata.utility.helpers import CommandBuilder
 
@@ -115,6 +117,9 @@ class DataAccess:
         """
         raise NotImplementedError
 
+    def get_profile_version(self, grid_model, kind):
+        return ProfileHelper.get_profile_version(grid_model, kind)
+
     def close(self):
         """Perform any necessary cleanup for the object."""
         pass
@@ -190,6 +195,26 @@ class LocalDataAccess(DataAccess):
             text=True,
         )
         return wrap(None), wrap(proc.stdout), wrap(proc.stderr)
+
+    def get_profile_version(self, grid_model, kind):
+        """Returns available raw profile from blob storage or local disk
+
+        :param str grid_model: grid model.
+        :param str kind: *'demand'*, *'hydro'*, *'solar'* or *'wind'*.
+        :return: (*list*) -- available profile version.
+        """
+        blob_versions = super().get_profile_version(grid_model, kind)
+        version_file = os.path.join(server_setup.LOCAL_DIR, "version.json")
+        if not os.path.exists(version_file):
+            return blob_versions
+        with open(version_file) as f:
+            version = json.load(f)
+            return list(
+                set(
+                    blob_versions
+                    + ProfileHelper.parse_version(grid_model, kind, version)
+                )
+            )
 
 
 class SSHDataAccess(DataAccess):
--- a/powersimdata/data_access/profile_helper.py
+++ b/powersimdata/data_access/profile_helper.py
@@ -48,13 +48,7 @@ class ProfileHelper:
         :param str kind: *'demand'*, *'hydro'*, *'solar'* or *'wind'*.
         :param dict version: json response
         :return: (*list*) -- available profile version.
-        :raises ValueError: if kind not one of *'demand'*, *'hydro'*, *'solar'* or
-            *'wind'*.
         """
-        profile_kind = {"demand", "hydro", "solar", "wind"}
-        if kind not in profile_kind:
-            raise ValueError("kind must be one of %s" % " | ".join(profile_kind))
-
         if grid_model in version and kind in version[grid_model]:
             return version[grid_model][kind]
         print("No %s profiles available." % kind)
--- a/powersimdata/data_access/profile_helper.py
+++ b/powersimdata/data_access/profile_helper.py
@@ -12,6 +12,13 @@ class ProfileHelper:
 
     @staticmethod
     def get_file_components(scenario_info, field_name):
+        """Get the file name and relative path for the given profile and
+        scenario.
+
+        :param dict scenario_info: a ScenarioInfo instance
+        :param str field_name: the kind of profile
+        :return: (*tuple*) -- file name and path
+        """
         version = scenario_info["base_" + field_name]
         file_name = field_name + "_" + version + ".csv"
         grid_model = scenario_info["grid_model"]
@@ -20,6 +27,12 @@ class ProfileHelper:
 
     @staticmethod
     def download_file(file_name, from_dir):
+        """Download the profile from blob storage at the given path
+
+        :param str file_name: profile csv
+        :param str from_dir: the path relative to the blob container
+        :return: (*str*) -- path to downloaded file
+        """
         print(f"--> Downloading {file_name} from blob storage.")
         url = f"{ProfileHelper.BASE_URL}/{from_dir}/{file_name}"
         dest = os.path.join(server_setup.LOCAL_DIR, from_dir, file_name)
--- a/powersimdata/data_access/profile_helper.py
+++ b/powersimdata/data_access/profile_helper.py
@@ -1,0 +1,72 @@
+import os
+
+import requests
+from tqdm.auto import tqdm
+
+from powersimdata.utility import server_setup
+
+
+class ProfileHelper:
+    BASE_URL = "https://bescienceswebsite.blob.core.windows.net/profiles"
+
+    @staticmethod
+    def get_file_components(scenario_info, field_name):
+        version = scenario_info["base_" + field_name]
+        file_name = field_name + "_" + version + ".csv"
+        grid_model = scenario_info["grid_model"]
+        from_dir = f"raw/{grid_model}"
+        return file_name, from_dir
+
+    @staticmethod
+    def download_file(file_name, from_dir):
+        print(f"--> Downloading {file_name} from blob storage.")
+        url = f"{ProfileHelper.BASE_URL}/{from_dir}/{file_name}"
+        dest = os.path.join(server_setup.LOCAL_DIR, from_dir, file_name)
+        os.makedirs(os.path.dirname(dest), exist_ok=True)
+        resp = requests.get(url, stream=True)
+        content_length = int(resp.headers.get("content-length", 0))
+        with open(dest, "wb") as f:
+            with tqdm(
+                unit="B",
+                unit_scale=True,
+                unit_divisor=1024,
+                miniters=1,
+                total=content_length,
+            ) as pbar:
+                for chunk in resp.iter_content(chunk_size=4096):
+                    f.write(chunk)
+                    pbar.update(len(chunk))
+
+        print("--> Done!")
+        return dest
+
+    @staticmethod
+    def parse_version(grid_model, kind, version):
+        """Parse available versions from the given spec
+
+        :param str grid_model: grid model.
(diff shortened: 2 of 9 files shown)


rouille:
Two comments about the code snippet:
* `--> Creating temporary folder on server for simulation inputs`. I believe now we have other installation possible sides the client/server one, I guess we can remove the `on server` in the printout.
* there is a warning:
```
Multiply demand in Far West (#301) by 1.09
/usr/local/lib/python3.8/site-packages/pandas/core/indexing.py:1843: SettingWithCopyWarning:
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead
```
I believed it is raised in the `_get_demand_profile` method of the `TransformProfile` class. I remember seeing it before. but I rarely run scenarios. Does it ring any bell @danielolsen ?

danielolsen:
@rouille I'm familiar with that warning, it's something that's not very hard to track down and fix.

powersimdata/input/input_data.py
```diff
@@ -12,6 +13,77 @@
...
+
+
+class ProfileHelper:
+    @staticmethod
+    def get_file_components(scenario_info, field_name):
+        ext = _file_extension[field_name]
+        version = scenario_info["base_" + field_name]
+        file_name = field_name + "_" + version + "." + ext
+        grid_model = scenario_info["grid_model"]
+        from_dir = f"{server_setup.BASE_PROFILE_DIR}/{grid_model}"
```
danielolsen, line 47: Is this a posixpath assumption?

danielolsen: Or is this a URL, and the folder name matches the `server_setup` but we're not actually using the server?

rouille: I would say that we don't need a BASE_PROFILE_DIR now that hat we fetch from the blob storage

danielolsen:
I think the warning can be fixed by changing [L.106 of transform_profile.py](https://github.com/Breakthrough-Energy/PowerSimData/blob/develop/powersimdata/input/transform_profile.py#L106) from:
```
demand = self._input_data.get_data(self.scenario_info, "demand")[zone_id]
```
to:
```
demand = self._input_data.get_data(self.scenario_info, "demand").loc[:, zone_id]
```

jon-hagg:
> I think the warning can be fixed by changing [L.106 of transform_profile.py](https://github.com/Breakthrough-Energy/PowerSimData/blob/develop/powersimdata/input/transform_profile.py#L106) 

Nice, that works - no warning when I re-ran the same test.

powersimdata/input/input_data.py
```diff
@@ -96,20 +147,11 @@ def get_profile_version(self, grid_model, kind):
...
-        )
-        stdin, stdout, stderr = self.data_access.execute_command("ls " + query)
-        if len(stderr.readlines()) != 0:
-            print("No %s profiles available." % kind)
-            version = []
-        else:
-            filename = [os.path.basename(line.rstrip()) for line in stdout.readlines()]
-            version = [f[f.rfind("_") + 1 : -4] for f in filename]
-        return version
+        resp = requests.get(f"{BASE_URL}/version.json")
```
ahurli, line 150: Do we want users to be able to use their own profiles that only exist locally? I'm thinking that in the previous Docker implementation, users could just add a new profile (like a new demand profile to test) to the `ScenarioData`directory and be able to use it, whereas now they'll either need to add it to our blob storage or create their own.

rouille: I think we are just downloading a JSON files enclosing the profile version here. Otherwise the logic that we had before is still in place, i.e., we only download if there is not local copy. Also in ***plug***, I would say the user can have its own profile and use them in the various script via the `set_base_profile` method.

danielolsen: What would get added to the ScenarioList if the user brought their own profile? "untracked"?

jon-hagg: Good point, there isn't a natural way to use a custom profile if we are strictly checking the version.json in blob storage. I think we'd need to do something like cache that file locally so it can be edited, or check both a local copy and the one in blob storage, etc. The profiles can still be put in the ScenarioData folder locally, but they would have a version that is considered valid to actually use it.

jon-hagg: > What would get added to the ScenarioList if the user brought their own profile? "untracked"?

Yeah it would get added to the user's ScenarioList. So their environment would be consistent, but couldn't be easily "merged" with another one (e.g. our server environment). I was going to suggest something like adding full references to files in the scenario list, but can't do much if the sources aren't public, e.g. in some other blob storage. Not sure if this is an issue at this point?

rouille: I would say you can use your own profile if they are stored in ***ScenarioData/raw/GRID_MODEL/*** with the following drawbacks:
* they need to be named ***demand_SOMETHING.cs*** or hydro/solar/wind
* the prompt won't tell they are available
* if they are erased you won't be able to recover you scenario

jon-hagg: Right, only thing is the local version would have to match the values in blob storage, otherwise the call to `Create.set_base_profile` would fail. I think this comes down to having `InputData.get_profile_version` return local options as well so a user doesn't have to trick the software.

rouille: You are right @jon-hagg, I forgot that we can only assign a profile if the version is in the returned list of available version:
```
    def set_base_profile(self, kind, version):
        """Sets demand profile.

        :param str kind: one of *'demand'*, *'hydro'*, *'solar'*, *'wind'*.
        :param str version: demand profile version.
        :raises ValueError: if no profiles are available or version is not available.
        """
        possible = self.get_base_profile(kind)
        if len(possible) == 0:
            raise ValueError("No %s profile available" % kind)
        elif version in possible:
            if kind == "demand":
                self.demand = version
            if kind == "hydro":
                self.hydro = version
            if kind == "solar":
                self.solar = version
            if kind == "wind":
                self.wind = version
        else:
            raise ValueError("Available %s profiles: %s" % (kind, " | ".join(possible)))
```

rouille: It is not a problem specific to the blob storage, we had the same issue before with the profiles stored on the server. No?

jon-hagg: For client/server setup that's true. But interestingly in docker we are using `LocalDataAccess` to list the available profiles, so it would have found anything a user added to their local folder.

ahurli: Yeah, I don't think this is a feature that has been widely used, but I could see it being useful for collaborators going forward to test things out before releasing "official" profiles.

I was also thinking along the same lines as Jon's solution to have `get_profile_version` return both local and blob storage profiles, making it clear which are our profiles vs local-only and maybe some warnings about using unsupported profiles.

rouille: That would be nice. How would we handle that in the client/server configuration? A user would be able to use its own profile, edit the ***ScenarioList*** file file with its profile version but none of the other user would be able to load this scenario because they don't have access to the profiles.

ahurli: Since `DataAccess` differentiates between configurations, we could add something like `data_access.get_local_profiles(kind)` that returns nothing for `SSHDataAccess` but returns any additional profiles for `LocalDataAccess`. We could call and append the results in `InputData`.

Depending on how fancy we want to get, we could also pass in the profiles already found in blob storage to both a) mark which profiles have been cached if a user wants to be careful of their internet usage and b) filter out only the extra profiles not in blob storage.

jon-hagg: Cool, I ended up thinking the same thing (my other ideas didn't pan out). Not sure I understand the part about passing in the profiles though?

ahurli: I was first thinking of passing in the profiles like `data_access.get_local_profiles(kind, blob_profiles)`, but looking at it now, there's no reason we can't do that filtering/marking at the `InputData` level instead of at the `DataAccess` level.

ahurli: Actually, sorry, I remembered what I was thinking. You would want to pass in the blob profiles so that in the `SSHDataAccess` instance, you could return the locally cached profiles but exclude anything not in the blob profiles list. In the `LocalDataAccess` instance, you would return both cached and local-only profiles.

jon-hagg: Ah thanks, that makes sense. Let me push a commit with how I'm thinking of implementing that.

powersimdata/data_access/profile_helper.py
```diff
@@ -0,0 +1,72 @@
...
+        :param str grid_model: grid model.
+        :param str kind: *'demand'*, *'hydro'*, *'solar'* or *'wind'*.
+        :param dict version: json response
+        :return: (*list*) -- available profile version.
+        :raises ValueError: if kind not one of *'demand'*, *'hydro'*, *'solar'* or
+            *'wind'*.
+        """
+        profile_kind = {"demand", "hydro", "solar", "wind"}
+        if kind not in profile_kind:
+            raise ValueError("kind must be one of %s" % " | ".join(profile_kind))
```
rouille, line 56: Not sure we need this check since it is already done in the `get_data` method of the `InputData` class:
```
if field_name in profile_kind:
    helper = ProfileHelper
```

rouille:
I think with 1f9291f, we have the best of both worlds. Did you have a chance to test it with both the client/server installation and the containers?

jon-hagg:
> I think with [1f9291f](https://github.com/Breakthrough-Energy/PowerSimData/commit/1f9291f6ad89d6c56c5d6f56f50575c397cc164a), we have the best of both worlds. Did you have a chance to test it with both the client/server installation and the containers?

So far only a quick test in the container, before and after adding a local version.json. That worked as expected, and the client/server behavior should be a subset of that (only checks blob storage). Planning to do a bit more, and look into adding unit tests, which was the original reason for creating the `parse_version` without the http request.

powersimdata/data_access/profile_helper.py
```diff
@@ -46,12 +47,13 @@ def parse_version(grid_model, kind, version):
...
+        :param dict version: version information per grid model
         :return: (*list*) -- available profile version.
         """
         if grid_model in version and kind in version[grid_model]:
             return version[grid_model][kind]
         print("No %s profiles available." % kind)
+        return []
 
     @staticmethod
     def get_profile_version(grid_model, kind):
```
rouille, line 59: Should this function be rename `get_profile_version_cloud`?

powersimdata/data_access/data_access.py
```diff
@@ -115,6 +116,9 @@ def push(self, file_name, checksum):
         """
         raise NotImplementedError
 
+    def get_profile_version(self, grid_model, kind):
```
rouille, line 119: Missing docstring

powersimdata/data_access/data_access.py
```diff
@@ -115,6 +116,9 @@ def push(self, file_name, checksum):
...
+        return ProfileHelper.get_profile_version_cloud(grid_model, kind)
```
rouille, line 120: We implicitly return the versions stored on the cloud. I see how it is useful but is that intuitive in the `DataAccess` class? Should we return instead an instance of `ProfileHelper` and call the functions in the child classes?

jon-hagg: I was thinking since blob storage is the source of truth, it makes sense as the default. Another way would be to `raise NotImplemented` here and have the child classes contain their own implementations (`SSHDataAccess` is just inheriting this, while `LocalDataAccess` appends the local versions if there are any). Would that be more intuitive?

powersimdata/data_access/data_access.py
jon-hagg: Ah right, I'll add the other missing ones as well

powersimdata/data_access/data_access.py
rouille: You are right. It makes sense it is the default.

powersimdata/data_access/profile_helper.py
```diff
@@ -0,0 +1,84 @@
...
+    @staticmethod
+    def get_profile_version_local(grid_model, kind):
+        """Returns available raw profile from local file
+
+        :param str grid_model: grid model.
+        :param str kind: *'demand'*, *'hydro'*, *'solar'* or *'wind'*.
+        :return: (*list*) -- available profile version.
+        """
+
+        version_file = os.path.join(server_setup.LOCAL_DIR, "version.json")
```
ahurli, line 92: Would it be easier just to list the files in the directory and filter out the ones that match the {kind}_{version}.csv format? Then again, making a user add a new profile to `version.json` is more intentional, so they're less likely to accidentally add something without understanding the risks.

powersimdata/data_access/profile_helper.py
```diff
@@ -0,0 +1,84 @@
...
+
+class ProfileHelper:
+    BASE_URL = "https://bescienceswebsite.blob.core.windows.net/profiles"
+
+    @staticmethod
+    def get_file_components(scenario_info, field_name):
+        version = scenario_info["base_" + field_name]
+        file_name = field_name + "_" + version + ".csv"
+        grid_model = scenario_info["grid_model"]
+        from_dir = f"raw/{grid_model}"
```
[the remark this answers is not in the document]
jon-hagg: Good call, I should probably do an `os.path.join` here

powersimdata/data_access/profile_helper.py
jon-hagg: Yeah this was kind of a trade off - it's less code to reuse the json format and provides at least one way for a user to customize. Figured it's ok for now, but definitely open to future improvements.

ahurli: Makes sense! We could also probably get some feedback from our users/collaborators to see what they think about usability. But agreed, this looks good for now.

powersimdata/data_access/profile_helper.py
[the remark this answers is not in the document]
ahurli: I should probably get this on my personal Windows machine to test, but I think we might run into problems calling the blob storage then. We're using the same `from_dir` for both the url and local directory in the `download_file` method.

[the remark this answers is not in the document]
jon-hagg: D'oh, will fix. We can look into automating windows tests, maybe in github actions or using windows containers.

ahurli (approved):
Looks good!

rouille (approved):
Thanks. This is a great feature.

PR #398, opened 2021-02-16
Use grid model to create scenario

rouille:
[Pull Request Etiquette doc](https://github.com/Breakthrough-Energy/REISE/wiki/Pull-Request-Etiquette)

### Purpose
Allow the user to specify the grid model he wants to use in his scenario and access the corresponding profiles.

### What the code is doing
It instantiates the `_Builder` class with an additional parameter `grid_model`. To use the `usa_tamu` grid model for the `Western` interconnect, the user will call the `set_builder` method in the `Create` class as follows: 
```
from powersimdata.scenario.scenario import Scenario

scenario = Scenario('')
scenario.state.set_builder("usa_tamu", "Western")
```
By default `set_builder` will instantiate the `Builder` class for the full U.S. TAMU grid.

### Testing
The following script:
```
from powersimdata.scenario.scenario import Scenario

scenario = Scenario('')

# Start building a scenario
scenario.state.set_builder("usa_tamu", "Western")

# set plan and scenario names
scenario.state.builder.set_name("test", "dummy")
# set start date, end date and interval
scenario.state.builder.set_time("2016-08-01 00:00:00",
                                "2016-08-31 23:00:00",
                                "24H")
# set demand profile version
scenario.state.builder.set_base_profile("demand", "vJan2021")
# set hydro profile version
scenario.state.builder.set_base_profile("hydro", "vJan2021")
# set solar profile version
scenario.state.builder.set_base_profile("solar", "vJan2021")
# set wind profile version
scenario.state.builder.set_base_profile("wind", "vJan2021")

# scale capacity of solar plants in WA and AZ by 5 and 2.5, respectively
scenario.state.builder.change_table.scale_plant_capacity(
    "solar", zone_name={"Washington": 5, "Arizona": 2.5})
# scale capacity of wind farms in OR and MT by 1.5 and 2, respectively
scenario.state.builder.change_table.scale_plant_capacity(
    "wind", zone_name={"Oregon": 1.5, "Montana Western": 2})
# scale capacity of branches in NV and WY by 2
scenario.state.builder.change_table.scale_branch_capacity(
    zone_name={"Nevada": 2, "Wyoming": 2})
# add AC lines in NM and CO
scenario.state.builder.change_table.add_branch(
    [{"capacity": 200, "from_bus_id": 2053002, "to_bus_id": 2053303},
     {"capacity": 150, "from_bus_id": 2060002, "to_bus_id": 2060046}])
# add DC line between CO and CA (Bay Area)
[...]

PR #420, opened 2021-03-18
refactor: improve usability of Scenario object

### LINKED ISSUE #418: Simplify calls to State methods from Scenario object
Currently, any calls to methods of the Scenario's `state` attribute (an instance of a `State` child class) need to be used as e.g. `scenario.state.create_scenario()`, `scenario.state.launch_simulation()`. It would be nice if the user didn't have to think about the state, and could simply call `scenario.create_scenario()` or `scenario.launch_simulation()`, while all of the same underlying logic is maintained (i.e., calling a method for a state that the Scenario is not in still gives an error).

danielolsen:
### Purpose
Improve usability of the main framework classes (`Scenario` and `Grid`). Closes #418.

### What the code is doing
- The Scenario can now be instantiated with no arguments (`Scenario()`)
- The `Scenario` and `Grid` objects are imported into the name namespace, so that we can do something like:
```python
import powersimdata as psd
scenario = psd.Scenario()
```
- The state methods can be accessed from the Scenario object (`scenario.create_scenario()` maps to `scenario.state.create_scenario()`, if the scenario is in `create` state).

### Usage Example/Visuals
```python
>>> import powersimdata as psd
>>> scenario = psd.Scenario()
>>> scenario.create_scenario()
-------------------
MISSING INFORMATION
-------------------
plan
name
grid_model
interconnect
base_demand
base_hydro
base_solar
base_wind
change_table
start_date
end_date
interval
engine
>>> scenario.launch_simulation()
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\scenario.py", line 49, in __getattr__
    raise AttributeError(
AttributeError: Scenario object in create state has no attribute launch_simulation
```

### Time estimate
15 minutes.

### DIFF
--- a/powersimdata/scenario/create.py
+++ b/powersimdata/scenario/create.py
@@ -69,6 +69,13 @@ class Create(State):
         else:
             raise AttributeError(f"Create object has no attribute {name}")
 
+    def __setattr__(self, name, value):
+        if name in _Builder.exported_methods:
+            raise AttributeError(
+                f"{name} is exported from Create.builder, edit it there if necessary"
+            )
+        super().__setattr__(name, value)
+
     def _update_scenario_info(self):
         """Updates scenario information."""
         if self.builder is not None:
--- a/powersimdata/scenario/create.py
+++ b/powersimdata/scenario/create.py
@@ -1,5 +1,6 @@
 import copy
 import pickle
+import warnings
 from collections import OrderedDict
 
 import numpy as np
@@ -20,6 +21,7 @@ default_exported_methods = (
     "get_bus_demand",
     "print_scenario_info",
     "set_builder",
+    "set_grid",
 )
 
 
@@ -173,8 +175,15 @@ class Create(State):
         for key, val in self._scenario_info.items():
             print("%s: %s" % (key, val))
 
-    def set_builder(self, grid_model="usa_tamu", interconnect="USA"):
-        """Sets builder.
+    def set_builder(self, *args, **kwargs):
+        """Alias to set_grid."""
+        warnings.warn(
+            "set_builder is deprecated, use set_grid instead", DeprecationWarning
+        )
+        self.set_grid(*args, **kwargs)
+
+    def set_grid(self, grid_model="usa_tamu", interconnect="USA"):
+        """Sets grid builder.
 
         :param str grid_model: name of grid model. Default is *'usa_tamu'*.
         :param str/list interconnect: name of interconnect(s). Default is *'USA'*.
--- a/powersimdata/scenario/create.py
+++ b/powersimdata/scenario/create.py
@@ -62,19 +62,12 @@ class Create(State):
         super().__init__(scenario)
 
     def __getattr__(self, name):
-        if self.builder is not None:
-            if name in self.builder.exported_methods:
-                return getattr(self.builder, name)
-            else:
-                raise AttributeError(f"Create object has no attribute {name}")
+        if self.builder is not None and name in self.builder.exported_methods:
+            return getattr(self.builder, name)
+        if self.builder is None and name in _Builder.exported_methods:
+            raise AttributeError(f"Call set_grid first to access {name} attribute")
         else:
-            raise AttributeError(
-                f"Create object without a builder set has no attribute {name}. "
-                "Did you forget to run set_builder?"
-            )
-
-    def _custom_getattr_error_message(self, name):
-        return self.__getattr__(name)
+            raise AttributeError(f"Create object has no attribute {name}")
 
     def _update_scenario_info(self):
         """Updates scenario information."""
--- a/powersimdata/scenario/create.py
+++ b/powersimdata/scenario/create.py
@@ -16,14 +16,6 @@ from powersimdata.scenario.execute import Execute
 from powersimdata.scenario.state import State
 from powersimdata.utility import server_setup
 
-default_exported_methods = (
-    "create_scenario",
-    "get_bus_demand",
-    "print_scenario_info",
-    "set_builder",
-    "set_grid",
-)
-
 
 class Create(State):
     """Scenario is in a state of being created.
@@ -33,6 +25,13 @@ class Create(State):
 
     name = "create"
     allowed = []
+    default_exported_methods = (
+        "create_scenario",
+        "get_bus_demand",
+        "print_scenario_info",
+        "set_builder",
+        "set_grid",
+    )
 
     def __init__(self, scenario):
         """Constructor."""
@@ -58,7 +57,7 @@ class Create(State):
                 ("engine", ""),
             ]
         )
-        self.exported_methods = set(default_exported_methods)
+        self.exported_methods = set(self.default_exported_methods)
         super().__init__(scenario)
 
     def __getattr__(self, name):
--- a/powersimdata/scenario/create.py
+++ b/powersimdata/scenario/create.py
@@ -24,6 +24,18 @@ class Create(State):
 
     name = "create"
     allowed = []
+    exported_methods = {
+        "create_scenario",
+        "get_ct",
+        "get_grid",
+        "get_demand",
+        "get_bus_demand",
+        "get_hydro",
+        "get_solar",
+        "get_wind",
+        "print_scenario_info",
+        "set_builder",
+    }
 
     def __init__(self, scenario):
         """Constructor."""
--- a/powersimdata/scenario/create.py
+++ b/powersimdata/scenario/create.py
@@ -17,13 +17,7 @@ from powersimdata.utility import server_setup
 
 default_exported_methods = (
     "create_scenario",
-    "get_ct",
-    "get_grid",
-    "get_demand",
     "get_bus_demand",
-    "get_hydro",
-    "get_solar",
-    "get_wind",
     "print_scenario_info",
     "set_builder",
 )
@@ -120,59 +114,6 @@ class Create(State):
         file_name = self._scenario_info["id"] + "_ct.pkl"
         self._data_access.move_to(file_name, server_setup.INPUT_DIR)
 
-    def get_ct(self):
-        """Returns change table.
-
-        :return: (*dict*) -- change table.
-        :raises Exception: if :attr:`builder` has not been assigned yet through
-            meth:`set_builder`.
-        """
-        if self.builder is not None:
-            return copy.deepcopy(self.builder.change_table.ct)
-        else:
-            raise Exception("change table not set")
-
-    def get_grid(self):
-        """Returns the Grid object.
-
-        :return: (*powersimdata.input.grid.Grid*) -- a Grid object.
-        :raises Exception: if :attr:`builder` has not been assigned yet through
-            meth:`set_builder`.
-        """
-        if self.builder is not None:
-            return self.builder.get_grid()
-        else:
-            raise Exception("grid not set")
-
-    def get_profile(self, kind):
-        """Returns demand, hydro, solar or wind  profile.
-
-        :param str kind: either *'demand'*, *'hydro'*, *'solar'*, *'wind'*.
-        :return: (*pandas.DataFrame*) -- profile.
-        :raises Exception: if :attr:`builder` has not been assigned yet through
-            meth:`set_builder` or if :meth:`_Builder.set_base_profile` has not been
-            called yet.
-        """
-        if getattr(self.builder, kind):
-            profile = TransformProfile(
-                {
-                    "grid_model": getattr(self.builder, "grid_model"),
-                    "base_%s" % kind: getattr(self.builder, kind),
-                },
-                self.get_grid(),
-                self.get_ct(),
-            )
-            return profile.get_profile(kind)
-        else:
-            raise Exception("%s profile version not set" % kind)
-
-    def get_demand(self):
-        """Returns demand profile.
-
-        :return: (*pandas.DataFrame*) -- data frame of demand (hour, zone id).
-        """
-        return self.get_profile("demand")
-
     def get_bus_demand(self):
         """Returns demand profiles, by bus.
 
@@ -181,27 +122,6 @@ class Create(State):
         self._update_scenario_info()
         grid = self.get_grid()
         return get_bus_demand(self._scenario_info, grid)
-
-    def get_hydro(self):
-        """Returns hydro profile.
-
-        :return: (*pandas.DataFrame*) -- data frame of hydro power output (hour, plant).
-        """
-        return self.get_profile("hydro")
-
-    def get_solar(self):
-        """Returns solar profile.
-
-        :return: (*pandas.DataFrame*) -- data frame of solar power output (hour, plant).
-        """
-        return self.get_profile("solar")
-
-    def get_wind(self):
-        """Returns wind profile.
-
-        :return: (*pandas.DataFrame*) -- data frame of wind power output (hour, plant).
-        """
-        return self.get_profile("wind")
 
     def create_scenario(self):
         """Creates scenario."""
@@ -301,11 +221,17 @@ class _Builder(object):
     wind = ""
     engine = "REISE.jl"
     exported_methods = {
-        "set_name",
-        "set_time",
         "set_base_profile",
         "set_engine",
+        "set_name",
+        "set_time",
+        "get_ct",
         "get_grid",
+        "get_demand",
+        "get_hydro",
+        "get_solar",
+        "get_wind",
+        "change_table",
     }
 
     def __init__(self, grid_model, interconnect, table):
@@ -319,6 +245,60 @@ class _Builder(object):
         self.change_table = ChangeTable(self.base_grid)
 
         self.existing = table[table.interconnect == self.interconnect]
+
+    def get_ct(self):
+        """Returns change table.
+
+        :return: (*dict*) -- change table.
+        """
+        return copy.deepcopy(self.change_table.ct)
+
+    def get_profile(self, kind):
+        """Returns demand, hydro, solar or wind  profile.
+
+        :param str kind: either *'demand'*, *'hydro'*, *'solar'*, *'wind'*.
+        :return: (*pandas.DataFrame*) -- profile.
+        """
+        if getattr(self, kind):
+            profile = TransformProfile(
+                {
+                    "grid_model": self.grid_model,
+                    "base_%s" % kind: getattr(self, kind),
+                },
+                self.get_grid(),
+                self.get_ct(),
+            )
+            return profile.get_profile(kind)
+        else:
+            raise Exception("%s profile version not set" % kind)
+
+    def get_demand(self):
+        """Returns demand profile.
+
+        :return: (*pandas.DataFrame*) -- data frame of demand (hour, zone id).
+        """
+        return self.get_profile("demand")
+
+    def get_hydro(self):
+        """Returns hydro profile.
+
+        :return: (*pandas.DataFrame*) -- data frame of hydro power output (hour, plant).
+        """
+        return self.get_profile("hydro")
+
+    def get_solar(self):
+        """Returns solar profile.
+
+        :return: (*pandas.DataFrame*) -- data frame of solar power output (hour, plant).
+        """
+        return self.get_profile("solar")
+
+    def get_wind(self):
+        """Returns wind profile.
+
+        :return: (*pandas.DataFrame*) -- data frame of wind power output (hour, plant).
+        """
+        return self.get_profile("wind")
 
     def set_name(self, plan_name, scenario_name):
         """Sets scenario name.
--- a/powersimdata/scenario/create.py
+++ b/powersimdata/scenario/create.py
@@ -15,6 +15,19 @@ from powersimdata.scenario.execute import Execute
 from powersimdata.scenario.state import State
 from powersimdata.utility import server_setup
 
+default_exported_methods = (
+    "create_scenario",
+    "get_ct",
+    "get_grid",
+    "get_demand",
+    "get_bus_demand",
+    "get_hydro",
+    "get_solar",
+    "get_wind",
+    "print_scenario_info",
+    "set_builder",
+)
+
 
 class Create(State):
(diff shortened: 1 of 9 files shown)


danielolsen:
I'm thinking it would be nice to be able to access the `Builder` methods as well from within `Create`, if we move the `exported` methods from a class attribute to something that's set in `init`, then we can add `set_name`, `set_time` etc. to `self.exported_methods` when the builder is set. What do you think?

rouille:
> I'm thinking it would be nice to be able to access the `Builder` methods as well from within `Create`, if we move the `exported` methods from a class attribute to something that's set in `init`, then we can add `set_name`, `set_time` etc. to `self.exported_methods` when the builder is set. What do you think?

It makes sense

powersimdata/scenario/scenario.py
```diff
@@ -41,6 +42,15 @@ def __init__(self, descriptor):
             except AttributeError:
                 return
 
+    def __getattr__(self, name):
```
jon-hagg, line 45: Any concern about this breaking calls to `getattr`? One similar idea would be to do a bunch of `setattr` calls in the constructor for the exported methods, which would leave the default behavior here, and make the methods available for tab completion on the scenario object (not sure if this is the case here).

danielolsen: From my limited research, the `__getattribute__` method is called any time you look up an attribute, and the `__getattr__` method is called only if `__getattribute__` fails.

What's an example of a call that could break here?

jon-hagg: Oh I see, not too worried in this case. I'm not sure if it's used by automated tools though, like pytest or sphinx, but we can see what happens. I still think it'd be nice to have the attributes set up front, so tab complete or `dir(scenario)` would have them, but we'd need to refresh when the state changes. What do you think?

danielolsen: Could we define `__dir__` so that it runs `super().__dir__()` and then adds `self.state.exported_methods`?

BainanXia: @danielolsen I think that should work.
```
def __dir___(self):
return sorted(set(super().__dir__() + self.state.exported_methods))
```

rouille: How your code would look like @jon-hagg with the `setattr`?

jon-hagg: Nevermind that, I think overriding `__dir__` achieves the same thing with less work. Bainan's snippet works but need to convert the exported_methods to a list, or something similar.

danielolsen: Done.

danielolsen:
With the latest commits, we can access the `Builder` methods from the `Scenario` object, if a builder is set. If a builder is not set, we get a custom error :
```python
>>> import powersimdata as psd
>>> scenario1 = psd.Scenario()
>>> scenario1.set_time()
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\scenario.py", line 49, in __getattr__
    return self.state._custom_getattr_error_message(name)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 82, in _custom_getattr_error_message
    return self.__getattr__(name)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 76, in __getattr__
    raise AttributeError(
AttributeError: Create object without a builder set has no attribute set_time. Did you forget to run set_builder?
>>> scenario1.set_builder(interconnect="Texas")
>>> scenario1.set_time()
TypeError: set_time() missing 3 required positional arguments: 'start_date', 'end_date', and 'interval'
>>> scenario1.set_timeee()
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\scenario.py", line 49, in __getattr__
    return self.state._custom_getattr_error_message(name)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 82, in _custom_getattr_error_message

  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 74, in __getattr__
    else:
AttributeError: Create object has no attribute set_timeee
```
The set of allowed methods is not shared among instances of `Create`:
```python
>>> scenario2 = psd.Scenario()
>>> scenario2.set_time()
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\scenario.py", line 49, in __getattr__
    return self.state._custom_getattr_error_message(name)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 82, in _custom_getattr_error_message
    return self.__getattr__(name)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 76, in __getattr__
    raise AttributeError(
AttributeError: Create object without a builder set has no attribute set_time. Did you forget to run set_builder?
```

We still maintain the same behavior for other states, which don't have a custom message defined:
```python
>>> scenario3 = psd.Scenario(1270)
>>> scenario3.foo
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\scenario.py", line 51, in __getattr__
    raise AttributeError(
AttributeError: Scenario object in analyze state has no attribute foo
```

BainanXia (approved):
I like it. Thanks!

danielolsen:
If we like everything so far, there's one more change that @rouille proposed, having the `Create` methods that depend on a builder being set to move to the Builder object (`get_ct`, `get_grid`, `get_profile`, `get_demand`, `get_bus_demand`, `get_hydro`, `get_solar`, `get_wind`). Now that I type all this out, I realize that the only things that would really be left in the `Create` object are `create_scenario`, `print_scenario_info`, and `set_builder`, and really `create_scenario` needs a builder to have been set too.

powersimdata/scenario/create.py
```diff
@@ -49,8 +56,24 @@ def __init__(self, scenario):
                 ("engine", ""),
             ]
         )
+        self.exported_methods = set(default_exported_methods)
         super().__init__(scenario)
 
+    def __getattr__(self, name):
```
jon-hagg, line 62: Does this fail when called with a method that is not in the builder now? e.g. the first if statement in `Scenario.__getattr__`

danielolsen:
```
>>> import powersimdata as psd
>>> scenario = psd.Scenario()
>>> scenario.foo
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\scenario.py", line 49, in __getattr__
    return self.state._custom_getattr_error_message(name)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 75, in _custom_getattr_error_message
    return self.__getattr__(name)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 69, in __getattr__
    raise AttributeError(
AttributeError: Create object without a builder set has no attribute foo. Did you forget to run set_builder?
>>> scenario.set_builder(interconnect="Texas")
Transferring ScenarioList.csv from server
100%|#######################################| 543k/543k [00:00<00:00, 4.18Mb/s]
Reading bus.csv
Reading plant.csv
Reading gencost.csv
Reading branch.csv
Reading dcline.csv
Reading sub.csv
Reading bus2sub.csv
Reading zone.csv
--> Summary
# Existing study
test | base | Anchor | Julia | Texas_Crisis
# Available profiles
demand: vJan2021
hydro: vJan2021
solar: vJan2021
wind: vJan2021
>>> scenario.foo
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\scenario.py", line 49, in __getattr__
    return self.state._custom_getattr_error_message(name)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 75, in _custom_getattr_error_message
    return self.__getattr__(name)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 67, in __getattr__
    raise AttributeError(f"Create object has no attribute {name}")
AttributeError: Create object has no attribute foo
```

jon-hagg: Thanks, TIL the difference between `getattr` and `__getattr__`

danielolsen:
All builder-requiring methods have been moved to the `_Builder` class.

danielolsen:
If we're hoping to make things as simple/logical as possible for the user, should we add an alias `set_grid` that points to `set_builder`?

rouille:
> If we're hoping to make things as simple/logical as possible for the user, should we add an alias `set_grid` that points to `set_builder`?

I don't mind. Should it be `set_scenario` instead? It would make since we have `create_scenario`

danielolsen:
> > If we're hoping to make things as simple/logical as possible for the user, should we add an alias `set_grid` that points to `set_builder`?
> 
> I don't mind. Should it be `set_scenario` instead? It would make since we have `create_scenario`

I would say that the scenario = the base grid + changes + the profile info, and when we're setting the builder we're only setting the base grid, so I would prefer `set_grid`, or `set_base_grid`, or something along those lines.

danielolsen:
I fixed one small bug where the `exported_methods` weren't being updated during state changes, and I've tested successfully on the server creating/executing/deleting.

rouille:
Should we extend the logic to the move state?

danielolsen:
> Should we extend the logic to the move state?

Oh yeah, I always forget about that one. I'll add it.

Unrelated point: do we want to go one layer deeper, and allow us to access the `ChangeTable` methods from a Scenario object in `Create` state too?

rouille:
> > Should we extend the logic to the move state?
> 
> 
> 
> Oh yeah, I always forget about that one. I'll add it.
> 
> 
> 
> Unrelated point: do we want to go one layer deeper, and allow us to access the `ChangeTable` methods from a Scenario object in `Create` state too?

Not sure. It personally sounds well to do scenario.set_time or scenario.get_wind but I am not sure about scenario.add_branch. I agree that the current way scenario.builder.add_branch is not great

danielolsen:
> > Unrelated point: do we want to go one layer deeper, and allow us to access the `ChangeTable` methods from a Scenario object in `Create` state too?
> 
> Not sure. It personally sounds well to do scenario.set_time or scenario.get_wind but I am not sure about scenario.add_branch. I agree that the current way scenario.builder.add_branch is not great

It's even worse, right now we need `scenario.state.builder.change_table.add_branch`.

rouille:
> > > Unrelated point: do we want to go one layer deeper, and allow us to access the `ChangeTable` methods from a Scenario object in `Create` state too?
> 
> > 
> 
> > Not sure. It personally sounds well to do scenario.set_time or scenario.get_wind but I am not sure about scenario.add_branch. I agree that the current way scenario.builder.add_branch is not great
> 
> 
> 
> It's even worse, right now we need `scenario.state.builder.change_table.add_branch`.

If we change set_builder to set_grid should we change builder to grid and access the change table method through scenario.grid.add_branch? The only hiccup would be either the scale_demand since it is not specifically grid related.

danielolsen:
> > > Not sure. It personally sounds well to do scenario.set_time or scenario.get_wind but I am not sure about scenario.add_branch. I agree that the current way scenario.builder.add_branch is not great
> > 
> > 
> > It's even worse, right now we need `scenario.state.builder.change_table.add_branch`.
> 
> If we change set_builder to set_grid should we change builder to grid and access the change table method through scenario.grid.add_branch? The only hiccup would be either the scale_demand since it is not specifically grid related.

What would `scenario.grid` be? If it's not a Grid object, that's sort of confusing, and if it is a Grid object then it's a little confusing that we can modifying it but only when in the Create state?

What if we just had `scenario.change_table` as a shortcut to `scenario.state.builder.change_table`?

danielolsen:
The latest commits add the `change_table` shortcut and the `set_grid` alias to `set_builder`. We can now do:
```python
>>> import powersimdata as psd
>>> scenario = psd.Scenario()
>>> scenario.set_grid(interconnect="Texas")
Transferring ScenarioList.csv from server
100%|#######################################| 546k/546k [00:00<00:00, 3.42Mb/s]
Reading bus.csv
Reading plant.csv
Reading gencost.csv
Reading branch.csv
Reading dcline.csv
Reading sub.csv
Reading bus2sub.csv
Reading zone.csv
--> Summary
# Existing study
test | base | Anchor | Julia | Texas_Crisis
# Available profiles
demand: vJan2021
hydro: vJan2021
solar: vJan2021
wind: vJan2021
>>> scenario.change_table.scale_demand(zone_id={301: 1.2})
```

powersimdata/scenario/create.py
```diff
@@ -49,8 +57,23 @@ def __init__(self, scenario):
...
+        if self.builder is not None and name in self.builder.exported_methods:
+            return getattr(self.builder, name)
+        if self.builder is None and name in _Builder.exported_methods:
+            raise AttributeError(
+                f"Create object without a grid set has no attribute {name}. "
+                "Did you forget to run set_grid?"
+            )
+        else:
+            raise AttributeError(f"Create object has no attribute {name}")
```
rouille, line 72: I would use `method` in place of `attribute`

powersimdata/scenario/create.py
```diff
@@ -49,8 +57,23 @@ def __init__(self, scenario):
...
             ]
         )
+        self.exported_methods = set(default_exported_methods)
         super().__init__(scenario)
 
+    def __getattr__(self, name):
+        if self.builder is not None and name in self.builder.exported_methods:
+            return getattr(self.builder, name)
+        if self.builder is None and name in _Builder.exported_methods:
+            raise AttributeError(
```
rouille, line 67: I would print: Call `set_grid` first to access {name} method

powersimdata/scenario/create.py
```diff
@@ -251,6 +201,10 @@ def set_builder(self, grid_model="usa_tamu", interconnect="USA"):
         self._scenario_info["grid_model"] = self.builder.grid_model
         self._scenario_info["interconnect"] = self.builder.interconnect
 
+    def set_grid(self, *args, **kwargs):
+        """Alias to set_builder."""
+        self.set_builder(*args, **kwargs)
```
rouille, line 213: Should we have a `DeprecationWarning` here to incite users to use this method and not `set_builder`? The idea would be to remove `set_builder` in a future release.

powersimdata/scenario/create.py
danielolsen: They can still be accessing any arbitrary attribute here.

powersimdata/scenario/create.py
danielolsen: Replacing all of the existing AttributeError test?

powersimdata/scenario/create.py
danielolsen: If we want to do that, then we should move the functional code to `set_grid` and have `set_builder` be the alias with the warning.

danielolsen:
Based on some conversations with @rouille, I added some protection to the alias names, so that if the user tries to assign anything to these names they will get an error. That way, if people get used to accessing `scenario.state.x` via `scenario.x` and try to assign something via `scenario.x`, they will get an error so that there won't be a new `scenario.x` as well as the old (different) `scenario.state.x`.
```python
>>> import powersimdata as psd
>>> scenario = psd.Scenario()
>>> scenario.set_time()
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\scenario.py", line 56, in __getattr__
    return self.state._custom_getattr_error_message(name)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 82, in _custom_getattr_error_message
    return self.__getattr__(name)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 67, in __getattr__
    raise AttributeError(
AttributeError: Create object without a grid set has no attribute set_time. Did you forget to run set_grid?
>>> scenario.set_color()
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\scenario.py", line 56, in __getattr__
    return self.state._custom_getattr_error_message(name)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 82, in _custom_getattr_error_message
    return self.__getattr__(name)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 72, in __getattr__
    raise AttributeError(f"Create object has no attribute {name}")
AttributeError: Create object has no attribute set_color
>>> scenario.set_grid = "foo"
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\scenario.py", line 72, in __setattr__
    raise AttributeError(
AttributeError: set_grid is exported from Scenario.state, edit it there if necessary
>>> scenario.set_base_profile = "foo"
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\scenario.py", line 67, in __setattr__
    raise AttributeError(
AttributeError: set_base_profile is exported from Scenario.state.builder, edit it there if necessary
>>> scenario.state.set_base_profile = "foo"
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 76, in __setattr__
    raise AttributeError(
AttributeError: set_base_profile is exported from Create.builder, edit it there if necessary
>>> scenario.foo
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\scenario.py", line 56, in __getattr__
    return self.state._custom_getattr_error_message(name)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 82, in _custom_getattr_error_message
    return self.__getattr__(name)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\create.py", line 72, in __getattr__
    raise AttributeError(f"Create object has no attribute {name}")
AttributeError: Create object has no attribute foo
>>> scenario.foo = "bar"
>>> scenario.foo
'bar'
```

powersimdata/scenario/create.py
rouille: You are right.

powersimdata/scenario/create.py
rouille: Not sure I understand. I was just suggesting to replace the error message 
```
f"Create object without a grid set has no attribute {name}. "
 "Did you forget to run set_grid?"
```
with:
```
f"Call set_grid first to access {name} attribute" 
```

danielolsen: Done.

powersimdata/scenario/create.py
danielolsen: Done.

ahurli (commented):
I'm glad we're simplifying the calls to use scenarios!

I think this is outside the scope of this PR, but this makes me curious about the functionality/future of the `Builder` class. Setting `set_builder` to `set_grid` makes it seem like we're going to use different Builders for different grids going forward, unless the plan is to remove it completely later. I could see different builders being good for different classes of problems/engines in the future, though. Are there any plans for this?

powersimdata/scenario/scenario.py
```diff
@@ -41,6 +50,34 @@ def __init__(self, descriptor):
...
+        if name in self.state.exported_methods:
+            return getattr(self.state, name)
+        elif hasattr(self.state, "_custom_getattr_error_message"):
```
BainanXia, line 56: I'm trying to understand what we are trying achieve here?

danielolsen: This is so that if we are in Create state, and we try to get something from the Scenario object that is only available when the builder has been set, we get a more informative message.

danielolsen:
> I'm glad we're simplifying the calls to use scenarios!
> 
> I think this is outside the scope of this PR, but this makes me curious about the functionality/future of the `Builder` class. Setting `set_builder` to `set_grid` makes it seem like we're going to use different Builders for different grids going forward, unless the plan is to remove it completely later. I could see different builders being good for different classes of problems/engines in the future, though. Are there any plans for this?

We had been using different child classes of the Builder for different interconnections, with the same sort of thought that maybe they needed to have different functionality, but that added functionality never developed so we re-simplified in #398. If we need to differentiate them again we can follow the old pattern there.

powersimdata/scenario/scenario.py
BainanXia: This is not limited to "Create" state only, right? Do you think the code block I posted will work from this perspective or we have to define this wrapper function `_custom_getattr_error_message` which simply returns `self.__getattr__(name)`?

BainanXia: I think the logic is if the scenario is in a `State` other than `Create` and the `name` is a valid attribute/method, it will go through `if` block, otherwise, we are going to raise an AttributeError anyway (a redundant try in this specific situation). Yours is more explicit.

danielolsen: Done.

powersimdata/scenario/create.py
```diff
@@ -15,6 +16,14 @@
 from powersimdata.scenario.state import State
 from powersimdata.utility import server_setup
 
+default_exported_methods = (
+    "create_scenario",
+    "get_bus_demand",
+    "print_scenario_info",
+    "set_builder",
+    "set_grid",
+)
```
rouille, line 25: Do we need this tuple outside of the class? We should be able to move it in the constructor:
```
self.exported_methods = (
    "create_scenario",
    "get_bus_demand",
    "print_scenario_info",
    "set_builder",
    "set_grid",
)
```
No?

danielolsen: I think you're right, as long it's not defined at class level so that it's not shared.

danielolsen: Done.

rouille (approved):
Very nice

PR #425, opened 2021-03-25
ci: add pep8-naming check to flake8 and ignore a bunch of warnings

jon-hagg:
### Purpose
Add pep8-naming check to flake8, so it is run by tox locally and in github. 
See https://github.com/Breakthrough-Energy/RenewableEnergyProject/issues/343

### What the code is doing
In this case, we ignore almost all the warnings. Note: adding the noqa on some lines also caused black to reformat into multi line declarations, which in turn requires moving the noqa back to the first line in that group. Hopefully we don't have to do this a lot. 

### Testing
`tox -e flake8`

### Time estimate
5 min

### DIFF
--- a/powersimdata/design/investment/create_mapping_files.py
+++ b/powersimdata/design/investment/create_mapping_files.py
@@ -223,7 +223,7 @@ def write_poly_shapefile():
     """
     fiona = _check_import("fiona")
     shapely_geometry = _check_import("shapely.geometry")
-    polygon = shapely_geometry.Polygon
+    Polygon = shapely_geometry.Polygon  # noqa: N806
     mapping = shapely_geometry.mapping
 
     outpath = const.reeds_wind_shapefile_path
@@ -258,7 +258,7 @@ def write_poly_shapefile():
             for j in poly_df.index:
                 ls += [(poly_df.loc[j, "long"], poly_df.loc[j, "lat"])]
 
-            poly = polygon(ls)
+            poly = Polygon(ls)
             c.write(
                 {
                     "geometry": mapping(poly),
--- a/powersimdata/design/investment/create_mapping_files.py
+++ b/powersimdata/design/investment/create_mapping_files.py
@@ -223,7 +223,7 @@ def write_poly_shapefile():
     """
     fiona = _check_import("fiona")
     shapely_geometry = _check_import("shapely.geometry")
-    Polygon = shapely_geometry.Polygon
+    polygon = shapely_geometry.Polygon
     mapping = shapely_geometry.mapping
 
     outpath = const.reeds_wind_shapefile_path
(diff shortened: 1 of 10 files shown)


BainanXia (approved):
Looks good. Thanks for taking care of this.

powersimdata/design/transmission/tests/test_upgrade.py
```diff
@@ -139,43 +139,43 @@ def setUp(self):
         self.grid.id2zone = {201: "Wahington", 202: "Oregon"}
         self.grid.zone2id = {"Washington": 201, "Oregon": 202}
 
-    def test_internal_W(self):
+    def test_internal_W(self):  # noqa: N802
```
danielolsen, line 142: These function names are an artifact of old tests where we had dummy zones `"E"` and `"W"`. Now that we've changed the tests to `Oregon`/`Washington`, it probably makes sense to change the test names.

powersimdata/design/investment/create_mapping_files.py
```diff
@@ -223,7 +223,7 @@ def write_poly_shapefile():
     """
     fiona = _check_import("fiona")
     shapely_geometry = _check_import("shapely.geometry")
-    Polygon = shapely_geometry.Polygon
+    polygon = shapely_geometry.Polygon
```
danielolsen, line 226: Should we keep this as `Polygon` and `#noqa` the error? What we're really doing is importing a class, which I think we should keep uppercase if possible.

jon-hagg: Sure, I think it makes sense in this case.

powersimdata/design/transmission/tests/test_upgrade.py
jon-hagg: I gave it a shot, let me know if it makes sense.

danielolsen:
@lanesmith can we come up with some more descriptive variable names for the cost curve visualizations?

lanesmith:
> @lanesmith can we come up with some more descriptive variable names for the cost curve visualizations?

@danielolsen, yeah certainly. What's the best way for me make those additions to this PR? Should I just checkout this branch?

danielolsen:
> @danielolsen, yeah certainly. What's the best way for me make those additions to this PR? Should I just checkout this branch?

I suggest that you check out this branch, create a new branch of your own with your commit(s), and push that. Then @jon-hagg can cherry-pick your commits into this branch when ready, and there's no risk of conflicts.

lanesmith:
@jon-hagg, I've finished updating the variable names for the cost curve visualization. The updates are in the branch `lane/cost_curves_refactor`. Let me know how you would like to proceed with this.

jon-hagg:
> @jon-hagg, I've finished updating the variable names for the cost curve visualization. The updates are in the branch `lane/cost_curves_refactor`. Let me know how you would like to proceed with this.

Thanks @lanesmith, probably the easiest way would be to create a PR to this branch, which (when merged) will effectively append your commits to the current PR.

danielolsen (approved):
Looks good, thanks!

PR #406, opened 2021-02-25
refactor: move generation of matfile to separate function

danielolsen:
### Purpose
Decouple the matfile generation from the client/server setup. Closes #364.

### What the code is doing
Code doesn't functionally change, it's just reorganized.

### Testing
Not tested yet to prepare a scenario, but tested for creating a matfile from a base grid.

### Usage Example/Visuals
```
>>> from powersimdata.input.grid import Grid
>>> from powersimdata.scenario.helpers import export_case_mat
>>> grid = Grid('Texas')
Reading bus.csv
Reading plant.csv
Reading gencost.csv
Reading branch.csv
Reading dcline.csv
Reading sub.csv
Reading bus2sub.csv
Reading zone.csv
>>> export_case_mat(grid, "dummy.mat")
```

### Time estimate
5 minutes to understand, maybe longer to decide exactly where you think this should live.

PR #427, opened 2021-03-26
fix: upload storage matfile to server

danielolsen:
### Purpose
We introduced a bug in #406 where the "case_storage.mat" file is generated but not actually sent to the server and therefore not used in the simulation engine.

### Testing
Tested that this fixes the bug, and **case_storage.mat** gets uploaded to the server where previous scenarios run with the exact same script don't have it.

### Time estimate
5 minutes.

### DIFF
--- a/powersimdata/scenario/execute.py
+++ b/powersimdata/scenario/execute.py
@@ -362,6 +362,10 @@ class SimulationInput(object):
         self._data_access.move_to(
             file_name, self.REL_TMP_DIR, change_name_to="case.mat"
         )
+        if len(self.grid.storage["gen"]) > 0:
+            self._data_access.move_to(
+                storage_file_name, self.REL_TMP_DIR, change_name_to="case_storage.mat"
+            )
 
     def prepare_profile(self, kind, profile_as=None):
         """Prepares profile for simulation.


BainanXia (approved):
Good catch. We haven't run scenarios with storage for a while.

danielolsen:
> Good catch. We haven't run scenarios with storage for a while.

I caught it the hard way, after running 162 scenarios that I thought had storage but in fact did not.

BainanXia:
> > Good catch. We haven't run scenarios with storage for a while.
> 
> I caught it the hard way, after running 162 scenarios that I thought had storage but in fact did not.

OMG, all nuclear scenarios with flexibility?!

danielolsen:
> OMG, all nuclear scenarios with flexibility?!

Not all of them, just the ones from this week (90% clean, 30% nuclear).

BainanXia:
> > OMG, all nuclear scenarios with flexibility?!
> 
> Not all of them, just the ones from this week (90% clean, 30% nuclear).

Does it affect the results in the presentation?

danielolsen:
> Does it affect the results in the presentation?

Not really, the ones running this week did not play a major role in what we presented.

rouille:
Do we need to delete a 162 scenarios?

PR #440, opened 2021-04-05
Correct bug for Scenario init with bad descriptor

danielolsen:
### Purpose
Add a test to check a bug for when we try to get a Scenario that isn't in the ScenarioList, and then fix the bug. This bug was introduced in https://github.com/Breakthrough-Energy/PowerSimData/pull/420.

### What the code is doing
We raise a `ValueError` before a `RecursionError` would be created (the previous bug).

### Testing
Unit test finds the error before (or crashed python), and confirms that the fix works.

### Time estimate
5 minutes.


BainanXia (approved):
Looks good.

jon-hagg (approved):

rouille (approved):
Thanks

PR #446, opened 2021-04-09
fix: cleanup temp file and fix PermissionError when file is in use

jon-hagg:
### Purpose
Small fixes for issues with the temp files

### What the code is doing
In `data_access.py` - wait til after the context manager to remove the file, so it won't be in use by paramiko
In `csv_store.py`  - delete the temp file if necessary. This prevents leftover copies generated by tests, and within containers (anything using `LocalDataAccess`, which does a no-op for it's `.push` method). 

### Testing
Only verified the temp file cleanup, hoping the other part is safe enough 

### Time estimate
5 min


rouille (approved):
It makes sense

BainanXia (approved):

PR #449, opened 2021-04-13
fix: close temp file before trying to move/delete

danielolsen:
### Purpose
Fix the PermissionsErrors we had been getting on Windows.

### What the code is doing
After writing to the tempfile, we tell the os to close the file before we try to do anything with it (e.g. move, delete). I got this from this stackoverflow post: https://stackoverflow.com/questions/34716996/cant-remove-a-file-which-created-by-tempfile-mkstemp-on-windows

### Testing
Tests on Windows now pass, and I've successfully tested that this can create/prepare/launch a scenario on the server (previously had been failing so I fell back to the 'v0.4.1' branch).

Previously all **test_execute_csv** and **test_scenario_csv** tests failed with an error like:
`PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\DanielOlsen\\ScenarioData\\tmpmvu_jnyl'`

### Time estimate
5-15 minutes, depending on how far you want to go down the low level file API rabbit hole.

### DIFF
--- a/powersimdata/data_access/csv_store.py
+++ b/powersimdata/data_access/csv_store.py
@@ -74,9 +74,10 @@ class CsvStore:
         :param pandas.DataFrame table: the data frame to save
         :param str checksum: the checksum prior to download
         """
-        _, tmp_path = mkstemp(dir=server_setup.LOCAL_DIR)
+        tmp_file, tmp_path = mkstemp(dir=server_setup.LOCAL_DIR)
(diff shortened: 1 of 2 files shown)


rouille (approved):
I don't see anything wrong with that

BainanXia (approved):
Make sense.

jon-hagg (approved):

Issue #439, opened 2021-04-05
Improve flexibility of congested branch upgrade algorithm


danielolsen:
### Background

Currently, in `powersimdata.design.transmission.upgrade.scale_congested_mesh_branches`, we estimate the benefit of upgrading a branch based on its `(shadow prices in 95th percentile of hours for a reference scenario)`, divided by the 'cost' of `{1, MW capacity in base grid, or MW-miles in base grid}` for the `"branches"`, `"MW"`, and `"MWmiles"` methods, respectively (controlled by the `method` parameter.

The 95th percentile is user-override-able, but the user cannot choose other ways of estimating the benefit part of the equation.

When this algorithm was originally developed, MW-miles was the best proxy for cost that we had, but not we have per-branch cost estimation methods in the investment cost module.

### Improved options
- [ ] Allow the user to choose between a percentile-base method or a mean-based method for estimating benefit (or other additional method)
- [x] Allow the user to choose cost estimation based on the investment cost analysis module. This will be slower, but more accurate, since the per-MW-mile cost of higher-voltage lines should be better than lower-voltage ones, and certain locations have higher cost multipliers than others. (Done via https://github.com/Breakthrough-Energy/PowerSimData/pull/452)
- [ ] Allow the algorithm to suggest the addition of new higher-voltage parallel lines between substations. This is definitely a stretch goal, but e.g. if we have a heavily congested 69 kV line between two substations that have 138 kV buses, we could add a new 138 kV branch, rather than just assuming that congestion will always be solved by adding parallel branches of the same voltage (effectively what we're doing now by linearly scaling rating and impedance).

danielolsen:
I've gotten started on the cost angle, it's a bit tricky because while we can get individual branch upgrade costs, the indices don't line up with the original branch IDs because of the pandas `merge`s within the AC cost calculation. Finding this problem was annoying but the fix shouldn't be too bad, I should have a PR up sometime this week.

BainanXia:
> I've gotten started on the cost angle, it's a bit tricky because while we can get individual branch upgrade costs, the indices don't line up with the original branch IDs because of the pandas `merge`s within the AC cost calculation. Finding this problem was annoying but the fix shouldn't be too bad, I should have a PR up sometime this week.

I don't quite understand but hopefully will get more info in the up coming PR🙏

danielolsen:
[...]

PR #297, opened 2020-09-18
Investment costs

nina-vincent:
Note: Not all tests are done. 

This adds cost calculations for all new (added since base_grid) transmission, generation, and storage capacity.

`create_mapping_files.py` contains functions that help create shapefiles and map lat/lon points to regions from shapefiles. 
`investment_costs.py` uses these mappings to find the regional cost multipliers (for both generation/storage and ac transmission).

### Fundamental calculations
There are 3 separate cost calculations: one for dclines, one for branches, one for plants.

**branches: calculate_ac_inv_costs**

For all capacity added on a **line**, the investment cost is:
`Cost ($2010) = rateA (MW) * lengthMi (miles) * costMWmi ($2010/MW-mi) * mult (regional cost multiplier, unitless)`
Then all lines are summed.

For all capacity added on a **transformer**, the investment cost is:
`Cost (per transformer at a given kV level, $2010)`

Now that I'm thinking about this again, this is a flawed method because if we're only slightly scaling a transformer, this is overestimating the cost. Might be better to apply to completely new transformers only? Either ignore this cost or reevaluate its calculation.

**dclines: calculate_dc_inv_costs**

For all capacity added on a dcline, the investment cost is:
`Cost ($2015) = Pmax (MW) * lengthMi (miles) * costMWmi ($2015/MW-mi) + dc_term_cost (DC terminal cost $2015)`
Then all line costs are summed.

**plant: calculate_gen_inv_costs**

For all capacity added on a plant, the investment cost is:
`Cost ($2018) = Pmax (MW) * CAPEX ($2018/MW) * reg_cap_cost_mult (regional capital cost multiplier, unitless)`
Then all costs are summed by technology (so you can ignore the non-renewables/storage if you want).

### Methods

**branches: calculate_ac_inv_costs**
1. Find new upgrades for each line/transformer (change in MW rating): `grid.branch.rateA - base_grid.branch.rateA`
2. Drop unchanged branches.
3. Map branches to corresponding kV.
4. Separate lines and transformers (TransformerWindings dropped).
5. Lines: Find closest kV present in the corresponding cost table using `select_kv` . Labeled "kV_cost." 
6. Lines: Find MW closest to the line's rateA that exists within the corresponding kV_cost sub-table. Then use that costMWmi base cost value.
7. (messy) Lines: Import bus to NEEM mapping file. Check that no buses you need are missing. Check that buses have the same lat/lon values as the pre-made mapping file. If any issues, re-map NEEM regions to those points.
[...]

PR #450, opened 2021-04-13
fix: ensure that per-element investment cost are calculated accurately

danielolsen:
### Purpose
- As originally mentioned in https://github.com/Breakthrough-Energy/PowerSimData/pull/297, fix AC investment cost calculations so that we don't use underground line costs when not appropriate (lines of appropriate rating.voltage that are entirely within `NEISO` and `NYISO J-K` regions).
- Add tests of individual element cost calculations, which illustrate some bugs (e.g. branches within the `SOCO` [Southern Company] region were getting `NaN` cost estimates, and underground line costs were being applied inappropriately to some branches).
- Fix the bugs.
- Cleanup from https://github.com/Breakthrough-Energy/PowerSimData/pull/391, where we enabled the cost calculation of nuclear generators, but didn't ensure that costs would be calculated for costs cases other than `"Moderate"`.

These code fixes are necessary for https://github.com/Breakthrough-Energy/PowerSimData/issues/439.

### What the code is doing
- In **powersimdata/tests/mock_grid.py**: we allow the user to pass data to the `MockGrid.storage["StorageData"]` field for tests.
- In **powersimdata/design/investment/tests/test_investment_costs.py**: we add tests of the individual element cost calculations (which initially fail)
- In **powersimdata/design/investment/investment_costs.py**:
  - We add a new function `merge_keep_index`, which lets us perform a dataframe merge without losing the index of the first dataframe (as demonstrated in the comments on https://github.com/Breakthrough-Energy/PowerSimData/issues/439), and we use this wherever we had been doing a `pandas.DataFrame.merge`.
  - We add a new function `append_keep_index_name` which lets us perform an append without losing the name of the first dataframe index, which we need for `merge_keep_index`.
  - In `_calculate_ac_inv_costs`:
    - We update the `select_mw` function to ensure that we only select the cost class for underground lines if the line is entirely within one of the two zones which have significant shares of underground lines (based on the data in the **LineRegMult.csv** file).
    - We generalize the `get_transformer_mult` function to be `get_branch_mult`, so that it can operate on all types of branches to return the average of the regional multiplier for the to & from ends of the branch. In the process, we fix a bug where if a Line has endpoints with no regional multiplier defined, it would return a `NaN` multiplier and then this branch would not count towards the total cost (and not be meaningful when results are returned un-summed).
    - Adding data to the branch data frame is moved after adding data to the bus data frame, so that we can take advantage of knowing the bus regions within the new `get_branch_mult` function.
  - In `calculate_gen_inv_costs`:
    - We modify the `load_cost` function so that if we request a cost case that is not `"Moderate"`, then we copy the `"Moderate"` nuclear costs into what we will return, and warn the user. This allows us to calculate costs for the `"Advanced"` and `"Conservative"` cost cases without having `NaN`s for nuclear.
    - We ensure that the indices of the storage units that we append to the plant data frame are preserved, so that there is no collision with the plant IDs in the returned non-summed data frame. We generate these non-colliding indices automatically whenever we add storage to the Grid, but they live in the `StorageData` storage table due to legacy design around the Matpower/MOST schema)

### Testing
Unit tests are added which detect the bugs, and the changes fix the bugs so that the tests run successfully. This branch has also been tested (with some additional code in the `daniel/congestion_cost_upgrade` branch) to successfully analyze per-branch upgrade costs for a real Eastern scenario.

### Time estimate
30-60 minutes. The `investment_cost` module is big, and requires aggregating many different tables to produce results, so it takes a while to trace through all the places where bugs can be introduced. There is definitely room to refactor this module to be structured a bit better, but that is of lower priority than fixing these bugs, which need to be addressed for the trainings that are coming up soon.

### DIFF
--- a/powersimdata/design/investment/investment_costs.py
+++ b/powersimdata/design/investment/investment_costs.py
@@ -13,6 +13,17 @@ from powersimdata.input.grid import Grid
 from powersimdata.utility.distance import haversine
 
 
+def merge_keep_index(df1, df2, **kwargs):
+    """Execute a pandas DataFrame merge, preserving the index of the first dataframe.
+
+    :param pandas.DataFrame df1: first data frame, to call pandas merge from.
+    :param pandas.DataFrame df2: second data frame, argument to pandas merge.
+    :param \\*\\*kwargs: arbitrary keyword arguments passed to pandas merge call.
+    :return: (*pandas.DataFrame*) -- df1 merged with df2 with indices preserved.
+    """
+    return df1.reset_index().merge(df2, **kwargs).set_index(df1.index.names)
+
+
 def calculate_ac_inv_costs(scenario, sum_results=True, exclude_branches=None):
     """Calculate cost of upgrading AC lines and/or transformers in a scenario.
     NEEM regions are used to find regional multipliers.
@@ -139,7 +150,7 @@ def _calculate_ac_inv_costs(grid_new, sum_results=True):
     # in region file
     bus_fix_index = bus[~bus.index.isin(bus_reg.index)].index
     bus_mask = bus[~bus.index.isin(bus_fix_index)]
-    bus_mask = bus_mask.merge(bus_reg, how="left", on="bus_id")
+    bus_mask = merge_keep_index(bus_mask, bus_reg, how="left", on="bus_id")
     # these buses have incorrect lat/lon values in the region mapping file.
     #   re-running the region mapping script on those buses only.
     bus_fix_index2 = bus_mask[
@@ -160,12 +171,20 @@ def _calculate_ac_inv_costs(grid_new, sum_results=True):
         id_vars=["kV", "MW"], var_name="name_abbr", value_name="mult"
     )
 
-    lines = lines.merge(bus_reg, left_on="to_bus_id", right_on="bus_id", how="inner")
-    lines = lines.merge(ac_reg_mult, on=["name_abbr", "kV", "MW"], how="left")
+    lines = merge_keep_index(
+        lines, bus_reg, left_on="to_bus_id", right_on="bus_id", how="inner"
+    )
+    lines = merge_keep_index(
+        lines, ac_reg_mult, on=["name_abbr", "kV", "MW"], how="left"
+    )
     lines.rename(columns={"name_abbr": "reg_to", "mult": "mult_to"}, inplace=True)
 
-    lines = lines.merge(bus_reg, left_on="from_bus_id", right_on="bus_id", how="inner")
-    lines = lines.merge(ac_reg_mult, on=["name_abbr", "kV", "MW"], how="left")
+    lines = merge_keep_index(
+        lines, bus_reg, left_on="from_bus_id", right_on="bus_id", how="inner"
+    )
+    lines = merge_keep_index(
+        lines, ac_reg_mult, on=["name_abbr", "kV", "MW"], how="left"
+    )
     lines.rename(columns={"name_abbr": "reg_from", "mult": "mult_from"}, inplace=True)
 
     # take average between 2 buses' region multipliers
@@ -402,7 +421,9 @@ def _calculate_gen_inv_costs(grid_new, year, cost_case, sum_results=True):
     gen_costs.replace(const.gen_inv_cost_translation, inplace=True)
     gen_costs.drop(["Key", "FinancialCase", "CRPYears"], axis=1, inplace=True)
     # ATB technology costs merge
-    plants = plants.merge(gen_costs, right_on="Technology", left_on="type", how="left")
+    plants = merge_keep_index(
+        plants, gen_costs, right_on="Technology", left_on="type", how="left"
+    )
 
     # REGIONAL COST MULTIPLIER
 
@@ -417,7 +438,9 @@ def _calculate_gen_inv_costs(grid_new, year, cost_case, sum_results=True):
     except FileNotFoundError:
         bus_reg = bus_to_reeds_reg(grid_new.bus.loc[plant_buses])
         bus_reg.sort_index().to_csv(const.bus_reeds_regions_path)
-    plants = plants.merge(bus_reg, left_on="bus_id", right_index=True, how="left")
+    plants = merge_keep_index(
+        plants, bus_reg, left_on="bus_id", right_index=True, how="left"
+    )
 
     # Determine one region 'r' for each plant, based on one of two mappings
     plants.loc[:, "r"] = ""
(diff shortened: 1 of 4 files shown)


powersimdata/design/investment/investment_costs.py
```diff
@@ -13,6 +14,30 @@
 from powersimdata.utility.distance import haversine
 
 
+def merge_keep_index(df1, df2, **kwargs):
+    """Execute a pandas DataFrame merge, preserving the index of the first dataframe.
+
+    :param pandas.DataFrame df1: first data frame, to call pandas merge from.
+    :param pandas.DataFrame df2: second data frame, argument to pandas merge.
+    :param \\*\\*kwargs: arbitrary keyword arguments passed to pandas merge call.
```
BainanXia, line 22: Missing `return` docstrings.

powersimdata/design/investment/investment_costs.py
```diff
@@ -13,6 +14,30 @@
...
+    return df1.reset_index().merge(df2, **kwargs).set_index(df1.index.names)
+
+
+def append_keep_index_name(df1, other, *args, **kwargs):
+    """Execute a pandas DataFrame append, preserving the index name of the dataframe.
+
+    :param pandas.DataFrame df1: first data frame, to call pandas append from.
+    :param pandas.DataFrame/pandas.Series/list: first argument to pandas append method.
+    :param \\*args: arbitrary positional arguments passed to pandas append call.
+    :param \\*\\*kwargs: arbitrary keyword arguments passed to pandas append call.
```
BainanXia, line 33: Same here.

powersimdata/design/investment/investment_costs.py
```diff
@@ -59,27 +84,35 @@ def select_mw(x, cost_df):
         :return: (*pandas.Series*) -- series of [*'MW'*, *'costMWmi'*] to be assigned
             to branch.
         """
-
+        underground_regions = ("NEISO", "NYISO J-K")
+        cost_df = cost_df.copy()
```
BainanXia, line 88: I understand the purpose here, however, it would be better to use a different variable name to reduce confusion.

powersimdata/design/investment/investment_costs.py
```diff
@@ -59,27 +84,35 @@ def select_mw(x, cost_df):
...
         # get rid of NaN values in this kV table
-        tmp = tmp[~tmp["MW"].isna()]
+        cost_df = cost_df[~cost_df["MW"].isna()]
         # find closest MW & corresponding cost
-        return tmp.iloc[np.argmin(np.abs(tmp["MW"] - x.rateA))][["MW", "costMWmi"]]
+        return_cols = ["MW", "costMWmi"]
+        return cost_df.iloc[np.argmin(np.abs(cost_df["MW"] - x.rateA))][return_cols]
 
-    def get_transformer_mult(x, bus_reg, ac_reg_mult, xfmr_lookup_alerted=set()):
+    def get_branch_mult(x, bus_reg, ac_reg_mult, branch_lookup_alerted=set()):
```
BainanXia, line 100: I think based on our convention, we should avoid mutable default values?

powersimdata/design/investment/investment_costs.py
```diff
@@ -155,21 +170,36 @@ def get_transformer_mult(x, bus_reg, ac_reg_mult, xfmr_lookup_alerted=set()):
 
     bus_reg.drop(["lat", "lon"], axis=1, inplace=True)
 
+    # Add extra information to branch data frame
+    branch = grid_new.branch
+    branch.loc[:, "kV"] = branch.apply(
+        lambda x: bus.loc[x.from_bus_id, "baseKV"], axis=1
+    )
```
BainanXia, line 177: Should this be equivalent to `branch["kV"] = bus.loc[branch["from_bus_id"], "baseKV"]`?

powersimdata/design/investment/investment_costs.py
danielolsen: This design is intentional. Since we re-define `get_branch_mult` for every call of `_calculate_ac_inv_costs`, it is never shared via user-accessible calls. We do want it to be shared via the many calls to `get_branch_mult` via the dataframe `apply`, so that it serves its purpose of not the user multiple times for the same missing regional multipliers.

I believe this was discussed at some point in the review for https://github.com/Breakthrough-Energy/PowerSimData/pull/297.

powersimdata/tests/mock_grid.py
```diff
@@ -151,6 +151,22 @@
...
+        "TerminalStoragePrice",
+        "MinStorageLevel",
+        "MaxStorageLevel",
+        "OutEff",
+        "InEff",
+        "LossFactor",
+        "rho",
+        "ExpectedTerminalStorageMax",
+        "ExpectedTerminalStorageMin",
+    ],
```
BainanXia, line 169: Nice.

powersimdata/tests/mock_grid.py
```diff
@@ -14,7 +14,7 @@
 }
 
 gencost_names = {"gencost_before": "before", "gencost_after": "after"}
-storage_names = {"storage_gen": "gen"}
+storage_names = {"storage_gen": "gen", "storage_StorageData": "StorageData"}
```
BainanXia, line 17: Shall we use "storage_data" instead of "storage_StorageData" or there is another `storage_data` defined somewhere else.

powersimdata/design/investment/investment_costs.py
BainanXia: Nevermind then.

BainanXia:
Thanks for updating the tests and the Mock object. The logic is clear and all tests passed. I think this is good to go after some minor concerns are addressed.

powersimdata/tests/mock_grid.py
danielolsen: The logic here was: `"storage_gen"` reflects the `"gen"` key in the `storage` dict. Therefore, `"storage_StorageData"` reflects the `"StorageData"` key in the `storage` dict. But it is arbitrary, I don't mind changing it to whatever you prefer.

There's a longer term project here to update our storage schema to something more intuitive (now that we don't need to maintain compatibility with Matpower), but that will require deeper design thinking.

danielolsen:
I found another edge case that also needs fixing: if there are no transformers in the upgraded branches then we get an error caused by trying to `apply` on an empty data frame. I will add some tests to detect the problem and a fix for it, and let you know when they're ready.

powersimdata/tests/mock_grid.py
BainanXia: I see. Either way. No strong preference ai this point.

powersimdata/design/investment/investment_costs.py
danielolsen: Done (simplified this one and the two after it).

powersimdata/design/investment/investment_costs.py
danielolsen: Done.

powersimdata/design/investment/investment_costs.py
danielolsen: Done.

powersimdata/design/investment/investment_costs.py
danielolsen: Done.

danielolsen:
> 
> 
> I found another edge case that also needs fixing: if there are no transformers in the upgraded branches then we get an error caused by trying to `apply` on an empty data frame. I will add some tests to detect the problem and a fix for it, and let you know when they're ready.

New tests and fixes are added, and all comments are addressed (I think).

BainanXia (approved):
Looks good. Thanks!

PR #452, opened 2021-04-20
feat: add 'cost' method to upgrade prioritization

danielolsen:
### Purpose
Allow user to prioritize branch upgrades based on estimated cost in the numerator (where the denominator remains the congestion shadow price at a certain quantile). See https://github.com/Breakthrough-Energy/PowerSimData/issues/439 for details.

### What the code is doing
The main logic is in the `if method == "cost":` block: we calculate the cost of upgrading each of the congested branches by 1x their original value, and then we divide the congestion metric by these values to get the branch upgrade ranking.

We also do a little simplification of the code where we calculate the bidirectional congestion, and move the application of the `composite_allow_list` filter up higher for better efficiency (so that we don't calculate quantiles for branches we know we're not going to select anyway), but these should have no effect on the results of the code.

### Time estimate
10 minutes. Most of the work to enable this feature was actually contained in https://github.com/Breakthrough-Energy/PowerSimData/pull/450.

### DIFF
--- a/powersimdata/design/transmission/upgrade.py
+++ b/powersimdata/design/transmission/upgrade.py
@@ -1,5 +1,6 @@
 import pandas as pd
 
+from powersimdata.design.investment.investment_costs import _calculate_ac_inv_costs
 from powersimdata.input.grid import Grid
 from powersimdata.network.model import area_to_loadzone
 from powersimdata.utility.distance import haversine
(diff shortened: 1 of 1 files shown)


powersimdata/design/transmission/upgrade.py
```diff
@@ -293,6 +293,17 @@ def _identify_mesh_branch_upgrades(
...
+        # Calculate costs for an upgrade dataframe containing only composite_allow_list
+        base_grid = Grid(
+            ref_scenario.info["interconnect"], ref_scenario.info["grid_model"]
+        )
+        base_grid.branch = base_grid.branch.filter(items=congested_indices, axis=0)
+        upgrade_costs = _calculate_ac_inv_costs(base_grid, sum_results=False)
+        # Merge the individual line/transformer data into a single Series
+        merged_upgrade_costs = pd.concat(
+            [v.Cost for v in upgrade_costs.values()], sort=True
+        )
```
BainanXia, line 306: What we are trying to achieve by passing `sort=True` here?

danielolsen: Originally in the `branch` data frame, branch IDs are sorted. If we do not add `sort=True` here, then we will get either all Line branch IDs and then all Transformer/TransformerWinding branch IDs in the concatenated data frame, or vice versa. Now that I think about it though, it probably doesn't matter, because pandas should automatically use the indices for the element-wise division, and we never use `merged_upgrade_costs` after that. I will verify and remove.

BainanXia: Agree with your second point, I think it won't matter given the pandas automatic index matching. The main concern I had here is `sort=True` in pd.concat will only sort non-concatenation axis (not the values), here is the column, which is trivial given there is only one column in pd.Series.

danielolsen: Confirmed: quick testing showed no change in with vs. without the `sort=True`, so it's been removed.

BainanXia (approved):
Very nice. Thanks.

PR #453, opened 2021-04-20
chore: create feature request issue template

rouille:
This is the automatic **Feature Request Issue Template** from GitHub. I did not write anything. We can change it. Pandas has a template that is quite nice (see https://github.com/pandas-dev/pandas/issues/new/choose)

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,36 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: Feature request
+labels: feature request
+assignees: ahurli, BainanXia, danielolsen, jon-hagg, rouille
+
+---
+
+ # :rocket:
+
+- [ ] Is your feature request essential for your project?
+
+
+### Describe the workflow you want to enable
+A clear and concise description of what can be enhanced, e.g., "I wish I could do [...]"
+
+### Describe your proposed implementation
+This should provide a description of the feature request, e.g.:
+* "The class `Foo` should have a new method `bar` that allows to [...]"
+* "Function `foo` needs a new arguments `bar` to set [...]"
+* "Create a new function `foo` to calculate [...]"
+
+If applicable, try to write a docstring for the desired feature. To illustrate, if you would like to add a new function in a module, provide:
+* the name of the function
+* a description of the task accomplished by the function
+* a list of the input and output parameters together with their types (e.g., `int`,
+  `str`, `pandas.DataFrame`, etc.) and a short description of its/their meaning
+
+### Describe alternatives you've considered, if relevant
+This should provide a description of any alternative solutions or features you've
+considered.
+
+### Additional context
+Add any other context or screenshots in this section, e.g., a plot from an article you
+believe would clearly communicate results.


BainanXia:
Should we add a metric of "urgency" in the template, something like "when you want this to be done" or "How fast you want this to happen", given most of the feature requests are tied up with project timelines? However, we should hope people take this seriously given this is a public repo and we can easily get tons of urgent requests.

rouille:
I prefer what `pandas` came up with:

#### Is your feature request related to a problem?

[this should provide a description of what the problem is, e.g. "I wish I could use pandas to do [...]"]

#### Describe the solution you'd like

[this should provide a description of the feature request, e.g. "`DataFrame.foo` should get a new parameter `bar` that [...]", try to write a docstring for the desired feature]

#### API breaking implications

[this should provide a description of how this feature will affect the API]

#### Describe alternatives you've considered

[this should provide a description of any alternative solutions or features you've considered]

#### Additional context

[add any other context, code examples, or references to existing implementations about the feature request here]

```python
# Your code here, if applicable

```

kasparm:
@BainanXia good point regarding prioritization. I think we should detach the prioritization from the request. In large communities the users can vote up features that are requested. As we start having more user we can go that path. For now I think we should decide internally what has high priority and what not.

rouille:
This is the one from `scikit-learn`:



#### Describe the workflow you want to enable

#### Describe your proposed solution

#### Describe alternatives you've considered, if relevant

#### Additional context

ahurli:
I agree prioritization can easily get abused, but I like the idea of having a place to differentiate between features that make it easier or more intuitive to use vs features that are blocking a current project. Maybe just a question like: "Is the absence of this feature currently blocking a research problem? If so, please describe the problem."

jon-hagg (approved):

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,32 @@
...
+### Describe the workflow you want to enable
+A clear and concise description of what can be enhanced, e.g., "I wish I could do [...]"
+
+### Describe your proposed implementation
+This should provide a description of the feature request, e.g.:
+* "The class `Foo` should have a new method `bar` that allows to [...]"
+* "Function `foo` needs a new arguments `bar` to set [...]"
+* "Create a new function `foo` to calculate [...]"
+
+If applicable, try to write a docstring for the desired feature.
```
danielolsen, line 24: `If applicable, try to write a docstring for the desired feature. If you are not familiar with writing docstrings, you can specify a list of inputs and outputs as necessary, with the types of each (e.g. floating-point number, integer, list, dict-like mapping, data frame [2D table]).`?

BainanXia: > `If applicable, try to write a docstring for the desired feature. If you are not familiar with writing docstrings, you can specify a list of inputs and outputs as necessary, with the types of each (e.g. floating-point number, integer, list, dict-like mapping, data frame [2D table]).`?

Would that be harder than writing doc strings assuming doc strings referring to the description of what a function do not including the I/O?

danielolsen: I guess for us, a docstring is two parts: a very short user-facing description of what the function does, plus a detailed descriptions of the inputs and outputs. Ideally, we want to encourage the requester to give us both, even if they have never heard of a docstring before.

BainanXia: Agree. Maybe we can be more explicit with what we mean by doc strings here:

- A description of what the function does
- I/O variable definitions 

And ask the user to provide as much info as they could come up with.

rouille: I have elaborated on docstring in 468297f8e35e206a979225ab361e363cd55bcb7f

BainanXia (approved):
This is nice.

danielolsen (approved):

PR #459, opened 2021-04-22
feat: os agnostic file existence check

jon-hagg:
### Purpose
Small chunk of work to support windows natively. This one splits the file existence check, (basically an `ls` command) into one for `LocalDataAccess` which is os independent, and the one for `SSHDataAccess` which remains unchanged. 

### What the code is doing
Split implementation from base class, and some related cleanup.

### Testing
Ran through creating and preparing a scenario in a container, so we exercise the base functionality and local implementation. Again, the server use case is not changed.

### Time estimate
5 min

### DIFF
--- a/powersimdata/data_access/data_access.py
+++ b/powersimdata/data_access/data_access.py
@@ -56,6 +56,14 @@ class DataAccess:
         command = CommandBuilder.remove(target, recursive, force)
         return self.execute_command(command)
 
+    def _exists(self, filepath):
+        """Return whether the file exists
+
+        :param str filepath: the path to the file
+        :return: (*bool*) -- whether the file exists
+        """
+        raise NotImplementedError
+
     def _check_file_exists(self, filepath, should_exist=True):
         """Check that file exists (or not) at the given path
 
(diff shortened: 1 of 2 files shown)


powersimdata/data_access/data_access.py
```diff
@@ -56,18 +56,26 @@ def remove(self, target, recursive=False, force=False):
         command = CommandBuilder.remove(target, recursive, force)
         return self.execute_command(command)
 
+    def _exists(self, filepath):
+        """Return whether the file exists
+
+        :param str filepath: the path to the file
+        :return: (*bool*) -- whether the file exists
+        """
+        raise NotImplementedError
```
BainanXia, line 65: I presume the reason we put this down here is for situations in the future, we have other classes that inherit `DataAccess` without overwriting `_exists` method?

powersimdata/data_access/data_access.py
```diff
@@ -56,18 +56,26 @@ def remove(self, target, recursive=False, force=False):
...
 
         :param str filepath: the full path to the file
         :param bool should_exist: whether the file is expected to exist
         :raises OSError: if the expected condition is not met
         """
-        _, _, stderr = self.execute_command(CommandBuilder.list(filepath))
+        result = self._exists(filepath)
         compare = operator.ne if should_exist else operator.eq
-        if compare(len(stderr.readlines()), 0):
+        if compare(result, True):
```
BainanXia, line 76: This is smart.

powersimdata/data_access/data_access.py
jon-hagg: Yep, that's pretty much it. I'd say a couple related reasons are 1) make it so that `self._exists` is defined in the context we're referencing it (make editor diagnostics happy), and 2) it's a form of documentation, so you know what needs to be implemented in a subclass.

BainanXia (approved):
Looks good.

powersimdata/data_access/data_access.py
```diff
@@ -56,18 +56,26 @@ def remove(self, target, recursive=False, force=False):
...
             msg = "not found" if should_exist else "already exists"
```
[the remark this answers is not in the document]
jon-hagg: Yeah it's a bit head bending.. I figured it's worth it if the usage is clear externally (treating the function as a black box). In your example, if `should_exist==False` and `result==False`, then we'd have the `compare` operator set to `==` so it becomes `if False == True`, and we won't enter the if statement.

PR #461, opened 2021-04-22
feat: os agnostic rm functionality

jon-hagg:
### Purpose
Another small chunk to support windows natively. This one splits the `remove` method so that the one in `LocalDataAccess` is os independent. 

### What the code is doing
* Add confirmation prompt before running `rm` command on server, and before deleting local glob matches
* Move the error check on stderr inside the `SSHDataAccess` implementation, since there is no analog for local version (return types have to match)
* Remove `force` parameter and hard code the resulting `-f` option since we always pass it to the server, and it's not needed locally
* Change value passed to `recursive` in move.py so it matches what we do in delete.py
* Rename test file to match module under test

### Testing
The main thing to test is the new glob used in the local implementation. I did this manually by creating some nested folders and checking the files matched. Also tested with the confirmation prompt using the server based container setup.

### Time estimate
15 min


powersimdata/data_access/data_access.py
```diff
@@ -196,6 +195,18 @@ def makedir(self, relative_path):
...
 
+    def remove(self, target, recursive=False):
+        """Remove target using rm semantics
+
+        :param str target: path to remove
+        :param bool recursive: delete directories recursively
+        """
+        if recursive:
+            target = os.path.join(target, "**")
+        files = glob.glob(target, recursive=recursive)
```
BainanXia, line 206: According to the documentation, it seems `glob` is used for Unix shell only?

jon-hagg: I believe "unix style" is just a reference to the origin. There are separate sections in the docs for [windows](https://docs.python.org/3/library/windows.html) and [unix](https://docs.python.org/3/library/unix.html) specific features. Will do a test to confirm though.

jon-hagg: Looks like there could be some inconsistencies (see this [post](https://python-forum.io/Thread-Version-of-glob-for-that-Supports-Windows-Wildcards)) but it _should_ work for our current use case.

BainanXia: Thanks for looking into that. I think it should be fine as long as it works for our current use case. As said, we can always revisit it in the future if our usage goes beyond the available scope.

BainanXia (approved):
Not sure how to test it throughout. But some native tests locally on my mac work as expected.

powersimdata/data_access/data_access.py
danielolsen: If this fails, it seems like we can use regex workarounds. I don't think we use any `[` or `]` characters _within_ our data directories, so hopefully any of those characters in parent directories won't be a problem.

powersimdata/data_access/data_access.py
```diff
@@ -196,6 +195,18 @@ def makedir(self, relative_path):
...
+        for f in files:
+            os.remove(f)
```
ahurli, line 217: Does this fail if when trying to remove a directory? Specifically, when it's used in the `delete` module for deleting the `tmp` folder on line 58? I'm not sure it's an edge case we necessarily want to worry about right now (deleting a scenario in a local setup), but might be good to note.

jon-hagg: Yeah, looks like it would fail due to the glob matching the `tmp/scenario_n` folder itself. Will fix.

jon-hagg: Done

danielolsen: Let's make extra sure we don't accidentally delete the `tmp` folder itself.

jon-hagg: For sure - I think we're safe there. We have the confirmation prompt by default, and recursive delete is only called from two places which specify the scenario folder within `tmp`.

powersimdata/scenario/delete.py
```diff
@@ -43,33 +46,24 @@ def delete_scenario(self):
         # Delete links to base profiles on server
         print("--> Deleting scenario input data on server")
         target = posixpath.join(self.path_config.input_dir(), wildcard)
-        _, _, stderr = self._data_access.remove(target, recursive=False, force=True)
-        if len(stderr.readlines()) != 0:
-            raise IOError("Failed to delete scenario input data on server")
+        self._data_access.remove(target, recursive=False, confirm=confirm)
 
         # Delete output profiles
```
rouille, line 51: We don't only have profiles so I would use `Delete output data`

powersimdata/scenario/move.py
```diff
@@ -67,36 +68,37 @@ def __init__(self, data_access, scenario_info):
         self._scenario_info = scenario_info
```
rouille, line 68: in the docstring: `scenario` --> `scenario_info`

powersimdata/scenario/delete.py
```diff
@@ -43,33 +46,24 @@ def delete_scenario(self):
         # Delete links to base profiles on server
```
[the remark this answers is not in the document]
jon-hagg: In this case, I think I'll remove the comments about deleting input/output data since they aren't adding anything new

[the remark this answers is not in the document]
rouille: Sounds good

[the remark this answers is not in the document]
jon-hagg: Done in latest commit

rouille (approved):
Looks good

PR #463, opened 2021-04-23
feat: improve flexibility of MW-miles calculation, for Create state and new branches

### LINKED ISSUE #457: Allow calculate_mw_miles to be used for Scenario in Create state
Currently, this fails with the following traceback:
```python
c:\python38\lib\site-packages\powersimdata\design\transmission\mwmiles.py in calculate_mw_miles(scenario, exclude_branches)
     15     """
     16 
---> 17     original_grid = Grid(scenario.info["interconnect"].split("_"))
     18     ct = scenario.state.get_ct()
     19     upgrades = _calculate_mw_miles(original_grid, ct, exclude_branches)

c:\python38\lib\site-packages\powersimdata\scenario\scenario.py in __getattr__(self, name)
     56             return getattr(self.state, name)
     57         elif hasattr(self.state, "__getattr__"):
---> 58             return self.state.__getattr__(name)
     59         else:
     60             raise AttributeError(

c:\python38\lib\site-packages\powersimdata\scenario\create.py in __getattr__(self, name)
     67             raise AttributeError(f"Call set_grid first to access {name} attribute")
     68         else:
---> 69             raise AttributeError(f"Create object has no attribute {name}")
     70 
     71     def __setattr__(self, name, value):

AttributeError: Create object has no attribute info
```

It is using our fancy new state attribute passthrough, but the failure really originates because we don't have an `info` attribute for the Scenario object. We can use a workaround via the `_calculate_mw_miles` function which takes a Grid object, instead of a Scenario, but it would be nice if we could avoid the use of this private function, which mostly exists for testability purposes (since instantiating a base Grid for a MockScenario is tricky).

### LINKED ISSUE #458: Include new branches in calculations of MW-miles of upgrades
Currently, the code only does calculations from up-scaled branches, not new branches.

danielolsen:
### Purpose
- Allow MW-miles calculations to be done on scenarios in Create state (closes #457).
- Include new branches in calculations of MW-miles (closes #458).
- ~Allow branches not in the base grid to be scaled via the change table.~ **EDIT**: removed. See comment: https://github.com/Breakthrough-Energy/PowerSimData/pull/463#issuecomment-826925439

### What the code is doing
- In **create.py** and **scenario.py**: we add the `info` parameter to the `Scenario` (see #457 for a discussion).
- ~In **transform_grid.py**: we move branch and DC line scaling after branch and DC line additions, to enable scaling of newly-added branches (we could equivalently scale the entries in `"new_branch"` and "`new_dcline"`, but this way allows `scale_congested_mesh_branches` to continue to work without modifications even when it identifies that a branch not in the base grid should be upgraded).~ **EDIT**: removed. See comment: https://github.com/Breakthrough-Energy/PowerSimData/pull/463#issuecomment-826925439
- In **mwmiles.py**: we make use of the `TransformGrid` class to be able to easily calculate differences in branch `rateA` for both scaled and new branches, and to automatically look up branch types, lats, and lons.
- In **test_mwmiles.py**: we add impedances to the fake data so that `TransformGrid` works properly.

### Testing
All unit tests still pass, and calculations work as expected on Scenarios with new branches (see 3822 or 3828 for examples). Testing on previously-analyzed scenarios with lots of branch upgrades (existing branches only) still produces identical results.

### Time estimate
15-30 minutes.

### DIFF
--- a/powersimdata/design/transmission/mwmiles.py
+++ b/powersimdata/design/transmission/mwmiles.py
@@ -1,4 +1,5 @@
 from powersimdata.input.grid import Grid
+from powersimdata.input.transform_grid import TransformGrid
 from powersimdata.utility.distance import haversine
 
 
@@ -51,20 +52,30 @@ def _calculate_mw_miles(original_grid, ct, exclude_branches=None):
     else:
         raise TypeError("exclude_branches must be None, list, tuple, or set")
 
-    base_branch = original_grid.branch
-    upgraded_branches = ct["branch"]["branch_id"]
-    for b, v in upgraded_branches.items():
+    base_branch_ids = set(original_grid.branch.index)
+    upgraded_branch_ids = set(ct["branch"]["branch_id"].keys())
+    transformed_branch = TransformGrid(original_grid, ct).get_grid().branch
+    if "new_branch" in ct:
+        upgraded_branch_ids |= set(transformed_branch.index) - base_branch_ids
+    for b in upgraded_branch_ids:
         if b in exclude_branches:
             continue
-        # 'upgraded' capacity is v-1 because a scale of 1 = an upgrade of 0
-        upgraded_capacity = base_branch.loc[b, "rateA"] * (v - 1)
-        device_type = base_branch.loc[b, "branch_device_type"]
+        if b in base_branch_ids:
+            # 'upgraded' capacity is (scale - 1) because a scale of 1 = an upgrade of 0
+            scale = ct["branch"]["branch_id"][b]
+            upgraded_capacity = transformed_branch.loc[b, "rateA"] / scale * (scale - 1)
+        else:
+            upgraded_capacity = transformed_branch.loc[b, "rateA"]
+        device_type = transformed_branch.loc[b, "branch_device_type"]
         if device_type == "Line":
             from_coords = (
-                base_branch.loc[b, "from_lat"],
-                base_branch.loc[b, "from_lon"],
+                transformed_branch.loc[b, "from_lat"],
+                transformed_branch.loc[b, "from_lon"],
             )
-            to_coords = (base_branch.loc[b, "to_lat"], base_branch.loc[b, "to_lon"])
+            to_coords = (
+                transformed_branch.loc[b, "to_lat"],
+                transformed_branch.loc[b, "to_lon"],
+            )
             addtl_mw_miles = upgraded_capacity * haversine(from_coords, to_coords)
             upgrades["mw_miles"] += addtl_mw_miles
             upgrades["num_lines"] += 1
@@ -75,6 +86,6 @@ def _calculate_mw_miles(original_grid, ct, exclude_branches=None):
             upgrades["transformer_mw"] += upgraded_capacity
             upgrades["num_transformers"] += 1
         else:
-            raise Exception("Unknown branch: " + str(b))
+            raise Exception("Unknown branch type: " + str(b))
 
     return upgrades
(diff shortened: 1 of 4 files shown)


BainanXia:
I have two questions regarding the logic: 
1. When adding new branches, what if the user doesn't want it be scaled given the `Pmax` is part of user's inputs, assuming we have branch scaling in a zone (currently we never did that though).
2. `scale_congested_meshed_branches` will never identify a branch not in the "base_grid" right? It takes `ref_scenario` as parameter, in which case the `base_grid` is the grid of the reference scenario. Any branches not in the grid of reference scenario can't be congested either, since the simulation hasn't been run yet.

danielolsen:
> I have two questions regarding the logic:
> 
>     1. When adding new branches, what if the user doesn't want it be scaled given the `Pmax` is part of user's inputs, assuming we have branch scaling in a zone (currently we never did that though).
> 
>     2. `scale_congested_meshed_branches` will never identify a branch not in the "base_grid" right? It takes `ref_scenario` as parameter, in which case the `base_grid` is the grid of the reference scenario. Any branches not in the grid of reference scenario can't be congested either, since the simulation hasn't been run yet.

For (1), I agree, maybe we want to keep TransformGrid the way it is so that there is no confusion where the user specifies one branch capacity and then it gets zone-scaled. I remember conversations with you and @rouille along those same lines for plants.

For (2), `scale_congested_meshed_branches` can identify any branch in a Grid from a previously-run scenario. We may have added a branch, run the scenario to see where the congestion is, and then we want to upgrade some branches, one of which is one that is not present in the base grid. I just ran into that situation myself. We can implement this logic without changing TransformGrid by modifying `powersimdata.design.transmission.upgrade.increment_branch_scaling` so that it can upgrade added branches as well using similar logic, once they're identified in `powersimdata.design.transmission.upgrade._identify_mesh_branch_upgrades`.

danielolsen:
I've removed the re-ordering of the TransformGrid so that this PR can be more focused. I have follow-up ideas about the TransformGrid, but I will save them for a separate issue.

BainanXia (approved):
Looks good.

powersimdata/design/transmission/mwmiles.py
```diff
@@ -75,6 +86,6 @@ def _calculate_mw_miles(original_grid, ct, exclude_branches=None):
             upgrades["transformer_mw"] += upgraded_capacity
             upgrades["num_transformers"] += 1
         else:
-            raise Exception("Unknown branch: " + str(b))
+            raise Exception("Unknown branch type: " + str(b))
```
ahurli, line 89: Would it be helpful to include the unknown branch type `device_type` in the error message and clarify that `b` is the branch id?

danielolsen: This else is more for debugging than anything else, since the code should never in normal operation get here. There are no branch device types in our current grid model that are not covered by these three options.

powersimdata/design/transmission/mwmiles.py
```diff
@@ -51,20 +52,30 @@ def _calculate_mw_miles(original_grid, ct, exclude_branches=None):
...
         raise TypeError("exclude_branches must be None, list, tuple, or set")
 
-    base_branch = original_grid.branch
-    upgraded_branches = ct["branch"]["branch_id"]
-    for b, v in upgraded_branches.items():
+    base_branch_ids = set(original_grid.branch.index)
+    upgraded_branch_ids = set(ct["branch"]["branch_id"].keys())
+    transformed_branch = TransformGrid(original_grid, ct).get_grid().branch
+    if "new_branch" in ct:
+        upgraded_branch_ids |= set(transformed_branch.index)
```
jon-hagg, line 59: Am I reading this correctly - we could technically use `upgraded_branch_ids = set(transformed_branch.index)` to begin with, but in the case where there are no new branches, it's easier to just grab the scaled branch ids from the change table since there will be fewer to consider?

danielolsen: I made a change when simplifying something here. We don't want the full transformed branch index, we want the entries in the change table plus any _new_ entries, which is the difference between the transformed branch and the base one. I've pushed a commit to fix.

jon-hagg (approved):

PR #466, opened 2021-04-27
feat: allow scaling of added elements

### LINKED ISSUE #464: Allow scaling of added elements
### Background

Currently, the process of using the TransformGrid class to create a new grid from a base grid and a change table operates in the order:
1. Scale generator attributes, and branch and DC line capacities.
2. Add new elements to the Grid.

Long ago, I believe we had discussed that we wanted to do scaling before addition, so that if we perform zone scaling for generators, and then add a generator of a certain capacity, that capacity does not get changed by zone scaling (unintuitive behavior from the user's side).

### Desired functionality

As originally mentioned in https://github.com/Breakthrough-Energy/PowerSimData/pull/463, I think it would be useful to be able to scale new elements as well. Use cases:
- We perform the congested branch upgrade heuristic on a Scenario that has had new branches added, and we potentially want to upgrade these branches. However, the upgrade heuristic works by adding scaling factors to the change table. A change table that has a scaling factor for an element ID that is not in the base grid will cause the TransformGrid method to fail, even if that element ID would be created in step (2) above.
- If I am a user who is manually examining a previously-run Grid, intending to copy over its change table and apply additional scaling factors to set up a follow-up scenario, I don't want to have to think about whether a plant or branch was part of a base Grid when deciding how to modify the ScenarioGrid.

### Implementation idea

Change the implementation of the `TransformGrid._apply_change_table` method, so that the new order is:
1. Scale generators and branches _by zone scaling factors only_.
2. Add new elements to the grid.
3. Scale elements _by ID only_.

I believe that this ordering avoids any un-intuitive scaling of newly-added elements, but still allows explicit (i.e., by ID) scaling of these elements if desired, to achieve the desired functionality.

danielolsen:

### Purpose
Allow scaling to be applied to elements that are added (i.e., not in the base grid). Closes #464.

### What the code is doing
We separate the scaling functions into scaling-by-zone and scaling-by-ID. We apply scaling-by-zone, then additions, then scaling-by-ID.

### Testing
Tested on Scenario 3856, which was created using test code along these lines, so the change table that's on the server is not valid with the current implementation of TransformGrid. With the new code, it works.
```python
from powersimdata.design.transmission.mwmiles import calculate_mw_miles
scenario = Scenario(3856)
calculate_mw_miles(scenario)
```

### Time estimate
15 minutes.

### DIFF
--- a/powersimdata/input/change_table.py
+++ b/powersimdata/input/change_table.py
@@ -133,7 +133,7 @@ class ChangeTable(object):
         """
         self.grid = grid
         self.ct = {}
-        self.new_bus_cache = {}
+        self._new_element_caches = {k: {} for k in {"branch", "bus", "dcline", "plant"}}
 
     @staticmethod
     def _check_resource(resource):
@@ -262,10 +262,8 @@ class ChangeTable(object):
                 if len(self.ct[ct_key]["zone_id"]) == 0:
                     self.ct.pop(ct_key)
             if plant_id is not None:
-                plant_id_interconnect = set(
-                    self.grid.plant.groupby("type").get_group(resource).index
-                )
-                diff = set(plant_id.keys()).difference(plant_id_interconnect)
+                anticipated_plant = self._get_df_with_new_elements("plant")
+                diff = set(plant_id.keys()) - set(anticipated_plant.index)
                 if len(diff) != 0:
                     err_msg = f"No {resource} plant(s) with the following id: "
                     err_msg += ", ".join(sorted([str(d) for d in diff]))
@@ -354,6 +352,7 @@ class ChangeTable(object):
             is (are) the id of the line(s) and the associated value is the
             scaling factor for the increase/decrease in capacity of the line(s).
         """
+        anticipated_branch = self._get_df_with_new_elements("branch")
         if bool(zone_name) or bool(branch_id) is True:
             if "branch" not in self.ct:
                 self.ct["branch"] = {}
@@ -368,8 +367,7 @@ class ChangeTable(object):
                 for z in zone_name.keys():
                     self.ct["branch"]["zone_id"][self.grid.zone2id[z]] = zone_name[z]
             if branch_id is not None:
-                branch_id_interconnect = set(self.grid.branch.index)
-                diff = set(branch_id.keys()).difference(branch_id_interconnect)
+                diff = set(branch_id.keys()) - set(anticipated_branch.index)
                 if len(diff) != 0:
                     print("No branch with the following id:")
                     for i in list(diff):
@@ -394,7 +392,8 @@ class ChangeTable(object):
         """
         if "dcline" not in self.ct:
             self.ct["dcline"] = {}
-        diff = set(dcline_id.keys()).difference(set(self.grid.dcline.index))
+        anticipated_dcline = self._get_df_with_new_elements("dcline")
+        diff = set(dcline_id.keys()) - set(anticipated_dcline.index)
         if len(diff) != 0:
             print("No dc line with the following id:")
             for i in list(diff):
@@ -506,7 +505,7 @@ class ChangeTable(object):
             "terminal_min",
             "terminal_max",
         }
-        anticipated_bus = self._get_new_bus()
+        anticipated_bus = self._get_df_with_new_elements("bus")
         for i, storage in enumerate(info):
             self._check_entry_keys(storage, i, "storage", required, None, optional)
             if storage["bus_id"] not in anticipated_bus.index:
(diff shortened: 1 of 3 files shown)


BainanXia (approved):
This is consistent with our previous discussion.

danielolsen:
I added a test of the TransformGrid functionality, which fails initially but passes with the new functionality.

Importantly, for this test, we have to edit the change table directly, since the change table method won't let us manually scale a newly-created branch yet. So this code means that Scenario 3856 works, since the change table was already created via the scale congested mesh branches method, but that the same scaling could not be applied manually by the user.

BainanXia:
> I added a test of the TransformGrid functionality, which fails initially but passes with the new functionality.
> 
> Importantly, for this test, we have to edit the change table directly, since the change table method won't let us manually scale a newly-created branch yet. So this code means that Scenario 3856 works, since the change table was already created via the scale congested mesh branches method, but that the same scaling could not be applied manually by the user.

I think we should implement `self._get_new_ac_line()` and `self._get_new_dc_line()` follow the same logic of `self._get_new_bus()` in change table.

danielolsen:
> > I added a test of the TransformGrid functionality, which fails initially but passes with the new functionality.
> > Importantly, for this test, we have to edit the change table directly, since the change table method won't let us manually scale a newly-created branch yet. So this code means that Scenario 3856 works, since the change table was already created via the scale congested mesh branches method, but that the same scaling could not be applied manually by the user.
> 
> I think we should implement `self._get_new_ac_line()` and `self._get_new_dc_line()` follow the same logic of `self._get_new_bus()` in change table.

I got started on this, with `self._get_new_branch`. I used `branch` rather than `ac_line` since what we're fetching is a modified (if necessary) `branch` attribute data frame. I'll add `self._get_new_dcline` and `self._get_new_plant`, and I think that should complete the functionality.

danielolsen:
Caches for new {buses, branches, DC lines, and plants} are added.

powersimdata/input/change_table.py
```diff
@@ -782,16 +781,19 @@ def add_bus(self, info):
             self.ct["new_bus"] = []
         self.ct["new_bus"] += new_buses
 
-    def _get_new_bus(self):
-        if "new_bus" not in self.ct:
-            return self.grid.bus
-        new_bus_tuple = tuple(tuple(sorted(b.items())) for b in self.ct["new_bus"])
-        if new_bus_tuple in self.new_bus_cache:
-            return self.new_bus_cache[new_bus_tuple]
+    def _get_df_with_new_elements(self, table):
```
BainanXia, line 784: Missing doc strings for table and return values.

danielolsen: Done.

BainanXia:
I've tested this with several corner cases that are made up myself. It works as expected. One minor concern is the cache can't be overwritten with new updates if one would like to track the changes step by step when building a change table, but that's not a typical user story.

danielolsen:
> I've tested this with several corner cases that are made up myself. It works as expected. One minor concern is the cache can't be overwritten with new updates if one would like to track the changes step by step when building a change table, but that's not a typical user story.

The original idea of the cache was that we wanted to make sure that we could add new elements step by step, without needing to instantiate a new TransformGrid each time (to get indices, voltages, interconnects etc. for the new buses), since the new bus info won't change as long as the `"new_bus"` entries in the change table don't change.

I'm not sure if I fully understand your user story, but I think for user who wants to track the grid changes step-by-step, they could call `scenario.get_grid()` between each change table method call. This cache and the function that uses it are not meant to be user-facing, which is why I added a leading underscore when I renamed the cache.

BainanXia:
> > I've tested this with several corner cases that are made up myself. It works as expected. One minor concern is the cache can't be overwritten with new updates if one would like to track the changes step by step when building a change table, but that's not a typical user story.
> 
> The original idea of the cache was that we wanted to make sure that we could add new elements step by step, without needing to instantiate a new TransformGrid each time (to get indices, voltages, interconnects etc. for the new buses), since the new bus info won't change as long as the `"new_bus"` entries in the change table don't change.
> 
> I'm not sure if I fully understand your user story, but I think for user who wants to track the grid changes step-by-step, they could call `scenario.get_grid()` between each change table method call. This cache and the function that uses it are not meant to be user-facing, which is why I added a leading underscore when I renamed the cache.

That's exactly what I'm referring to. I was lazy and didn't use `get_grid()` but using `_get_df_with_new_elements` instead and found the changes were not as expected. Then I realized that I shouldn't track changes in that way. That's what I mean by "not a typical user story". I think this is good to go.

PR #474, opened 2021-05-03
feat: allow congested mesh branches to be ranked by mean congestion

### LINKED ISSUE #439: Improve flexibility of congested branch upgrade algorithm

danielolsen:

### Purpose
Allow congested mesh branches to be ranked by mean congestion, rather than only by a quantile value of congestion. Closes #439 (we will create a separate, more detailed issue for the third bullet of #439, which requires more design thought).

### What the code is doing
Previously, we had a `method` parameter, which only applied to the denominator of the `(congestion metric) / (upgrade cost proxy)` ratio. Now, we have `congestion_metric` and `cost_metric` (replaces `method`). Most of the code changes are simply renaming `method` -> `cost_metric`, and elaborating in the documentation. Within `_identify_mesh_branch_upgrades`, we call `.mean()` rather than `.quantile()` on the congestion data frame as applicable, and we rename some variables so that we create a `congestion_metric_values` variable regardless of whether it's a quantile or a mean (previously, this variable had been called `quantile_cong_abs`).

### Testing
Previous unit tests pass, new ones have been added which return different results if we select `"mean"` rather than `"quantile"` as the `congestion_metric`.

### Time estimate
15-30 minutes. Functional code changes are <10 lines, but a few other things are renamed or reorganized for clarity.

### DIFF
--- a/powersimdata/design/transmission/upgrade.py
+++ b/powersimdata/design/transmission/upgrade.py
@@ -197,12 +197,16 @@ def scale_congested_mesh_branches(
     upgrade_n=100,
     allow_list=None,
     deny_list=None,
-    quantile=0.95,
+    congestion_metric="quantile",
+    cost_metric="branches",
+    quantile=None,
     increment=1,
-    method="branches",
 ):
-    """Use a reference scenario as a baseline for branch scaling, and further
-    increment branch scaling based on observed congestion duals.
+    """Use a reference scenario to detect congested mesh branches (based on values of
+    the shadow price of congestion), and increase the capacity of a subset of them.
+    Branches are ranked by a ratio of (congestion metric) / (cost metric), and the top N
+    branches are selected for upgrading, where N is specified by ``upgraade_n``, with
+    each upgraded by their base capacity multiplied by ``increment``.
 
     :param powersimdata.input.change_table.ChangeTable change_table: the
         change table instance we are operating on.
@@ -211,10 +215,15 @@ def scale_congested_mesh_branches(
     :param int upgrade_n: the number of branches to upgrade.
     :param list/set/tuple/None allow_list: only select from these branch IDs.
     :param list/set/tuple/None deny_list: never select any of these branch IDs.
-    :param float quantile: the quantile to use to judge branch congestion.
-    :param [float/int] increment: branch increment, relative to original
+    :param str congestion_metric: numerator method: 'quantile' or 'mean'.
+    :param str cost_metric: denominator method: 'branches', 'cost', 'MW', or
+        'MWmiles'.
+    :param float quantile: if ``congestion_metric`` == 'quantile', this is the quantile
+        to use to judge branch congestion (otherwise it is unused). If None, a default
+        value of 0.95 is used, i.e. we evaluate the shadow price for the worst 5% of
+        hours.
+    :param float/int increment: branch increment, relative to original
         capacity.
-    :param str method: prioritization method: 'branches', 'MW', or 'MWmiles'.
     :return: (*None*) -- the change_table is modified in-place.
     """
     # To do: better type checking of inputs.
@@ -223,10 +232,11 @@ def scale_congested_mesh_branches(
     branches_to_upgrade = _identify_mesh_branch_upgrades(
         ref_scenario,
         upgrade_n=upgrade_n,
-        quantile=quantile,
-        method=method,
+        congestion_metric=congestion_metric,
+        cost_metric=cost_metric,
         allow_list=allow_list,
         deny_list=deny_list,
+        quantile=quantile,
     )
     _increment_branch_scaling(
         change_table, branches_to_upgrade, ref_scenario, value=increment
--- a/powersimdata/design/transmission/upgrade.py
+++ b/powersimdata/design/transmission/upgrade.py
@@ -246,24 +246,32 @@ def scale_congested_mesh_branches(
 def _identify_mesh_branch_upgrades(
     ref_scenario,
     upgrade_n=100,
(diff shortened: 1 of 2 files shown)


powersimdata/design/transmission/upgrade.py
```diff
@@ -211,10 +215,15 @@ def scale_congested_mesh_branches(
...
     :param list/set/tuple/None deny_list: never select any of these branch IDs.
-    :param float quantile: the quantile to use to judge branch congestion.
+    :param str congestion_metric: numerator method: 'quantile' or 'mean'.
+    :param str cost_metric: denominator method: 'branches', 'cost', 'MW', or
+        'MWmiles'.
+    :param float quantile: if ``congestion_metric`` == 'quantile', this is the quantile
+        to use to judge branch congestion (otherwise it is unused). If None, a default
+        value of 0.95 is used, i.e. we evaluate the shadow price for the worst 5% of
+        hours.
     :param [float/int] increment: branch increment, relative to original
```
rouille, line 225: [float/int] --> float/int

danielolsen: Done.

powersimdata/design/transmission/upgrade.py
```diff
@@ -197,12 +197,16 @@ def scale_congested_mesh_branches(
...
+    cost_metric="branches",
+    quantile=None,
     increment=1,
-    method="branches",
 ):
-    """Use a reference scenario as a baseline for branch scaling, and further
-    increment branch scaling based on observed congestion duals.
+    """Use a reference scenario to detect congested mesh branches (based on values of
+    the shadow price of congestion), and increase the capacity of a subset of them.
+    Branches are ranked by a ratio of (congestion metric) / (cost metric), and the top N
```
BainanXia, line 207: N -> ``upgrade_n``?

danielolsen: `the top N, where N is specified by \`\`upgrade_n\`\``?

danielolsen: I like 'top N' for a narrative description, even though I agree it is clearer to indicate that it's the `upgrade_n` parameter that specifies.

powersimdata/design/transmission/upgrade.py
```diff
@@ -248,52 +259,81 @@ def _identify_mesh_branch_upgrades(
...
-    :raises ValueError: if 'method' not recognized, or not enough branches to
-        upgrade.
+    :param str congestion_metric: numerator method: 'quantile' or 'mean'.
+    :param str cost_metric: denominator method: 'branches', 'cost', 'MW', or
+        'MWmiles'.
+    :param float quantile: if ``congestion_metric`` == 'quantile', this is the quantile
+        to use to judge branch congestion (otherwise it is unused). If None, a default
+        value of 0.95 is used, i.e. we evaluate the shadow price for the worst 5% of
+        hours.
+    :raises ValueError: if ``congestion_metric`` or ``cost_metric`` not recognized,
```
BainanXia, line 271: not recognized -> is not recognized

powersimdata/design/transmission/upgrade.py
```diff
@@ -248,52 +259,81 @@ def _identify_mesh_branch_upgrades(
...
-    num_congested = len(quantile_cong_abs)
+    num_congested = len(congested_indices)
     if num_congested < upgrade_n:
         err_msg = "not enough congested branches: "
         err_msg += f"{upgrade_n} desired, but only {num_congested} congested."
+        if congestion_metric == "quantile":
+            err_msg += (
+                f" The quantile used is {quantile}; increasing this value will increase"
+                " the number of branches which qualify as having 'frequent-enough'"
+                " congestion and can be selected for upgrades."
```
BainanXia, line 331: Nice.

BainanXia (approved):
Looks good. Thanks!

powersimdata/design/transmission/upgrade.py
danielolsen: Done.

powersimdata/design/transmission/upgrade.py
danielolsen: Done.

PR #475, opened 2021-05-03
Additional support for windows

jon-hagg:
### Purpose
Address several places that were not platform independent.

### What the code is doing
* Finished splitting up commands between local/ssh and removed the `execute_command` in the local case, as it relied on `/bin/bash`
* Add the `DataAccess.join` attribute according to the subclass, removing hard coded `posixpath.join` in calling code
* Added `get_base_dir` and `match_scenario_files` to encapsulate the logic of creating paths, along with tests for these
* Fix potential issue in `InputData` and `OutputData` which were relying on the values from `server_setup.py`, which had posixpath implicitly hard coded (e.g. `INPUT_DIR = data/input`). We delay the path join by using a tuple instead

### Testing
* Ran a scenario in docker (using `LocalDataAccess`)
* Created and prepared a scenario on the server and deleted it (using `SSHDataAccess`)
* Added a few unit tests

### Time estimate
20 mins

### DIFF
--- a/powersimdata/data_access/data_access.py
+++ b/powersimdata/data_access/data_access.py
@@ -219,9 +219,9 @@ class LocalDataAccess(DataAccess):
         :param str change_name_to: new name for file when copied to data store.
         """
         self._check_filename(file_name)
-        src = os.path.join(server_setup.LOCAL_DIR, file_name)
+        src = self.join(server_setup.LOCAL_DIR, file_name)
         file_name = file_name if change_name_to is None else change_name_to
-        dest = os.path.join(self.root, to_dir, file_name)
+        dest = self.join(self.root, to_dir, file_name)
         print(f"--> Moving file {src} to {dest}")
         self._check_file_exists(dest, should_exist=False)
         self.makedir(os.path.dirname(dest))
@@ -248,10 +248,10 @@ class LocalDataAccess(DataAccess):
         :param bool recursive: create directories recursively
         :param bool update: ignored
         """
-        self.makedir(dest)
         if recursive:
             shutil.copytree(src, dest)
         else:
+            self.makedir(dest)
             func = lambda s: shutil.copy(s, dest)  # noqa: E731
             LocalDataAccess._fapply(func, src)
 
@@ -366,7 +366,7 @@ class SSHDataAccess(DataAccess):
         to_dir = os.path.join(self.local_root, from_dir)
         os.makedirs(to_dir, exist_ok=True)
 
-        from_path = posixpath.join(self.root, from_dir, file_name)
+        from_path = self.join(self.root, from_dir, file_name)
         to_path = os.path.join(to_dir, file_name)
         self._check_file_exists(from_path, should_exist=True)
 
@@ -397,8 +397,8 @@ class SSHDataAccess(DataAccess):
             )
 
         file_name = file_name if change_name_to is None else change_name_to
-        to_dir = posixpath.join(self.root, "" if to_dir is None else to_dir)
-        to_path = posixpath.join(to_dir, file_name)
+        to_dir = self.join(self.root, "" if to_dir is None else to_dir)
+        to_path = self.join(to_dir, file_name)
         self.makedir(to_dir)
         self._check_file_exists(to_path, should_exist=False)
 
@@ -434,7 +434,7 @@ class SSHDataAccess(DataAccess):
         :param str relative_path: path relative to root
         :return: (*str*) -- the checksum of the file
         """
-        full_path = posixpath.join(self.root, relative_path)
+        full_path = self.join(self.root, relative_path)
         self._check_file_exists(full_path)
 
         command = f"sha1sum {full_path}"
@@ -455,9 +455,9 @@ class SSHDataAccess(DataAccess):
         self.move_to(file_name, change_name_to=backup)
 
         values = {
-            "original": posixpath.join(self.root, new_name),
-            "updated": posixpath.join(self.root, backup),
-            "lockfile": posixpath.join(self.root, "scenario.lockfile"),
+            "original": self.join(self.root, new_name),
+            "updated": self.join(self.root, backup),
+            "lockfile": self.join(self.root, "scenario.lockfile"),
             "checksum": checksum,
         }
 
--- a/powersimdata/data_access/data_access.py
+++ b/powersimdata/data_access/data_access.py
@@ -42,6 +42,13 @@ class DataAccess:
         raise NotImplementedError
 
     def get_base_dir(self, kind, backup=False):
+        """Get path to given kind relative to instance root
+
+        :param str kind: one of {input, output, tmp}
+        :param bool backup: pass True if relative to backup root dir
+        :raises ValueError: if kind is invalid
+        :return: (*str*) -- the specified path
+        """
         _allowed = ("input", "output", "tmp")
         if kind not in _allowed:
             raise ValueError(f"Invalid 'kind', must be one of {_allowed}")
@@ -52,6 +59,13 @@ class DataAccess:
         return self.join(root, "data", kind)
 
     def match_scenario_files(self, scenario_id, kind, backup=False):
+        """Get path matching the given kind of scenario data
+
+        :param int/str scenario_id: the scenario id
+        :param str kind: one of {input, output, tmp}
+        :param bool backup: pass True if relative to backup root dir
+        :return: (*str*) -- the specified path
+        """
         base_dir = self.get_base_dir(kind, backup)
         if kind == "tmp":
             return self.join(base_dir, f"scenario_{scenario_id}")
@@ -514,8 +528,9 @@ class SSHDataAccess(DataAccess):
         return len(stderr.readlines()) == 0
 
     def close(self):
-        """Close the connection that was opened when the object was created."""
-        self.ssh.close()
+        """Close the connection if one is open"""
+        if self._ssh is not None:
+            self._ssh.close()
 
 
 def progress_bar(*args, **kwargs):
--- a/powersimdata/data_access/data_access.py
+++ b/powersimdata/data_access/data_access.py
@@ -18,6 +18,12 @@ from powersimdata.utility.helpers import CommandBuilder
 class DataAccess:
     """Interface to a local or remote data store."""
 
+    def __init__(self, root=None):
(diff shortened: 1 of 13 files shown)


powersimdata/data_access/data_access.py
```diff
@@ -14,10 +14,22 @@
 from powersimdata.utility import server_setup
 from powersimdata.utility.helpers import CommandBuilder
 
+_dirs = {
+    "tmp": (server_setup.EXECUTE_DIR,),
```
jon-hagg, line 18: using a tuple here makes it so the logic in `get_base_dir` doesn't require an if/else

powersimdata/data_access/data_access.py
```diff
@@ -180,21 +219,41 @@ def move_to(self, file_name, to_dir, change_name_to=None):
...
         """
         self._check_filename(file_name)
-        src = posixpath.join(server_setup.LOCAL_DIR, file_name)
+        src = os.path.join(server_setup.LOCAL_DIR, file_name)
         file_name = file_name if change_name_to is None else change_name_to
-        dest = posixpath.join(self.root, to_dir, file_name)
+        dest = os.path.join(self.root, to_dir, file_name)
         print(f"--> Moving file {src} to {dest}")
         self._check_file_exists(dest, should_exist=False)
-        self.copy(src, dest)
```
jon-hagg, line 188: Avoiding the use of `copy` because `dest` is a file, and we don't it to automatically create `/path/to/file.txt/` as a folder

powersimdata/data_access/data_access.py
rouille: You lost me here. Why a `tuple` is not required for "input" and "output"?

powersimdata/data_access/data_access.py
```diff
@@ -35,6 +47,34 @@ def move_to(self, file_name, to_dir, change_name_to=None):
...
+        :param bool backup: pass True if relative to backup root dir
+        :raises ValueError: if kind is invalid
+        :return: (*str*) -- the specified path
+        """
+        _allowed = list(_dirs.keys())
+        if kind not in _allowed:
+            raise ValueError(f"Invalid 'kind', must be one of {_allowed}")
+
+        root = self.root if not backup else self.backup_root
+        return self.join(root, *_dirs[kind])
```
rouille, line 63: Not sure I understand what this does with `self.join` being `None`

powersimdata/data_access/data_access.py
jon-hagg: They are changed to tuples already in `server_setup.py`

powersimdata/data_access/data_access.py
jon-hagg: We set `self.join` in the subclasses - assumption is we never work with base class instances, but we could probably enforce that

rouille: I don't know why I forgot about the child classes ... Sorry about that.

danielolsen: `raise NotImplementedError` in the parent?

powersimdata/data_access/data_access.py
```diff
@@ -180,21 +219,41 @@ def move_to(self, file_name, to_dir, change_name_to=None):
         :param str change_name_to: new name for file when copied to data store.
         """
         self._check_filename(file_name)
-        src = posixpath.join(server_setup.LOCAL_DIR, file_name)
+        src = os.path.join(server_setup.LOCAL_DIR, file_name)
```
rouille, line 222: we could use `self.join` here, no?

powersimdata/data_access/data_access.py
```diff
@@ -180,21 +219,41 @@ def move_to(self, file_name, to_dir, change_name_to=None):
...
         file_name = file_name if change_name_to is None else change_name_to
-        dest = posixpath.join(self.root, to_dir, file_name)
+        dest = os.path.join(self.root, to_dir, file_name)
```
rouille, line 224: here too I would say

powersimdata/data_access/data_access.py
jon-hagg: This method is already in the base class, so we would have to do something like set `self.join` to a function that raises the error. Currently I think we would get an `AttributeError` but figure it's ok since nothing would work anyway given a pure `DataAccess` instance.

powersimdata/data_access/data_access.py
```diff
@@ -180,21 +219,41 @@ def move_to(self, file_name, to_dir, change_name_to=None):
...
 
-        :param str relative_path: the path, without filename, relative to root
+        :param str src: path to original
+        :param str dest: destination path
+        :param bool recursive: create directories recursively
+        :param bool update: ignored
         """
-        target = os.path.join(self.root, relative_path)
-        os.makedirs(target, exist_ok=True)
+        self.makedir(dest)
```
rouille, line 251: It seems that `copytree` will create `dest`. Should we move this line in the `else` then?

powersimdata/data_access/data_access.py
rouille: It looks like that `LocalDataAccess` and `SSHDataAccess` don't have much in common anymore.

jon-hagg: True, mainly they share an interface, but not much implementation. That may be a good thing though - there is a bit more code but it's also more straightforward.

powersimdata/data_access/data_access.py
```diff
@@ -371,7 +408,7 @@ def move_to(self, file_name, to_dir=None, change_name_to=None):
 
         os.remove(from_path)
 
-    def execute_command(self, command):
+    def _execute_command(self, command):
         """Execute a command locally at the data access.
```
rouille, line 412: Execute a command on the server?

Also, do we need to have a function? I mean it used throughout the class but what is the gain of doing `self._execute_command(command)` and not  `self.ssh.exec_command(command)`? The method could then be removed from the parent class.

powersimdata/scenario/delete.py
```diff
@@ -41,25 +40,21 @@ def delete_scenario(self, confirm=True):
         self._scenario_list_manager.delete_entry(scenario_id)
         self._execute_list_manager.delete_entry(scenario_id)
 
-        wildcard = f"{scenario_id}_*"
-
         print("--> Deleting scenario input data on server")
```
rouille, line 43: I believe we can delete locally and hence just print "Deleting scenario input data"

powersimdata/scenario/delete.py
```diff
@@ -41,25 +40,21 @@ def delete_scenario(self, confirm=True):
...
-        target = posixpath.join(self.path_config.input_dir(), wildcard)
+        target = self._data_access.match_scenario_files(scenario_id, "input")
         self._data_access.remove(target, recursive=False, confirm=confirm)
 
         print("--> Deleting scenario output data on server")
```
rouille, line 47: And here "Deleting scenario output data"

powersimdata/scenario/delete.py
```diff
@@ -41,25 +40,21 @@ def delete_scenario(self, confirm=True):
...
         print("--> Deleting temporary folder on server")
-        tmp_dir = posixpath.join(
-            self.path_config.execute_dir(), f"scenario_{scenario_id}"
-        )
+        tmp_dir = self._data_access.match_scenario_files(scenario_id, "tmp")
         self._data_access.remove(tmp_dir, recursive=True, confirm=confirm)
 
         print("--> Deleting input and output data on local machine")
-        target = os.path.join(server_setup.LOCAL_DIR, "data", "**", wildcard)
+        target = os.path.join(server_setup.LOCAL_DIR, "data", "**", f"{scenario_id}_*")
```
rouille, line 57: I think, to be consistent with the above commands we can use the '`join` method:
```
target = self._data_access.join(server_setup.LOCAL_DIR, "data", "**", f"{scenario_id}_*")
```
We then can remove the `import os` and make the class fully OS independent.

rouille, line 57: Can we do a `LocalDataAccess().join` here and get rid of the `import os`. It will be nice to be fully OS independent in the class.

powersimdata/scenario/execute.py
```diff
@@ -331,22 +331,22 @@ def __init__(self, data_access, scenario_info, grid, ct):
...
-        self.REL_TMP_DIR = posixpath.join(
-            server_setup.EXECUTE_DIR, self.scenario_folder
+        self.REL_TMP_DIR = self._data_access.join(
+            server_setup.EXECUTE_DIR, f"scenario_{self.scenario_id}"
         )
+        self.TMP_DIR = self._data_access.match_scenario_files(self.scenario_id, "tmp")
 
     def create_folder(self):
         """Creates folder on server that will enclose simulation inputs."""
         print("--> Creating temporary folder on server for simulation inputs")
```
rouille, line 343: I guess the the temporary folder can be crated either on the local machine or the server. Perhaps we can the `description` attribute here too.

powersimdata/scenario/execute.py
```diff
@@ -371,20 +371,13 @@ def prepare_profile(self, kind, profile_as=None):
...
-            )
-            to_dir = posixpath.join(
-                self.server_config.execute_dir(), self.scenario_folder
-            )
-            _, _, stderr = self._data_access.copy(f"{from_dir}/{kind}.csv", to_dir)
-            if len(stderr.readlines()) != 0:
-                raise IOError(f"Failed to copy {kind}.csv on server")
+            from_dir = self._data_access.match_scenario_files(profile_as, "tmp")
+            to_dir = self.TMP_DIR
+            self._data_access.copy(f"{from_dir}/{kind}.csv", to_dir)
```
rouille, line 383: It is kind of confusing to `move to` using relative path but to `copy` with absolute path

jon-hagg: Agree, I just didn't know a cleaner way to do it. The reason for this is mainly how we use `copy` within `move.py`, where the destination (backup disk) is not relative to the `self.root`. Maybe we can revisit this in another PR.

rouille: Sounds good

rouille (approved):
Looks good

jon-hagg:
Rebased and fixed the test that was still calling `data_access._execute_command("whoami")`

PR #477, opened 2021-05-06
Support local+native usage

jon-hagg:
### Purpose
Add the remaining code changes so that the framework can be used locally, outside docker, on any os, using a local REISE.jl installation.

### What the code is doing
* Factor out `launch_scenario` into subclasses for each use case, and similarly for extracting output.
* Add `State.refresh` method so state changes have up to date information without re-initializing the instance
* Split up configuration into classes, making it easy to override values specific to a given environment
* various cleanup

### Testing
* Ran a simulation end to end on windows. Ran into issues due to using julia 1.6 but upgrading PyCall and CSV packages fixed the issues. 
* Ran simulation on MacOS using julia 1.5 (there are issues with 1.6)
* Ran simulation in plug to verify no regressions

### Usage Example/Visuals
The PR for documentation is https://github.com/Breakthrough-Energy/docs/pull/62. For testing locally, I changed the name of the data directory to `ScenarioDataLocal` (in both PowerSimData and REISE.jl) so it won't collide with the existing data from the server. 

### Time estimate
40 mins

### DIFF
--- a/powersimdata/data_access/context.py
+++ b/powersimdata/data_access/context.py
@@ -1,4 +1,5 @@
 from powersimdata.data_access.data_access import LocalDataAccess, SSHDataAccess
+from powersimdata.data_access.launcher import HttpLauncher, NativeLauncher, SSHLauncher
 from powersimdata.utility import server_setup
 from powersimdata.utility.config import DeploymentMode, get_deployment_mode
 
@@ -23,3 +24,16 @@ class Context:
         if mode == DeploymentMode.Server:
             return SSHDataAccess(root)
         return LocalDataAccess(root)
+
+    @staticmethod
+    def get_launcher(scenario):
+        """Return instance for interaction with simulation engine
+
+        :param powersimdata.Scenario scenario: a scenario object
+        """
+        mode = get_deployment_mode()
+        if mode == DeploymentMode.Server:
+            return SSHLauncher(scenario)
+        elif mode == DeploymentMode.Container:
+            return HttpLauncher(scenario)
+        return NativeLauncher(scenario)
--- a/powersimdata/data_access/context.py
+++ b/powersimdata/data_access/context.py
@@ -1,6 +1,6 @@
 from powersimdata.data_access.data_access import LocalDataAccess, SSHDataAccess
 from powersimdata.utility import server_setup
-from powersimdata.utility.server_setup import DeploymentMode, get_deployment_mode
+from powersimdata.utility.config import DeploymentMode, get_deployment_mode
 
 
 class Context:
--- a/powersimdata/data_access/context.py
+++ b/powersimdata/data_access/context.py
@@ -14,6 +14,8 @@ class Context:
 
         :param str data_loc: pass "disk" if using data from backup disk,
             otherwise leave the default.
+        :return: (:class:`powersimdata.data_access.data_access.DataAccess`) -- a data access
+            instance
         """
         if data_loc == "disk":
             root = server_setup.BACKUP_DATA_ROOT_DIR
@@ -29,7 +31,8 @@ class Context:
     def get_launcher(scenario):
         """Return instance for interaction with simulation engine
 
-        :param powersimdata.Scenario scenario: a scenario object
+        :param powersimdata.scenario.scenario.Scenario scenario: a scenario object
+        :return: (:class:`powersimdata.data_access.launcher.Launcher`) -- a launcher instance
         """
         mode = get_deployment_mode()
         if mode == DeploymentMode.Server:
--- a/powersimdata/data_access/launcher.py
+++ b/powersimdata/data_access/launcher.py
@@ -30,14 +30,16 @@ def _check_solver(solver):
         raise ValueError(f"Invalid solver: options are {solvers}")
 
 
-# TODO - check_progress (just print a message for SSHLauncher)
-# TODO - extract_data
 class Launcher:
     def __init__(self, scenario):
         self.scenario = scenario
 
     def _launch(self, threads=None, solver=None, extract_data=True):
         raise NotImplementedError
+
+    def extract_simulation_output(self):
+        """Extracts simulation outputs {PG, PF, LMP, CONGU, CONGL} on server."""
+        pass
 
     def launch_simulation(self, threads=None, solver=None, extract_data=True):
         _check_threads(threads)
@@ -88,12 +90,11 @@ class SSHLauncher(Launcher):
         :return: (*subprocess.Popen*) -- new process used to launch simulation.
         """
         extra_args = []
-
-        if threads:
+        if threads is not None:
             # Use the -t flag as defined in call.py in REISE.jl
             extra_args.append("--threads " + str(threads))
 
-        if solver:
+        if solver is not None:
             extra_args.append("--solver " + solver)
 
         if not isinstance(extract_data, bool):
@@ -102,6 +103,18 @@ class SSHLauncher(Launcher):
             extra_args.append("--extract-data")
 
         return self._run_script("call.py", extra_args=extra_args)
+
+    def extract_simulation_output(self):
+        """Extracts simulation outputs {PG, PF, LMP, CONGU, CONGL} on server.
+
+        :return: (*subprocess.Popen*) -- new process used to extract output
+            data.
+        """
+        print("--> Extracting output data on server")
+        return self._run_script("extract_data.py")
+
+    def check_progress(self):
+        print("Information is available on the server.")
 
 
 class HttpLauncher(Launcher):
@@ -124,6 +137,16 @@ class HttpLauncher(Launcher):
             )
         return resp
 
+    def check_progress(self):
+        """Get the status of an ongoing simulation, if possible
+
+        :return: (*dict*) -- json response
+        """
+        scenario_id = self.scenario.scenario_id
+        url = f"http://{server_setup.SERVER_ADDRESS}:5000/status/{scenario_id}"
+        resp = requests.get(url)
+        return resp.json()
+
 
 class NativeLauncher(Launcher):
     def _launch(self, threads=None, solver=None, extract_data=True):
@@ -134,6 +157,13 @@ class NativeLauncher(Launcher):
         :param str solver: the solver used for optimization. This defaults to
             None, which translates to gurobi
         :param bool extract_data: always True
+        :return: (*dict*) -- json response
+        """
+        pass
+
+    def check_progress(self):
+        """Get the status of an ongoing simulation, if possible
+
         :return: (*dict*) -- json response
         """
         pass
--- a/powersimdata/data_access/launcher.py
+++ b/powersimdata/data_access/launcher.py
@@ -45,7 +45,7 @@ class Launcher:
     def launch_simulation(self, threads=None, solver=None, extract_data=True):
         _check_threads(threads)
         _check_solver(solver)
-        self._launch(threads, solver, extract_data)
+        return self._launch(threads, solver, extract_data)
 
 
 class SSHLauncher(Launcher):
@@ -127,7 +127,8 @@ class HttpLauncher(Launcher):
         :param str solver: the solver used for optimization. This defaults to
             None, which translates to gurobi
         :param bool extract_data: always True
-        :return: (*requests.Response*) -- the http response object
+        :return: (*requests.Response*) -- http response from the engine, with a json
+            body as is returned by check_progress
         """
         scenario_id = self.scenario.scenario_id
         url = f"http://{server_setup.SERVER_ADDRESS}:5000/launch/{scenario_id}"
@@ -141,7 +142,8 @@ class HttpLauncher(Launcher):
     def check_progress(self):
         """Get the status of an ongoing simulation, if possible
 
-        :return: (*dict*) -- json response
+        :return: (*dict*) -- contains "output", "errors", "scenario_id", and "status"
+            keys which map to stdout, stderr, and the respective scenario attributes
         """
         scenario_id = self.scenario.scenario_id
         url = f"http://{server_setup.SERVER_ADDRESS}:5000/status/{scenario_id}"
@@ -158,7 +160,8 @@ class NativeLauncher(Launcher):
         :param str solver: the solver used for optimization. This defaults to
             None, which translates to gurobi
         :param bool extract_data: always True
-        :return: (*dict*) -- json response
+        :return: (*dict*) -- contains "output", "errors", "scenario_id", and "status"
+            keys which map to stdout, stderr, and the respective scenario attributes
         """
         sys.path.append(server_setup.ENGINE_DIR)
         from pyreisejl.utility import app
@@ -168,7 +171,8 @@ class NativeLauncher(Launcher):
     def check_progress(self):
         """Get the status of an ongoing simulation, if possible
 
-        :return: (*dict*) -- json response
+        :return: (*dict*) -- contains "output", "errors", "scenario_id", and "status"
+            keys which map to stdout, stderr, and the respective scenario attributes
         """
         sys.path.append(server_setup.ENGINE_DIR)
         from pyreisejl.utility import app
--- a/powersimdata/data_access/launcher.py
+++ b/powersimdata/data_access/launcher.py
@@ -32,6 +32,11 @@ def _check_solver(solver):
 
 
 class Launcher:
+    """Base class for interaction with simulation engine.
+
+    :param powrsimdata.scenario.scenario.Scenario scenario: scenario instance
+    """
+
     def __init__(self, scenario):
         self.scenario = scenario
 
--- a/powersimdata/data_access/launcher.py
+++ b/powersimdata/data_access/launcher.py
@@ -1,4 +1,5 @@
 import posixpath
+import sys
 
 import requests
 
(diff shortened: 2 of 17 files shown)


powersimdata/data_access/context.py
```diff
@@ -23,3 +24,16 @@ def get_data_access(data_loc=None):
         if mode == DeploymentMode.Server:
             return SSHDataAccess(root)
         return LocalDataAccess(root)
+
+    @staticmethod
+    def get_launcher(scenario):
+        """Return instance for interaction with simulation engine
+
+        :param powersimdata.Scenario scenario: a scenario object
```
rouille, line 32: Do you know if Sphinx will be able to link to `powersimdata.scenario.scenario.Scenario`?

rouille: Also, `:return:` is missing

powersimdata/utility/config.py
```diff
@@ -0,0 +1,59 @@
...
+    INPUT_DIR = ("data", "input")
+    OUTPUT_DIR = ("data", "output")
+    LOCAL_DIR = os.path.join(Path.home(), "ScenarioData", "")
+
+
+@dataclass(frozen=True)
+class ServerConfig(Config):
+    SERVER_ADDRESS = os.getenv("BE_SERVER_ADDRESS", "becompute01.gatesventures.com")
+    SERVER_SSH_PORT = os.getenv("BE_SERVER_SSH_PORT", 22)
+    BACKUP_DATA_ROOT_DIR = "/mnt/RE-Storage/v2"
```
rouille, line 24: Should we set `BACKUP_DATA_ROOT_DIR` to `None` or `not existing` or anything else that can be interpreted by a `Move` object as there is no backup disk?

powersimdata/data_access/launcher.py
```diff
@@ -0,0 +1,180 @@
...
+    solvers = ("gurobi", "glpk")
+    if solver is not None and solver.lower() not in solvers:
+        raise ValueError(f"Invalid solver: options are {solvers}")
+
+
+class Launcher:
+    def __init__(self, scenario):
+        self.scenario = scenario
+
+    def _launch(self, threads=None, solver=None, extract_data=True):
```
rouille, line 38: missing docstring

powersimdata/data_access/launcher.py
```diff
@@ -0,0 +1,180 @@
...
+        raise NotImplementedError
+
+    def extract_simulation_output(self):
+        """Extracts simulation outputs {PG, PF, LMP, CONGU, CONGL} on server."""
+        pass
+
+    def launch_simulation(self, threads=None, solver=None, extract_data=True):
```
rouille, line 45: missing docstring

powersimdata/data_access/launcher.py
```diff
@@ -0,0 +1,180 @@
...
+        extra_args = []
+        if threads is not None:
+            # Use the -t flag as defined in call.py in REISE.jl
+            extra_args.append("--threads " + str(threads))
+
+        if solver is not None:
+            extra_args.append("--solver " + solver)
+
+        if not isinstance(extract_data, bool):
+            raise TypeError("extract_data must be a boolean: 'True' or 'False'")
```
rouille, line 102: Do we need to explicitly mentioned `True` or `False` in the error message

ahurli (commented):
I like the new classes for the different ways to access the engine! I'm curious as your thoughts about how this design can support multiple engines in the future. It looks like we're only supporting `REISE.jl` in our local implementation (as opposed to including `REISE` like the ssh implementation does), which is probably fine for now, but as we add new engines (e.g. `switch`, `PowerSimulations.jl`), would we add new launchers (e.g. `NativeSwitchLauncher`, 'NativeREISEjlLauncher`) or do you think there's a better way of doing that?

powersimdata/data_access/context.py
jon-hagg: You're right, fixed. Also, the return needed a new syntax in order to generate a link to the class

jon-hagg:
> I like the new classes for the different ways to access the engine! I'm curious as your thoughts about how this design can support multiple engines in the future. It looks like we're only supporting `REISE.jl` in our local implementation (as opposed to including `REISE` like the ssh implementation does), which is probably fine for now, but as we add new engines (e.g. `switch`, `PowerSimulations.jl`), would we add new launchers (e.g. `NativeSwitchLauncher`, 'NativeREISEjlLauncher`) or do you think there's a better way of doing that?

Good point - I think we'll have to refactor a bit for that. Ideally if we have a standard interface to each engine, I'm hoping we can call them the same way, just with a different configuration. Maybe something like `NativeLauncher(engine='switch')` or equivalent. But the signatures you mentioned might also work - if we make it so that new classes like that can be defined almost entirely by inheritance, it could be just as simple to add. Thinking here of a class hierarchy like this:
```
class NativeLauncherBase:
class HttpLauncherBase:
class REISEjlBase:
# implementations here
```

and subclassing via

```
class NativeREISEjlLauncher(REISEjlBase, NativeLauncherBase):
    # inheritance does all the work
    pass
```

Not sure exactly if/how this would work, but just sharing potential design.

powersimdata/utility/config.py
```diff
@@ -0,0 +1,59 @@
...
+    Server = "SERVER"
+    Container = "CONTAINER"
+    Local = "LOCAL"
+
+    CONFIG_MAP = {Server: ServerConfig, Container: ContainerConfig, Local: LocalConfig}
+
+
+def get_deployment_mode():
+    mode = os.getenv("DEPLOYMENT_MODE")
+    if mode is None:
```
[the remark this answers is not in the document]
jon-hagg: Yeah it's unclear to me what the default should be in the future, since the docker roadmap and (corresponding user stories) is still being scoped out. But it's something to keep in mind.

powersimdata/data_access/launcher.py
```diff
@@ -0,0 +1,202 @@
...
+
+    :param str solver: the solver used for the optimization
+    :raises ValueError: if invalid solver provided
+    """
+    solvers = ("gurobi", "glpk")
+    if solver is not None and solver.lower() not in solvers:
+        raise ValueError(f"Invalid solver: options are {solvers}")
+
+
+class Launcher:
```
rouille, line 34: Missing docstring. And, usually, we add here the parameter (here `scenario`) that is needed to instantiate the class

powersimdata/scenario/create.py
```diff
@@ -176,6 +173,9 @@ def set_grid(self, grid_model="usa_tamu", interconnect="USA"):
         self._scenario_info["grid_model"] = self.builder.grid_model
         self._scenario_info["interconnect"] = self.builder.interconnect
 
+    def _leave(self):
```
rouille, line 176: missing docstring. You can use `"""Cleans when leaving state."""` as used in the `Analyze` classs

powersimdata/scenario/execute.py
```diff
@@ -29,16 +27,20 @@ class Execute(State):
...
 
     def __init__(self, scenario):
         """Constructor."""
-        self._scenario_info = scenario.info
-        self._scenario_status = scenario.status
         super().__init__(scenario)
+        self.refresh(scenario)
+
+    @property
+    def scenario_id(self):
```
rouille, line 40: missing docstring

powersimdata/scenario/execute.py
```diff
@@ -29,16 +27,20 @@ class Execute(State):
...
+        return self._scenario_info["id"]
 
+    def refresh(self, scenario):
```
rouille, line 43: missing docstring

powersimdata/scenario/state.py
```diff
@@ -1,7 +1,3 @@
-from powersimdata.data_access.execute_list import ExecuteListManager
-from powersimdata.data_access.scenario_list import ScenarioListManager
-
-
 class State(object):
     """Defines an interface for encapsulating the behavior associated with a
         particular state of the Scenario object.
```
rouille, line 3: Should be unindented

powersimdata/utility/config.py
```diff
@@ -0,0 +1,58 @@
...
+        return DeploymentMode.Server
+    if mode == "1" or mode.lower() == "container":
+        return DeploymentMode.Container
+    if mode == "2" or mode.lower() == "local":
+        return DeploymentMode.Local
+
+
+def get_config():
+    mode = get_deployment_mode()
+    return DeploymentMode.CONFIG_MAP[mode]()
```
rouille, line 58: I guess we need docstrings for all the classes and functions in this module

kasparm:
Running into the following error when trying to run a scenario:
```
--> Launching simulation on local
  File "test.py", line 65, in <module>
    process_run = scenario.launch_simulation()
  File "/Users/kmueller/EGM/temp/testPSDWindows/PowerSimData/powersimdata/scenario/execute.py", line 178, in launch_simulation
    return self._launcher.launch_simulation(threads, solver, extract_data)
  File "/Users/kmueller/EGM/temp/testPSDWindows/PowerSimData/powersimdata/data_access/launcher.py", line 75, in launch_simulation
    return self._launch(threads, solver, extract_data)
  File "/Users/kmueller/EGM/temp/testPSDWindows/PowerSimData/powersimdata/data_access/launcher.py", line 196, in _launch
    return app.launch_simulation(self.scenario.scenario_id, threads, solver)
TypeError: launch_simulation() takes 1 positional argument but 3 were given
```

rouille:
> Running into the following error when trying to run a scenario:
> 
> ```
> --> Launching simulation on local
> Traceback (most recent call last):
>   File "test.py", line 65, in <module>
>     process_run = scenario.launch_simulation()
>   File "/Users/kmueller/EGM/temp/testPSDWindows/PowerSimData/powersimdata/scenario/execute.py", line 178, in launch_simulation
>     return self._launcher.launch_simulation(threads, solver, extract_data)
>   File "/Users/kmueller/EGM/temp/testPSDWindows/PowerSimData/powersimdata/data_access/launcher.py", line 75, in launch_simulation
>     return self._launch(threads, solver, extract_data)
>   File "/Users/kmueller/EGM/temp/testPSDWindows/PowerSimData/powersimdata/data_access/launcher.py", line 196, in _launch
>     return app.launch_simulation(self.scenario.scenario_id, threads, solver)
> TypeError: launch_simulation() takes 1 positional argument but 3 were given
> ```

What REISE.jl did you check out?

kasparm:
Good point, not jon/windows.

rouille:
> Good point, not jon/windows.

This is one of this cross-package feature we love so much!

kasparm:
Anybody go this setup to run a scenario? For me it hangs with the status "running".

jon-hagg:
> Anybody go this setup to run a scenario? For me it hangs with the status "running".

That's farther than I got

rouille:
> Anybody go this setup to run a scenario? For me it hangs with the status "running".

What are you running?

kasparm:
@rouille https://breakthrough-energy.github.io/docs/powersimdata/index.html#creating-a-scenario

rouille:
> @rouille https://breakthrough-energy.github.io/docs/powersimdata/index.html#creating-a-scenario

You could try to run a Scenatio in ERCOT (see this [example](https://github.com/Breakthrough-Energy/plug/blob/main/standalone/demo_texas.py))

jon-hagg:
Currently I'm able to run the simulation using julia once it's prepared. Narrowed down the segfault to this [line](https://github.com/Breakthrough-Energy/REISE.jl/blob/develop/pyreisejl/utility/launchers.py#L117) when we try to load julia modules within python. Not sure if I'm the only one experiencing this.

powersimdata/data_access/launcher.py
```diff
@@ -0,0 +1,207 @@
...
+        :raises NotImplementedError: always - this must be implemented in a subclass
+        """
+        raise NotImplementedError
+
+    def extract_simulation_output(self):
+        """Extracts simulation outputs {PG, PF, LMP, CONGU, CONGL} on server."""
+        pass
+
+    def launch_simulation(self, threads=None, solver=None, extract_data=True):
+        """Launches simulation on target environment
```
BainanXia, line 61: Minor concern: "Launches" or "Launch"? It seems we are using both in such situations.

powersimdata/data_access/launcher.py
```diff
@@ -0,0 +1,207 @@
...
+
+        :param int/None threads: the number of threads to be used. This defaults to None,
```
BainanXia, line 63: We don't put `None` here, do we?

powersimdata/data_access/launcher.py
```diff
@@ -0,0 +1,207 @@
...
+            where None means auto.
+        :param str solver: the solver used for optimization. This defaults to
+            None, which translates to gurobi
+        :param bool extract_data: whether the results of the simulation engine should
+            automatically extracted after the simulation has run. This defaults to True.
+        :return: (*subprocess.Popen*) or (*requests.Response*) - either the
```
BainanXia, line 69: or (*dict*) ?

powersimdata/data_access/launcher.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+def _check_solver(solver):
+    """Validate solver argument
+
+    :param str solver: the solver used for the optimization
+    :raises ValueError: if invalid solver provided
+    """
+    solvers = ("gurobi", "glpk")
+    if solver is not None and solver.lower() not in solvers:
```
BainanXia, line 30: Should we also check `solver` is a str similarly as we did in `_check_threads`?

powersimdata/data_access/launcher.py
```diff
@@ -0,0 +1,207 @@
...
+        """
+        print("--> Extracting output data on server")
+        return self._run_script("extract_data.py")
+
+    def check_progress(self):
+        print("Information is available on the server.")
+
+
+class HttpLauncher(Launcher):
+    def _launch(self, threads=None, solver=None, extract_data=True):
```
BainanXia, line 149: Do we need `extract_data` here, given it is not passed to `request` call anyway.

powersimdata/data_access/launcher.py
```diff
@@ -0,0 +1,207 @@
...
+            keys which map to stdout, stderr, and the respective scenario attributes
+        """
+        scenario_id = self.scenario.scenario_id
+        url = f"http://{server_setup.SERVER_ADDRESS}:5000/status/{scenario_id}"
+        resp = requests.get(url)
+        return resp.json()
+
+
+class NativeLauncher(Launcher):
+    def _launch(self, threads=None, solver=None, extract_data=True):
```
BainanXia, line 182: Same comment for `extract_data`.

powersimdata/data_access/launcher.py
jon-hagg: Yeah each `_launch` method has to have the same signature, since we call it the same way (from `Launcher.launch_simulation`) without knowing which subclass it is. I think we might end up passing it to the request eventually though.

powersimdata/scenario/execute.py
```diff
@@ -187,110 +158,31 @@ def _check_if_ready(self):
...
-    def launch_simulation(self, threads=None, extract_data=True, solver=None):
-        """Launches simulation on target environment (server or container)
-
-        :param int/None threads: the number of threads to be used. This defaults to None,
-            where None means auto.
         :param bool extract_data: whether the results of the simulation engine should
             automatically extracted after the simulation has run. This defaults to True.
-        :param str solver: the solver used for optimization. This defaults to
-            None, which translates to gurobi
         :return: (*subprocess.Popen*) or (*requests.Response*) - either the
```
BainanXia, line 170: or (*dict*) ?

powersimdata/data_access/launcher.py
jon-hagg: Not intentionally, I will remove

powersimdata/scenario/execute.py
```diff
@@ -187,110 +158,31 @@ def _check_if_ready(self):
...
-            return self._launch_on_server(threads, solver, extract_data)
-        return self._launch_in_container(threads, solver)
+        return self._launcher.launch_simulation(threads, solver, extract_data)
 
     def check_progress(self):
-        """Get the lastest information from the server container
+        """Get the status of an ongoing simulation, if possible
 
-        :raises NotImplementedError: if not running in container mode
+        :return: (*dict*) -- progress information, or None
```
BainanXia, line 183: Maybe we should elaborate the returns for different launchers, SSH, HTTP, native?

powersimdata/data_access/launcher.py
jon-hagg: I'm not as worried about that here, since it will still fail due to not matching one of the predefined solvers or not having a `.lower()` method.

BainanXia: Yeah, I was thinking about the error message.

powersimdata/utility/config.py
```diff
@@ -0,0 +1,104 @@
...
+
+    :return: (*str*) -- the deployment mode
+    """
+    mode = os.getenv("DEPLOYMENT_MODE")
+    if mode is None:
+        return DeploymentMode.Server
+    if mode == "1" or mode.lower() == "container":
+        return DeploymentMode.Container
+    if mode == "2" or mode.lower() == "local":
+        return DeploymentMode.Local
```
BainanXia, line 95: Why we need two assignment patterns for `mode`: numbers and string, i.e. 1 == Container 2 == Local ?

jon-hagg: Hmm, I guess one reason is to decouple the values we set for the environment variable from the internal details. Not sure of a concrete example of when we need this, but feels like it's more flexible if the numbers don't imply any particular behavior/semantics.

powersimdata/data_access/launcher.py
jon-hagg: standardized on "Launch"

rouille:
What is the status of the tests?
* Does it successfully run on Windows? (@jon-hagg and @danielolsen)
* Does it successfully run on Mac? (@kasparm)
* Does it successfully run with **plug**? (@jon-hagg )
* Does it run with the client/server set up (@BainanXia and @danielolsen)?
By successfully, I mean creating, running and annlyzing a scenario (full integration test)

jon-hagg:
> What is the status of the tests?

I've done successful tests natively on windows/mac and using plug.

rouille (approved):
I like it like that

PR #480, opened 2021-05-14
chore: make psycopg2 optional

jon-hagg:
### Purpose
Simplify local installation by removing psycopg2 from requirements.txt and separate the code that depends on it, to prevent import errors if psycopg2 is not installed.

### What the code is doing
* Move psycopg2 to optional requirements
* Nothing is done with Pipfile, so tox will still run the associated sql tests. Anyone who needs to run those tests but isn't using pipenv will need to also install the optional requirements (or just `pip install psycopg2`). 
* Remove psycopg2 from setup.py - this will cause dependent packages not to install it as well, so we could regenerate the Pipfile.lock in PostREISE and PreREISE. 

### Testing
tox

### Time estimate
5 min


powersimdata/data_access/tests/test_data_access.py
```diff
@@ -67,7 +67,7 @@ def _check_content(filepath):
         assert CONTENT == f.read()
 
 
-root_dir = server_setup.DATA_ROOT_DIR
+root_dir = server_setup.DATA_ROOT_DIR.rstrip("/")
```
jon-hagg, line 70: This is a funny artifact of the `config.ini` file I'm using with my other branch - it ends up overriding the `DATA_ROOT_DIR` with `LOCAL_DIR`, which is expected, but the trailing slash breaks a couple of tests here which use string concatenation to join paths for simplicity/readability. Real functionality uses `os.path.join` so this isn't a problem elsewhere.

rouille: You lost me. In your other branch, `jon/windows` I presume, you have a `config.ini` file where you set the path for the various variables, including `DATA_ROOT_DIR`. But for what reason it is overwritten by `LOCAL_DIR`?

jon-hagg: The config.ini sets the deployment mode, which results in using a `LocalConfig`, where we default the `DATA_ROOT_DIR` to be equal to `LOCAL_DIR` since there is no need for separate locations. This caused a test failure since `LOCAL_DIR` has an extra `/` at the end. So this is just to make the tests work whether or not there is a config.ini causing this effect.

rouille (approved):
Looks good

PR #483, opened 2021-05-19
fix: path join for profiles_as

jon-hagg:
### Purpose
Fix a spot which still assumed posix style paths.

### What the code is doing
Use the join method on the data access instance which is context aware.

### Testing
We do the same thing elsewhere, so didn't manually check this

### Time estimate
3 min


danielolsen (approved):
Thanks for the quick fix!

rouille (approved):
🙏

Issue #482, opened 2021-05-19
Add function to create profiles without uploading to the server


danielolsen:
# :rocket:

- [ ] Is your feature request essential for your project?

### Describe the workflow you want to enable
I wish there was a way to write the profiles for a scenario to CSVs without the requirement to interact with the server. This would be useful for troubleshooting, as well as for passing profiles as inputs to other workflows (e.g. as inputs to capacity expansion). This would be the profile analog to what we did for being able to export a Grid to a case.mat file without touching the server (see #406).

### Describe your proposed implementation

There should be a function with a name like 'export_profiles' which would take a grid, profile versions (probably as a dict, with keys overlapping with the `scenario_info` so that the profile export could use this same dict or just the required keys), and a change table, and write each of the profiles to a CSV in a given folder, using the same TransformProfile-based workflow internally but without any file moves/uploads. The current `prepare_profile` function could be refactored to use this function, and then also perform the required moves/uploads.

rouille:
Should I take care of it? It seems this will needed once github/Breakthrough-Energy/SwitchWrapper#56 is merged

danielolsen:
> Should I take care of it? It seems this will needed once github/[Breakthrough-Energy/SwitchWrapper#56](https://github.com/Breakthrough-Energy/SwitchWrapper/pull/56) is merged

You are probably the most knowledgeable about the chain through `Execute -> TransformProfile -> InputData -> etc.`

PR #486, opened 2021-05-21
Add function to export profiles

jon-hagg:
### Purpose
Address #482 

### What the code is doing
Define `export_scaled_profile` function which can be called given a profile kind/version, grid, change table, and destination path. We now call this function from `SimulationInput.prepare_profile`, but it could be called independent of a scenario.

### Testing
Prepared a scenario in a container with a change table and it worked as expected.

### Time estimate
10 min

### DIFF
--- a/powersimdata/input/abstract_grid.py
+++ b/powersimdata/input/abstract_grid.py
@@ -1,7 +1,7 @@
 import pandas as pd
 
 
-class AbstractGrid(object):
+class AbstractGrid:
     """Grid Builder."""
 
     def __init__(self):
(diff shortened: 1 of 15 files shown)


BainanXia (approved):
Looks good. Thanks for the clean up, `object` is no longer needed with Python 3.0+

danielolsen (approved):
Clean implementation, thanks.

rouille:
Kind of wondering, if it would make sense to have an `export_data` module in `powersimdata.input` that will gather the function @jon-hagg just implemented and the `export_case_mat` function.

BainanXia:
> Kind of wondering, if it would make sense to have an `export_data` module in `powersimdata.input` that will gather the function @jon-hagg just implemented and the `export_case_mat` function.

That would definitely be more explicit from the user's perspective.

PR #491, opened 2021-05-28
Fix TypeError raised when using using default solver

jon-hagg:
### Purpose
If no solver is provided we should use the default which is determined in REISE.jl. I think I introduced this bug by adding the validation after having tested the code, and it was not re-tested (d'oh). 

### What the code is doing
Fix validation and add tests.

### Testing
New unit tests

### Time estimate
3 min

### DIFF
--- a/powersimdata/data_access/tests/test_launcher.py
+++ b/powersimdata/data_access/tests/test_launcher.py
@@ -1,0 +1,23 @@
+import pytest
+
+from powersimdata.data_access.launcher import _check_solver, _check_threads
+
+
+def test_check_solver():
+    _check_solver(None)
+    _check_solver("gurobi")
+    _check_solver("GLPK")
+    with pytest.raises(TypeError):
+        _check_solver(123)
+    with pytest.raises(ValueError):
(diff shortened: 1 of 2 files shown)


powersimdata/data_access/tests/test_launcher.py
```diff
@@ -0,0 +1,24 @@
...
+from powersimdata.data_access.launcher import _check_solver, _check_threads
+
+
+def test_check_solver():
+    _check_solver(None)
+    _check_solver("gurobi")
+    _check_solver("GLPK")
+    with pytest.raises(TypeError):
+        _check_solver(123)
+
```
BainanXia, line 12: Do we need an empty line here?

jon-hagg: Nope

BainanXia (approved):
I think this addresses the question brought up in our Slack community.

jon-hagg:
> I think this addresses the question brought up in our Slack community.

Yeah I was going to respond but seems it was deleted? Can just message him directly though.

PR #494, opened 2021-06-03
refactor: set grid_model property on grid

jon-hagg:
### Purpose
Add the property to grid and mock grid classes as it's commonly used and set at init time.

### What the code is doing
Simplify the grid constructor by adding `self.grid_model` as an instance variable - no need to try/catch to set `ModelImmutables`. Also set the property in `MockGrid` since it's already available. 

### Testing
Existing unit tests

### Time estimate
5 min

### DIFF
--- a/powersimdata/design/tests/test_scenario_info.py
+++ b/powersimdata/design/tests/test_scenario_info.py
@@ -128,7 +128,6 @@ class TestScenarioInfo(unittest.TestCase):
             wind=mock_wind,
             hydro=mock_hydro,
         )
-        scenario.state.grid.get_grid_model = lambda: "usa_tamu"
         scenario.state.grid.zone2id = {"Oregon": 202, "Arizona": 209}
         self.scenario_info = ScenarioInfo(scenario)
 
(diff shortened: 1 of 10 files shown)


powersimdata/design/transmission/upgrade.py
```diff
@@ -176,7 +176,7 @@ def get_branches_by_area(grid, area_names, method="either"):
 
     branch = grid.branch
     selected_branches = set()
-    grid_model = grid.get_grid_model()
+    grid_model = grid.grid_model
     for a in area_names:
         load_zone_names = area_to_loadzone(grid_model, a)
```
rouille, line 181: Can we use `grid.grid_model` here and remove ` grid_model = grid.grid_model` since there is the only place where it is used

rouille (approved):
Looks good

powersimdata/design/tests/test_scenario_info.py
```diff
@@ -79,36 +79,21 @@
...
+        if mock_plant["type"][i] == type
+    ]
+
+
+solar_plant_id = _select_plant_id("solar")
+wind_plant_id = _select_plant_id("wind")
+ng_plant_id = _select_plant_id("ng")
+coal_plant_id = _select_plant_id("coal")
+dfo_plant_id = _select_plant_id("dfo")
+hydro_plant_id = _select_plant_id("hydro")
```
rouille, line 96: nice

PR #495, opened 2021-06-04
feat: add Grid validation function

### LINKED ISSUE #481: Add validation function for Grid objects
# :rocket:

- [ ] Is your feature request essential for your project?
Not essential, but would serve to improve the workflow of multiple different projects.

### Describe the workflow you want to enable
I wish I could check whether a Grid object that is being loaded from outside of PowerSimData code (e.g. from a matfile, from a set of CSVs, or from another source) is internally consistent and contains all required information. By 'internally consistent', I mean:
- Every bus is connected to at least one branch or DC line.
- Every bus ID that is present in the branch, plant, or DC line tables should be present in the bus table.
- Every bus should be present in the `bus2sub` table.
- AC lines should only connect between buses of the same voltage, within the same interconnect.
- Transformers should only connect between buses within the same substation.
- Interpreting the AC branches as edges of a graph, there should be exactly one connected component per interconnection.
- etc.

### Describe your proposed implementation
There should be a new function that takes a Grid object as input, and raises a ValueError if any of the internal consistency checks fails.

### Additional context
This project would serve the validation of new grid models (so that we can tell that there is a problem before sending the grid and profiles to the simulation engine) as well as interpreting results from tools besides production cost models (e.g. a capacity expansion model run).

danielolsen:

### Purpose
Since we are increasingly looking to constructing Grid objects from datasets besides the `usa_tamu` CSVs, we are increasingly in need of a check that the Grid objects that we are constructing are internally consistent. This PR adds that check (closes #481), and also makes a couple of changes to some of the Western offshore wind buses we added, so that their voltages are consistent with their connected onshore voltages (found by running the tests).

### What the code is doing
There is one main user-facing `check_grid` function, which activates nine lower-level functions which check the bullet points from #481. Each of these is pretty self-explanatory, with the exception of the connected components check, for which we need to make a slight modification in the model immutables so that we know that the `"USA"` interconnect should really have three connected components. I don't see the `interconnect_combinations` parameter being used anywhere, so I think this will be okay, but maybe @rouille knows more.

### Testing
Tests pass on all of our Grids.

### Time estimate
15-30 minutes if you want to dig into all of the lower-level tests.

### DIFF
--- a/powersimdata/input/check.py
+++ b/powersimdata/input/check.py
@@ -1,0 +1,201 @@
+import networkx as nx
+
+from powersimdata.input.grid import Grid
+
+
+def check_grid(grid):
+    """Check whether an object is an internally-consistent Grid object.
+
+    :param powersimdata.input.grid.Grid grid: grid or grid-like object to check.
+    :raises TypeError: if ``grid`` is not a Grid object.
+    """
+    error_messages = []
+    if not isinstance(grid, Grid):
+        error_messages.append("grid must be a Grid object")
+    _check_attributes(grid, error_messages)
+    _check_for_islanded_buses(grid, error_messages)
+    _check_for_undescribed_buses(grid, error_messages)
+    _check_bus_against_bus2sub(grid, error_messages)
+    _check_ac_interconnects(grid, error_messages)
+    _check_transformer_substations(grid, error_messages)
+    _check_line_voltages(grid, error_messages)
+    _check_plant_against_gencost(grid, error_messages)
+    _check_connected_components(grid, error_messages)
+    if len(error_messages) > 0:
+        collected = "\n".join(error_messages)
+        raise ValueError(f"Problem(s) found with grid:\n{collected}")
+
+
+def _check_attributes(grid, error_messages):
+    """Check whether a Grid object has the required attributes.
+
+    :param powersimdata.input.grid.Grid grid: grid or grid-like object to check.
+    :param list error_messages: list, to be appended to with a str if:
+        ``grid`` is missing one or more required attributes.
+    """
+    required = {
+        "branch",
+        "bus",
+        "bus2sub",
+        "dcline",
+        "data_loc",
+        "gencost",
+        "grid_model",
+        "interconnect",
+        "model_immutables",
+        "plant",
+        "storage",
+        "sub",
+    }
+    for r in required:
+        if not hasattr(grid, r):
+            error_messages.append(f"grid object must have attribute {r}.")
+
+
+def _check_for_islanded_buses(grid, error_messages):
+    """Check whether a transmission network (AC & DC) does not connect to one or more
+    buses.
+
+    :param powersimdata.input.grid.Grid grid: grid or grid-like object to check.
+    :param list error_messages: list, to be appended to with a str if:
+        branches/DC lines exist in the ``grid``, but one or more buses are islanded.
+    """
+    if len(grid.branch) + len(grid.dcline) > 0:
+        connected_buses = set().union(
+            set(grid.branch.from_bus_id),
+            set(grid.branch.to_bus_id),
+            set(grid.dcline.from_bus_id),
+            set(grid.dcline.to_bus_id),
+        )
+        isolated_buses = set(grid.bus.index) - connected_buses
+        if len(isolated_buses) > 0:
+            error_messages.append(f"islanded buses detected: {isolated_buses}.")
+
+
+def _check_for_undescribed_buses(grid, error_messages):
+    """Check whether any transmission elements are connected to buses that are not
+    described in the bus table.
+
+    :param powersimdata.input.grid.Grid grid: grid or grid-like object to check.
+    :param list error_messages: list, to be appended to with a str if:
+        any transmission elements are connected to buses that are not described in the
+        bus table of the ``grid``.
+    """
+    expected_buses = set().union(
+        set(grid.branch.from_bus_id),
+        set(grid.branch.to_bus_id),
+        set(grid.dcline.from_bus_id),
+        set(grid.dcline.to_bus_id),
+    )
+    undescribed_buses = expected_buses - set(grid.bus.index)
+    if len(undescribed_buses) > 0:
+        error_messages.append(
+            "buses present in transmission network but missing from bus table: "
+            f"{undescribed_buses}."
+        )
+
+
+def _check_bus_against_bus2sub(grid, error_messages):
+    """Check whether indices of bus and bus2sub tables match.
+
+    :param powersimdata.input.grid.Grid grid: grid or grid-like object to check.
+    :param list error_messages: list, to be appended to with a str if:
+        indices of bus and bus2sub tables of the ``grid`` don't match.
+    """
+    if not set(grid.bus.index) == set(grid.bus2sub.index):
+        error_messages.append("indices for bus and bus2sub don't match.")
+
+
+def _check_ac_interconnects(grid, error_messages):
+    """Check whether any AC branches bridge across interconnections.
+
+    :param powersimdata.input.grid.Grid grid: grid or grid-like object to check.
+    :param list error_messages: list, to be appended to with a str if:
+        any AC branches bridge across interconnections of the ``grid``.
+    """
+    from_interconnect = grid.branch.from_bus_id.map(grid.bus.interconnect)
+    to_interconnect = grid.branch.to_bus_id.map(grid.bus.interconnect)
+    if not all(from_interconnect == to_interconnect):
+        non_matching_ids = grid.branch.index[from_interconnect != to_interconnect]
+        error_messages.append(
+            "branch(es) connected across multiple interconnections: "
+            f"{non_matching_ids}."
+        )
+
+
+def _check_transformer_substations(grid, error_messages):
+    """Check whether any transformers are are not within exactly one same substation.
+
+    :param powersimdata.input.grid.Grid grid: grid or grid-like object to check.
+    :param list error_messages: list, to be appended to with a str if:
+        any transformers in the ``grid`` are not within exactly one same substation.
+    """
+    txfmr_branch_types = {"Transformer", "TransformerWinding"}
+    branch = grid.branch
+    transformers = branch.loc[branch.branch_device_type.isin(txfmr_branch_types)]
+    from_sub = transformers.from_bus_id.map(grid.bus2sub.sub_id)
+    to_sub = transformers.to_bus_id.map(grid.bus2sub.sub_id)
+    if not all(from_sub == to_sub):
+        non_matching_transformers = transformers.index[from_sub != to_sub]
+        error_messages.append(
+            "transformer(s) connected across multiple substations: "
+            f"{non_matching_transformers}."
+        )
+
+
+def _check_line_voltages(grid, error_messages):
+    """Check whether any lines connect across different voltage levels.
+
+    :param powersimdata.input.grid.Grid grid: grid or grid-like object to check.
+    :param list error_messages: list, to be appended to with a str if:
+        any lines in the ``grid`` connect across different voltage levels.
+    """
+    lines = grid.branch.query("branch_device_type == 'Line'")
+    from_kV = lines.from_bus_id.map(grid.bus.baseKV)  # noqa: N806
+    to_kV = lines.to_bus_id.map(grid.bus.baseKV)  # noqa: N806
+    if not all(from_kV == to_kV):
+        non_matching_lines = lines.index[from_kV != to_kV]
+        error_messages.append(
+            f"line(s) connected across multiple voltages: {non_matching_lines}."
+        )
+
+
+def _check_plant_against_gencost(grid, error_messages):
+    """Check whether indices of plant and gencost tables match.
+
+    :param powersimdata.input.grid.Grid grid: grid or grid-like object to check.
+    :param list error_messages: list, to be appended to with a str if:
+        indices of plant and gencost tables of the ``grid`` don't match.
+    """
+    if not (
+        set(grid.plant.index)
+        == set(grid.gencost["before"].index)
+        == set(grid.gencost["after"].index)
+    ):
+        error_messages.append("indices for plant and gencost don't match.")
+
+
+def _check_connected_components(grid, error_messages):
+    """Check whether connected components and listed interconnects of a grid match.
+
+    :param powersimdata.input.grid.Grid grid: grid or grid-like object to check.
+    :param list error_messages: list, to be appended to with a str if:
+        connected components and listed interconnects of a ``grid`` don't match.
+    """
+    g = nx.from_pandas_edgelist(grid.branch, "from_bus_id", "to_bus_id")
+    num_connected_components = len([c for c in nx.connected_components(g)])
+    if len(grid.interconnect) == 1:
+        # Check for e.g. ['USA'] interconnect, which is really three interconnects
+        interconnect_aliases = grid.model_immutables.zones["interconnect_combinations"]
+        if grid.interconnect[0] in interconnect_aliases:
+            num_interconnects = len(interconnect_aliases[grid.interconnect[0]])
+        else:
+            num_interconnects = 1
+    else:
+        num_interconnects = len(grid.interconnect)
+    if num_interconnects != num_connected_components:
+        error_messages.append(
+            f"This grid contains {num_connected_components} connected components, "
+            f"but is specified as having {num_interconnects} interconnects: "
+            f"{grid.interconnect}."
+        )
--- a/powersimdata/input/tests/test_check.py
+++ b/powersimdata/input/tests/test_check.py
@@ -1,0 +1,27 @@
+from powersimdata import Grid
+from powersimdata.input.check import check_grid
+
+
+def test_check_eastern():
+    grid = Grid("Eastern")
+    check_grid(grid)
+
+
+def test_check_western():
+    grid = Grid("Western")
+    check_grid(grid)
+
+
+def test_check_texas():
+    grid = Grid("Texas")
+    check_grid(grid)
+
+
+def test_check_western_texas():
+    grid = Grid(["Western", "Texas"])
+    check_grid(grid)
+
+
+def test_check_usa():
+    grid = Grid(["USA"])
+    check_grid(grid)
--- a/powersimdata/network/usa_tamu/constants/zones.py
+++ b/powersimdata/network/usa_tamu/constants/zones.py
@@ -30,7 +30,10 @@ _exports = [
 mappings = {"loadzone", "state", "state_abbr", "interconnect"}
 
 # Define combinations of interconnects
-interconnect_combinations = {"USA", "Texas_Western"}
+interconnect_combinations = {
+    "USA": {"Eastern", "Western", "Texas"},
+    "Texas_Western": {"Western", "Texas"},
+}
 
 
 # Map state abbreviations to state name
--- a/Pipfile
+++ b/Pipfile
@@ -10,10 +10,11 @@ coverage = "*"
 pytest-cov = "*"
 
 [packages]
+networkx = "~=2.5"
 numpy = "~=1.20"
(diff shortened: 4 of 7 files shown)


jon-hagg:
We should probably add networkx to the setup.py as well

BainanXia:
Should we move the grid modification to a separate PR so that it is more trackable in the future, although these changes on bus voltages are very unlikely to cause any trouble given we created them.

powersimdata/input/check.py
```diff
@@ -0,0 +1,195 @@
...
+        num_interconnects = (
+            grid.model_immutables.zones["interconnect_combinations"][interconnect]
+            if interconnect in grid.model_immutables.zones["interconnect_combinations"]
+            else 1
+        )
+    else:
+        num_interconnects = len(grid.interconnect)
+    if num_interconnects != num_connected_components:
+        raise ValueError(
+            f"This grid constaints {num_connected_components} connected components, "
```
jon-hagg, line 192: constaints -> contains

powersimdata/input/check.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+def _check_for_islanded_buses(grid):
+    """Check whether a transmission network (AC & DC) does not connect to one or more
+    buses.
+
+    :param powersimdata.input.grid.Grid grid: grid or grid-like object to check.
+    :raises ValueError: if branches/DC lines exist, but one or more buses are islanded.
+    """
+    if len(grid.branch) + len(grid.dcline) > 0:
```
BainanXia, line 57: Not sure whether we need this check or not.

powersimdata/input/check.py
```diff
@@ -0,0 +1,195 @@
...
+        connected_buses = set().union(
+            set(grid.branch.from_bus_id),
+            set(grid.branch.to_bus_id),
+            set(grid.dcline.from_bus_id),
+            set(grid.dcline.to_bus_id),
+        )
+        isolated_buses = set(grid.bus.index) - connected_buses
+        if len(isolated_buses) > 0:
+            raise ValueError(f"Islanded buses detected: {isolated_buses}")
```
BainanXia, line 66: Maybe `Islanded` -> `islanded` to be consistent?

powersimdata/input/check.py
```diff
@@ -0,0 +1,195 @@
...
+
+def _check_plant_against_gencost(grid):
+    """Check whether indices of plant and gencost tables match.
+
+    :param powersimdata.input.grid.Grid grid: grid or grid-like object to check.
+    :raises ValueError: if indices of plant and gencost tables don't match.
+    """
+    if not (
+        set(grid.plant.index) == set(grid.gencost["before"].index)
+        and set(grid.plant.index) == set(grid.gencost["after"].index)
```
jon-hagg, line 166: I think we could chain the comparison here, e.g. `if not a == b == c: ...`

powersimdata/input/check.py
danielolsen: What if we have a single-bus Grid?

powersimdata/input/check.py
```diff
@@ -0,0 +1,195 @@
...
+    expected_buses = set().union(
+        set(grid.branch.from_bus_id),
+        set(grid.branch.to_bus_id),
+        set(grid.dcline.from_bus_id),
+        set(grid.dcline.to_bus_id),
+    )
+    undescribed_buses = expected_buses - set(grid.bus.index)
+    if len(undescribed_buses) > 0:
+        raise ValueError(
+            "Buses present in transmission network but missing from bus table: "
```
BainanXia, line 86: Same comment `Buses` -> `buses`, or the other way around.

danielolsen:
> Should we add networkx to the setup.py as well?

Good catch, yes

> Should we move the grid modification to a separate PR so that it is more trackable in the future, although these changes on bus voltages are very unlikely to cause any trouble given we created them.

If you think it is cleaner. It's already a distinct commit, and if we move it to a separate PR then we will need to merge that PR before this one, otherwise the tests in this one will fail.

powersimdata/network/usa_tamu/constants/zones.py
```diff
@@ -30,7 +30,7 @@
 mappings = {"loadzone", "state", "state_abbr", "interconnect"}
 
 # Define combinations of interconnects
-interconnect_combinations = {"USA", "Texas_Western"}
+interconnect_combinations = {"USA": 3, "Texas_Western": 2}
```
rouille, line 33: Would it be more useful to have a list of interconnects as values, i.e..:
```
interconnect_combinations = {"USA": ["Eastern", "Texas", "Western"], "Texas_Western": ["Texas", "Western"]}
```

danielolsen: I think that's probably cleaner to someone looking at just this file, yeah.

powersimdata/input/check.py
```diff
@@ -0,0 +1,195 @@
...
+            raise ValueError(f"Islanded buses detected: {isolated_buses}")
+
+
+def _check_for_undescribed_buses(grid):
+    """Check whether any transmission elements are connected to buses that are not
+    described in the bus table.
+
+    :param powersimdata.input.grid.Grid grid: grid or grid-like object to check.
+    :raises ValueError: if any transmission elements are connected to buses that are not
+    described in the bus table.
```
rouille, line 75: should be indented

powersimdata/input/check.py
```diff
@@ -0,0 +1,195 @@
...
+    bus = grid.bus
+    branch["from_interconnect"] = bus.loc[branch.from_bus_id, "interconnect"].tolist()
+    branch["to_interconnect"] = bus.loc[branch.to_bus_id, "interconnect"].tolist()
+    interconnect_match = branch["from_interconnect"] == branch["to_interconnect"]
+    if not interconnect_match.all():
+        non_matching_branches = interconnect_match.index[~interconnect_match]
+        raise ValueError(
+            "Branch(es) connected across multiple interconnections: "
+            f"{non_matching_branches}."
+        )
```
BainanXia, line 117: Maybe we could do following here:
```python
all(grid.branch.from_bus_id.map(grid.bus.interconnect) == grid.branch.to_bus_id.map(grid.bus.interconnect))
```

powersimdata/input/check.py
```diff
@@ -0,0 +1,195 @@
...
+    :raises ValueError: if any lines connect across difference voltage levels.
+    """
+    lines = grid.branch.query("branch_device_type == 'Line'").copy()
+    lines["from_kV"] = grid.bus.loc[lines.from_bus_id, "baseKV"].tolist()
+    lines["to_kV"] = grid.bus.loc[lines.to_bus_id, "baseKV"].tolist()
+    line_voltage_match = lines["from_kV"] == lines["to_kV"]
+    if not line_voltage_match.all():
+        non_matching_voltages = line_voltage_match.index[~line_voltage_match]
+        raise ValueError(
+            f"Lines(s) connected across multiple voltages: {non_matching_voltages}."
```
jon-hagg, line 154: Lines(s) -> Line(s)

powersimdata/input/check.py
```diff
@@ -0,0 +1,195 @@
...
+    :param powersimdata.input.grid.Grid grid: grid or grid-like object to check.
+    :raises ValueError: if any transformers are not within exactly one same substation.
+    """
+    txfmr_branch_types = {"Transformer", "TransformerWinding"}
+    branch = grid.branch
+    bus2sub = grid.bus2sub
+    transformers = branch.loc[branch.branch_device_type.isin(txfmr_branch_types)].copy()
+    transformers["from_sub"] = bus2sub.loc[transformers.from_bus_id, "sub_id"].tolist()
+    transformers["to_sub"] = bus2sub.loc[transformers.to_bus_id, "sub_id"].tolist()
+    transformer_sub_match = transformers["from_sub"] == transformers["to_sub"]
```
BainanXia, line 132: Same here
```python
all(transformers.from_sub.map(bus2sub.sub_id) == transformers.to_sub.map(bus2sub.sub_id))
```

powersimdata/input/check.py
BainanXia: Aha, then it is an island, isn't it? Yeah, you are right. We should keep this given we don't want this check to fail in that case.

powersimdata/input/check.py
```diff
@@ -0,0 +1,195 @@
...
+def _check_line_voltages(grid):
+    """Check whether any lines connect across difference voltage levels.
+
+    :param powersimdata.input.grid.Grid grid: grid or grid-like object to check.
+    :raises ValueError: if any lines connect across difference voltage levels.
+    """
+    lines = grid.branch.query("branch_device_type == 'Line'").copy()
+    lines["from_kV"] = grid.bus.loc[lines.from_bus_id, "baseKV"].tolist()
+    lines["to_kV"] = grid.bus.loc[lines.to_bus_id, "baseKV"].tolist()
+    line_voltage_match = lines["from_kV"] == lines["to_kV"]
```
BainanXia, line 150: Maybe same trick as above.

powersimdata/input/check.py
BainanXia: Maybe `line(s)` instead.

powersimdata/input/check.py
```diff
@@ -0,0 +1,195 @@
...
+    g = nx.from_pandas_edgelist(grid.branch, "from_bus_id", "to_bus_id")
+    num_connected_components = len([c for c in nx.connected_components(g)])
+    if len(grid.interconnect) == 1:
+        # Check for e.g. ['USA'] interconnect, which is really three interconnects
+        interconnect = grid.interconnect[0]
+        num_interconnects = (
+            grid.model_immutables.zones["interconnect_combinations"][interconnect]
+            if interconnect in grid.model_immutables.zones["interconnect_combinations"]
+            else 1
+        )
```
BainanXia, line 187: This should work equivalently:
```python
num_interconnects = grid.model_immutables.zones["interconnect_combinations"].get(interconnect, 1)
```

powersimdata/input/check.py
```diff
@@ -0,0 +1,195 @@
...
+def _check_connected_components(grid):
+    """Check whether connected components and listed interconnects of a grid match.
+
+    :param powersimdata.input.grid.Grid grid: grid or grid-like object to check.
+    :raises ValueError: if connected components and listed interconnects of a grid
+        don't match.
+    """
+    g = nx.from_pandas_edgelist(grid.branch, "from_bus_id", "to_bus_id")
+    num_connected_components = len([c for c in nx.connected_components(g)])
+    if len(grid.interconnect) == 1:
```
BainanXia, line 180: Maybe I need to refresh my brain here, the only situation we have `len(grid.interconnect)` > 1 is TexasWestern, right? If TexasWestern is handled in `else` block, what does `Texas_Western` entry do in `interconnect_combinations`?

powersimdata/input/tests/test_check.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def test_check_western_texas():
+    grid = Grid(["Western", "Texas"])
+    check_grid(grid)
+
+
+def test_check_usa():
+    grid = Grid(["USA"])
+    check_grid(grid)
```
BainanXia, line 27: This is neat! I was thinking of using mocks and realized there is too much trouble on that track.

powersimdata/input/check.py
danielolsen: Good question. I don't know if/when/how `interconnect_combinations` was used prior to this PR. We can init a `Grid(["Western", "Eastern"])` and then `grid.interconnect == ["Western", "Eastern"]`.

BainanXia:
> > Should we add networkx to the setup.py as well?
> 
> Good catch, yes
> 
> > Should we move the grid modification to a separate PR so that it is more trackable in the future, although these changes on bus voltages are very unlikely to cause any trouble given we created them.
> 
> If you think it is cleaner. It's already a distinct commit, and if we move it to a separate PR then we will need to merge that PR before this one, otherwise the tests in this one will fail.

Either way. I don't have strong preference. What do you think @rouille

powersimdata/input/check.py
danielolsen: Not if we change `grid.model_immutables.zones["interconnect_combinations"]` to produce the set of interconnections, rather than the number.

BainanXia:
Given this feature could be useful in many situations especially when we are adopting a new model, do you think it is worthwhile to make all the check run regardless of whether the previous one fails or not, so that the user will be aware of all the problems in one shot?

powersimdata/input/check.py
BainanXia: Then we could do 
```python
num_interconnects = len(grid.model_immutables.zones["interconnect_combinations"].get(interconnect, [dummy]))
```
instead?

danielolsen:
> Given this feature could be useful in many situations especially when we are adopting a new model, do you think it is worthwhile to make all the check run regardless of whether the previous one fails or not, so that the user will be aware of all the problems in one shot?

In this case, would we need to refactor `check_grid` so that it collects the exception from each test, then raises one mega-exception at the end?

powersimdata/input/check.py
danielolsen: We could, but I don't like using `[dummy]` here.

powersimdata/input/check.py
danielolsen: Done.

powersimdata/input/check.py
danielolsen: Done.

powersimdata/input/check.py
danielolsen: Done.

powersimdata/input/check.py
danielolsen: Done.

powersimdata/network/usa_tamu/constants/zones.py
danielolsen: Done.

powersimdata/input/check.py
danielolsen: Done.

powersimdata/input/check.py
danielolsen: Done.

BainanXia:
> > Given this feature could be useful in many situations especially when we are adopting a new model, do you think it is worthwhile to make all the check run regardless of whether the previous one fails or not, so that the user will be aware of all the problems in one shot?
> 
> In this case, would we need to refactor `check_grid` so that it collects the exception from each test, then raises one mega-exception at the end?

That's what I could think of. Maybe there are better solutions @jon-hagg @rouille ?

powersimdata/input/check.py
danielolsen: Done.

powersimdata/input/check.py
danielolsen: Done.

powersimdata/input/check.py
danielolsen: Done.

jon-hagg:
> > > Given this feature could be useful in many situations especially when we are adopting a new model, do you think it is worthwhile to make all the check run regardless of whether the previous one fails or not, so that the user will be aware of all the problems in one shot?
> > 
> > 
> > In this case, would we need to refactor `check_grid` so that it collects the exception from each test, then raises one mega-exception at the end?
> 
> That's what I could think of. Maybe there are better solutions @jon-hagg @rouille ?

What I'd do is have each function return a list of error messages and collect those, then raise an exception from the top level `_check_grid` if there are any errors. Or maybe just print them - no need to raise an error unless it's intended to be used to interrupt control flow

powersimdata/input/check.py
jon-hagg: Couple more letters :)

danielolsen: Done (for real this time, I think).

danielolsen:
> > > > Given this feature could be useful in many situations especially when we are adopting a new model, do you think it is worthwhile to make all the check run regardless of whether the previous one fails or not, so that the user will be aware of all the problems in one shot?
> > > 
> > > 
> > > In this case, would we need to refactor `check_grid` so that it collects the exception from each test, then raises one mega-exception at the end?
> > 
> > 
> > That's what I could think of. Maybe there are better solutions @jon-hagg @rouille ?
> 
> What I'd do is have each function return a list of error messages and collect those, then raise an exception from the top level `_check_grid` if there are any errors. Or maybe just print them - no need to raise an error unless it's intended to be used to interrupt control flow

Good call, we don't need to raise exceptions in the lower-level functions, they can return either a string or None and then we can combine any returned strings as necessary for generating the final exception. I think the high-level function should interrupt control flow here if there is a problem: grid integrity is important.

danielolsen:
Mega-Exception logic is added. We get:
```
ValueError: Problem(s) found with grid: islanded buses detected: {2090023}. indices for bus and bus2sub don't match.
```
or
```
ValueError: Problem(s) found with grid: buses present in transmission network but missing from bus table: {2090023}. indices for bus and bus2sub don't match. branch(es) connected across multiple interconnections: Int64Index([104191], dtype='int64', name='branch_id'). line(s) connected across multiple voltages: Int64Index([104191], dtype='int64', name='branch_id').
```
etc when we muck with the usa_tamu data, otherwise all tests pass.

powersimdata/input/check.py
```diff
@@ -0,0 +1,200 @@
...
+    _check_for_islanded_buses(grid, error_messages)
+    _check_for_undescribed_buses(grid, error_messages)
+    _check_bus_against_bus2sub(grid, error_messages)
+    _check_ac_interconnects(grid, error_messages)
+    _check_transformer_substations(grid, error_messages)
+    _check_line_voltages(grid, error_messages)
+    _check_plant_against_gencost(grid, error_messages)
+    _check_connected_components(grid, error_messages)
+    if len(error_messages) > 0:
+        raise ValueError(f"Problem(s) found with grid: {' '.join(error_messages)}")
```
BainanXia, line 25: This is nice! Would it be better to use `'\n'.join(error_messages)` instead?

powersimdata/input/check.py
```diff
@@ -0,0 +1,189 @@
...
+    if not all(from_sub == to_sub):
+        non_matching_transformers = transformers.index[from_sub != to_sub]
+        raise ValueError(
+            "transformer(s) connected across multiple substations: "
+            f"{non_matching_transformers}."
+        )
+
+
+def _check_line_voltages(grid):
+    """Check whether any lines connect across difference voltage levels.
```
ahurli, line 146: difference -> different?

powersimdata/input/check.py
danielolsen: Is there a convention for long error messages, whether they should be line wrapped or not? I am agnostic.

powersimdata/input/check.py
```diff
@@ -0,0 +1,189 @@
...
+
+    :param powersimdata.input.grid.Grid grid: grid or grid-like object to check.
+    :raises ValueError: if any lines connect across difference voltage levels.
```
ahurli: Same as above: difference -> different?

powersimdata/input/check.py
BainanXia: Good call. I don't know either. If there is a convention, then we should follow. I'm simply thinking of reading convenience.

danielolsen: `\n` won't work in an f-string

powersimdata/input/check.py
danielolsen: Done.

powersimdata/input/check.py
danielolsen: Done.

powersimdata/input/check.py
BainanXia: I see. A work around could be 
image

danielolsen: Done.

BainanXia (approved):
Very nice. This could be very useful. Thanks for the patience and being responsive.

danielolsen:
One thing that was not done here, that probably should be: checking for null values in all of the data frames.

PR #496, opened 2021-06-07
More error handling for grid validation

jon-hagg:
### Purpose
I assume we would want not proceed with the consistency checks unless it's a `Grid` object. Thought I commented on the original PR but evidently it wasn't submitted since only I can see it (something about "pending").  

### What the code is doing
Update the docstring and raise separate error for type check. Also added try/except around each check.

### Testing
tox

### Time estimate
5 min

### DIFF
--- a/powersimdata/input/check.py
+++ b/powersimdata/input/check.py
@@ -1,3 +1,5 @@
+import sys
+
 import networkx as nx
 
 from powersimdata.input.grid import Grid
@@ -13,15 +15,23 @@ def check_grid(grid):
     if not isinstance(grid, Grid):
         raise TypeError("grid must be a Grid object")
     error_messages = []
-    _check_attributes(grid, error_messages)
-    _check_for_islanded_buses(grid, error_messages)
-    _check_for_undescribed_buses(grid, error_messages)
-    _check_bus_against_bus2sub(grid, error_messages)
-    _check_ac_interconnects(grid, error_messages)
-    _check_transformer_substations(grid, error_messages)
-    _check_line_voltages(grid, error_messages)
-    _check_plant_against_gencost(grid, error_messages)
-    _check_connected_components(grid, error_messages)
+    for check in [
+        _check_attributes,
+        _check_for_islanded_buses,
+        _check_for_undescribed_buses,
+        _check_bus_against_bus2sub,
+        _check_ac_interconnects,
+        _check_transformer_substations,
+        _check_line_voltages,
+        _check_plant_against_gencost,
+        _check_connected_components,
+    ]:
+        try:
+            check(grid, error_messages)
+        except Exception:
+            error_messages.append(
+                f"Exception during {check.__name__}: {sys.exc_info()}"
+            )
     if len(error_messages) > 0:
         collected = "\n".join(error_messages)
         raise ValueError(f"Problem(s) found with grid:\n{collected}")
--- a/powersimdata/input/check.py
+++ b/powersimdata/input/check.py
@@ -30,7 +30,7 @@ def check_grid(grid):
             check(grid, error_messages)
         except Exception:
             error_messages.append(
-                f"Exception during {check.__name__}: {sys.exc_info()}"
+                f"Exception during {check.__name__}: {sys.exc_info()[1]!r}"
             )
     if len(error_messages) > 0:
         collected = "\n".join(error_messages)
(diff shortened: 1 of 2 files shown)


danielolsen:
Are there more 'first-order' checks that we would want to check to raise an exception early, rather than potentially causing an unintentional exception because we proceeded anyway with `error_messages` collection? E.g. we check that we have all of the required attributes, but we don't check that what's in them are dataframes (or dictionaries of dataframes), so we could easily run into an AttributeError if one of those isn't the right type.

My original design in #495 was to raise things ASAP, but we decided to collect as many errors as possible to report to the user.

Would a big `try`/`finally` block in `check_grid`--where `finally` prints as many error messages as it can--balance these interests?

jon-hagg:
Good point, I like the idea of wrapping the whole thing in a try block, or perhaps looping over the checks and wrapping each of those in a try/except.

danielolsen:
If we loop over all of them, and we hit one where there's an unexpected exception, what do we append to the `error_messages` list for the `finally` display? `"Encountered unexpected error in func_check_foo, possibly due to upstream issue"`?

jon-hagg:
> If we loop over all of them, and we hit one where there's an unexpected exception, what do we append to the `error_messages` list for the `finally` display? `"Encountered unexpected error in func_check_foo, possibly due to upstream issue"`?

That's what I was thinking

BainanXia:
> If we loop over all of them, and we hit one where there's an unexpected exception, what do we append to the `error_messages` list for the `finally` display? `"Encountered unexpected error in func_check_foo, possibly due to upstream issue"`?

There should be a way to get system error info during the exception. We could append something like `sys.exc_info()` instead of "unexpected error".

danielolsen:
> > If we loop over all of them, and we hit one where there's an unexpected exception, what do we append to the `error_messages` list for the `finally` display? `"Encountered unexpected error in func_check_foo, possibly due to upstream issue"`?
> 
> There should be a way to get system error info during the exception. We could append something like `sys.exc_info()` instead of "unexpected error".

We can do something like `except Exception as e: error_messages.append([something encompassing e.__str__])`. This exception would probably not be meaningful except to the advanced user, so maybe we combine the proposed messages about 'something upstream' with the lower-level exception information, then the message is as useful as possible no matter the user.

jon-hagg:
This is what it looks like with `sys.exc_info()` after running `del grid.dcline`

```
In [11]: check_grid(grid)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-11-38b1bc919390> in <module>
----> 1 check_grid(grid)

~/code/PowerSimData/powersimdata/input/check.py in check_grid(grid)
     35     if len(error_messages) > 0:
     36         collected = "\n".join(error_messages)
---> 37         raise ValueError(f"Problem(s) found with grid:\n{collected}")
     38
     39

ValueError: Problem(s) found with grid:
grid object must have attribute dcline.
Exception during _check_for_islanded_buses: (<class 'AttributeError'>, AttributeError("'Grid' object has no attribute 'dcline'"), <traceback object at 0x130ca1e40>)
Exception during _check_for_undescribed_buses: (<class 'AttributeError'>, AttributeError("'Grid' object has no attribute 'dcline'"), <traceback object at 0x130ca17c0>
```

powersimdata/input/check.py
```diff
@@ -8,19 +10,28 @@ def check_grid(grid):
...
+        _check_transformer_substations,
+        _check_line_voltages,
+        _check_plant_against_gencost,
+        _check_connected_components,
+    ]:
+        try:
+            check(grid, error_messages)
+        except Exception:
+            error_messages.append(
+                f"Exception during {check.__name__}: {sys.exc_info()[1]!r}"
```
danielolsen, line 33: `!r`! I hadn't seen that before.

powersimdata/input/tests/test_check.py
```diff
@@ -1,27 +1,19 @@
...
-    grid = Grid(["Western", "Texas"])
-    check_grid(grid)
+    del grid.dcline
+    with pytest.raises(ValueError):
+        check_grid(grid)
 
 
-def test_check_usa():
-    grid = Grid(["USA"])
+@pytest.mark.parametrize(
```
danielolsen, line 14: Fancy!

danielolsen (approved):
Very nice, thanks for following up on this.

BainanXia (approved):
Very nice.

powersimdata/input/check.py
jon-hagg: Yeah I haven't had many chances to use it, but came in handy this time since just doing `sys.exc_info()[1]` only prints the message, but no exception type.

PR #497, opened 2021-06-08
Daniel/column constants

danielolsen:

### Purpose
Consolidate column names/types in one place, to reduce duplicate information and potential headaches when making changes.

### What the code is doing
No code. Only constants being moved around.

### Testing
All tests still pass.

### Time estimate
5 minutes.


BainanXia:
In this case, we don't need `column_name_provider` or `column_type_provider` any more in `scenario_grid.py`, do we?

danielolsen:
> 
> 
> In this case, we don't need `column_name_provider` or `column_type_provider` any more in `scenario_grid.py`, do we?

There are still some uses of these that would need to be refactored:
https://github.com/Breakthrough-Energy/PowerSimData/blob/b6f15a7d7354bf065ccad503af85e125ec87d968/powersimdata/input/scenario_grid.py#L169

BainanXia (approved):
Make sense.

jon-hagg (approved):

PR #504, opened 2021-06-16
Josephine/calculate interzone capacity

ywang303:

### Purpose
The method calculates the aggregate transmission capacity (from + to) at zone level for each zone id. It is implemented in a similar way as "calculate_substation_capacity" method, but mapping for zone level.

### What the code is doing

- assign corresponding zone_id to the ends of each branch according to 'from_bus_id' and 'to'_bus_id'

- filter branches: exclude those branches from and to the same zone

- calculate the marginal transmission capacities by each unique 'from_zone_id" and by each unique 'to_zone_id' from filtered branches 

- calculate total transmission capacity for each zone by adding up 'from capacity' and 'to capacity' by each zone id

### Value -> Values
grid->total transmission capacity by each zone for the grid

### Testing
from powersimdata.design.transmission.substations import calculate_zone_capacity
from powersimdata.input.grid import Grid
grid = Grid("Western")
calculate_interzone_capacity(grid)

Screen Shot 2021-06-16 at 12 16 19 PM

### DIFF
--- a/powersimdata/design/transmission/zones.py
+++ b/powersimdata/design/transmission/zones.py
@@ -1,0 +1,19 @@
+def calculate_interzone_capacity(grid):
+    """For each zone in a grid, calculate the aggreagte zone transmission
+    capacity (in a transport model, ignoring power flow).
+
+    :param powersimdata.input.grid.Grid grid: a grid instance.
+    :return: (*pandas.Series*) -- index is zone IDs, values are total transmission
+        capacity (MW).
+    """
+    # Get new branch data frame with 'from_zone' and 'to_zone' columns
+    branch = grid.branch.assign(
+        from_zone_id=grid.branch.from_bus_id.map(grid.bus["zone_id"]),
+        to_zone_id=grid.branch.to_bus_id.map(grid.bus["zone_id"]),
+    )
+    # Calculate total substation capacity for matching 'from_zone' branches
+    filtered_branch = branch.query("from_zone_id != to_zone_id")
+    from_cap = filtered_branch.groupby("from_zone_id").sum()["rateA"]
+    to_cap = filtered_branch.groupby("to_zone_id").sum()["rateA"]
+    total_capacities = from_cap.combine(to_cap, lambda x, y: x + y, fill_value=0)
+    return total_capacities


rouille:
I am wondering if the name of the function clearly communicate what is being done. I feel like discarding intra-zone branches should be reflected somehow in the name of the function. What about `calculate_interzone_capacity`?

Josephine_commit/calculate_zone_capacity.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python
```
jon-hagg, line 1: I believe this file was added to git unintentionally since it's a duplicate of the other one

Josephine_commit/calculate_zone_capacity.py
```diff
@@ -0,0 +1,25 @@
...
+# coding: utf-8
+
+# In[ ]:
+
+
+def calculate_zone_capacity(grid):
+    """For each zone in a grid, calculate the aggreagte zone transmission
+    capacity (in a transport model, ignoring power flow).
```
rouille, line 9: A blank line should be left here

rouille, line 9: A blank line should be left here

powersimdata/design/transmission/zones.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# In[ ]:
+
+
+def calculate_zone_capacity(grid):
+    """For each zone in a grid, calculate the aggreagte zone transmission
+    capacity (in a transport model, ignoring power flow).
```
rouille, line 9: A blank line should be left here

powersimdata/design/transmission/zones.py
```diff
@@ -0,0 +1,25 @@
...
+    :param powersimdata.input.grid.Grid grid: a grid instance.
+    :return: (*pandas.Series*) -- index is zone IDs, value are total transmission
```
BainanXia, line 11: value -> values

Josephine_commit/calculate_zone_capacity.py
rouille: you can use `git rm` to remove it.

powersimdata/design/transmission/zones.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
ywang303: > I am wondering if the name of the function clearly communicate what is being done. I feel like discarding intra-zone branches should be reflected somehow in the name of the function. What about `calculate_interzone_capacity`?

How about delete the branch filtering line?

powersimdata/design/transmission/zones.py
ywang303: While add this explanation!

rouille:
We have a git guide available [here](https://breakthrough-energy.github.io/docs/user/git_guide.html) where you will find our  commit message convention. You can reword your commit message to follow our convention with an interactive rebase.

powersimdata/design/transmission/zones.py
ywang303: <br class="Apple-interchange-newline">I am wondering if the name of the function clearly communicate what is being done. I feel like discarding intra-zone branches should be reflected somehow in the name of the function. What about&nbsp;calculate_interzone_capacity?  

How about just deleting the branch filtering line?

rouille: It depends what you want to calculate. I guess we have three options:
* sum the `rateA` of all the branches within a zone, i.e, excluding branches that cross zone borders
* sum the `rateA` of all the branches that have one end in the zone
* sum the `rateA` in a zone as you do, i.e. by considering branches that only cross borders

If we want all the above options, we can have thre functions or one function with keyword arguments.

powersimdata/design/transmission/zones.py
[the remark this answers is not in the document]
ywang303: I guess your second option shall be sum the rateA of all the branches that have at least one end in the zone? Otherwise, it is just be the third, which is my calculation. Please correct me if I misunderstand.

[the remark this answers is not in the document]
ywang303: I will change the name of method to calculate_interzone_capacity

ywang303:
Finished all the script revision, would you please recheck to see if I shall make any modifications @BainanXia @jon-hagg @rouille

BainanXia (approved):
Clean.

PR #506, opened 2021-06-22
feat: add base_grid input to investment cost calculation functions

danielolsen:

### Purpose
Currently, the user-facing functions to calculate investment cost for each of {ac branches, dc branches, generation/storage} automatically compare against the base Grid in the repository. This PR adds a new parameter `base_grid` to each of these functions, so that another Grid can be passed instead.

This is useful because the base Grid in the repository does not correspond to any particular point in time, so calculating investment costs between for example a 2021 scenario and a 2030 scenario currently involves calculating the investment costs of each of them compared to the base grid in the repository, and then comparing those differences against each other. If instead the user can say that the 2021 scenario is the 'base' grid to compare against, then all of the implementation details of comparing the two deltas is avoided.

### What the code is doing
If a `base_grid` is provided to any of the functions, this object is used to calculate investment costs. If this parameter is not specified, then the code proceeds as currently by spinning up a new base Grid.

### Testing
No tests were performed. Currently, this instantiation of a base Grid limits the testability of the user-facing functions, since we can't pass a MockScenario with a MockGrid. This refactor improves this testability by enabling tests which avoid instantiating a Grid.

### Time estimate
10 minutes.

### DIFF
--- a/powersimdata/design/investment/investment_costs.py
+++ b/powersimdata/design/investment/investment_costs.py
@@ -10,6 +10,7 @@ from powersimdata.design.investment.create_mapping_files import (
     bus_to_reeds_reg,
 )
 from powersimdata.design.investment.inflation import calculate_inflation
+from powersimdata.input.check import _check_grid_models_match
 from powersimdata.input.grid import Grid
 from powersimdata.utility.distance import haversine
 
@@ -40,22 +41,32 @@ def append_keep_index_name(df1, other, *args, **kwargs):
     return new_df
 
 
-def calculate_ac_inv_costs(scenario, sum_results=True, exclude_branches=None):
+def calculate_ac_inv_costs(
+    scenario,
+    sum_results=True,
+    exclude_branches=None,
+    base_grid=None,
+):
     """Calculate cost of upgrading AC lines and/or transformers in a scenario.
     NEEM regions are used to find regional multipliers.
 
     :param powersimdata.scenario.scenario.Scenario scenario: scenario instance.
     :param bool sum_results: whether to sum data frame for each branch type. Defaults to
         True.
+    :param powersimdata.input.grid.Grid base_grid: a Grid to compare against. If None,
+        the grid model and interconnect from the ``scenario`` are used to instantiate a
+        corresponding unmodified Grid.
     :return: (*dict*) -- keys are {'line_cost', 'transformer_cost'}, values are either
         float if ``sum_results``, or pandas Series indexed by branch ID.
         Whether summed or not, values are $USD, inflation-adjusted to today.
     """
-
-    base_grid = Grid(
-        scenario.info["interconnect"].split("_"), source=scenario.info["grid_model"]
-    )
     grid = scenario.state.get_grid()
+    if base_grid is None:
+        base_grid = Grid(
+            scenario.info["interconnect"].split("_"), source=scenario.info["grid_model"]
+        )
+    else:
+        _check_grid_models_match(base_grid, grid)
 
     # find upgraded AC lines
     grid_new = cp.deepcopy(grid)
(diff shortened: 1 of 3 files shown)


powersimdata/design/investment/investment_costs.py
```diff
@@ -40,21 +40,29 @@ def append_keep_index_name(df1, other, *args, **kwargs):
...
         Whether summed or not, values are $USD, inflation-adjusted to today.
     """
-
-    base_grid = Grid(
-        scenario.info["interconnect"].split("_"), source=scenario.info["grid_model"]
-    )
+    if base_grid is None:
+        base_grid = Grid(
+            scenario.info["interconnect"].split("_"), source=scenario.info["grid_model"]
+        )
```
rouille, line 65: if `base_grid` is not None, do we need to make sure that the grid model is the same than the one in the `scenario` object?

powersimdata/design/investment/investment_costs.py
```diff
@@ -257,19 +265,23 @@ def get_branch_mult(x, bus_reg, ac_reg_mult, branch_lookup_alerted=set()):
...
         inflation-adjusted to today. If ``sum_results``, a float is returned, otherwise
         a Series.
     """
-    base_grid = Grid(
-        scenario.info["interconnect"].split("_"), source=scenario.info["grid_model"]
-    )
+    if base_grid is None:
+        base_grid = Grid(
+            scenario.info["interconnect"].split("_"), source=scenario.info["grid_model"]
+        )
```
rouille, line 284: same question here

powersimdata/design/investment/investment_costs.py
```diff
@@ -349,10 +370,10 @@ def calculate_gen_inv_costs(scenario, year, cost_case, sum_results=True):
...
         curves.
     """
-
-    base_grid = Grid(
-        scenario.info["interconnect"].split("_"), source=scenario.info["grid_model"]
-    )
+    if base_grid is None:
+        base_grid = Grid(
+            scenario.info["interconnect"].split("_"), source=scenario.info["grid_model"]
+        )
```
rouille, line 376: same question here.

powersimdata/design/investment/investment_costs.py
BainanXia: I think the purpose of having user defined `base_grid` here is to provide the capability of comparing the difference between the grid in the `scenario` and any grid instead of the `base_grid` in PowerSimdata only.

rouille: @BainanXia, my question is can we have a usa_tamu model in the `scenario` object, i.e.,`scenario.get_grid().grid_model == "usa_tamu"`), and a "hifld" model for the grid that is passed, i.e., `base_grid.grid_model == "hifld"`

danielolsen: I agree, at the very least we want a check that the grid models are the same. We _may_ also want a check that the interconnections match too, or we need to be sure that we have well-defined behavior for comparing just the overlapping parts of two Grids (e.g. comparing a Western expanded scenario to a USA base scenario).

danielolsen: Done. We now check for the matching grid model, and we should be able to handle interconnections via the reindexing logic that we add in #507.

rouille:
I guess that once we merge #507 we will be able to quickly check that the scenario object that is passed to the functions in the `investment_cost module` is in the analyze state.

rouille (approved):
Nice

danielolsen:
> I guess that once we merge #507 we will be able to quickly check that the scenario object that is passed to the functions in the `investment_cost module` is in the analyze state.

I don't think we want to prohibit a Create state scenario from investment cost analysis, do we? All we need is `get_grid()`.

PR #507, opened 2021-06-23
refactor: add compare module, move Grid/Scenario check functions from PostREISE to PowerSimData

danielolsen:

### Purpose
- Add a new module, `powersimdata.design.compare`, to consolidate code which compares Grid data tables.
- In the process of moving code from PostREISE to PowerSimData, move associated check functions which do not depend on PCM outputs.

### What the code is doing
- `powersimdata.design.compare` gets comparison logic that had been performed elsewhere in PowerSimData or PostREISE.
- `powersimdata.design.investment` uses the newly consolidated comparison logic.
- `powersimdata.input.check` and `powersimdata.scenario.check` get the Grid and Scenario check functions (with associated tests).

### Testing
Existing unit tests still pass. The existing comparison logic in PowerSimData was untested, but now has a path to getting tests, since it is separated from the broader investment cost calculations.

### Where to look
The only real new code is in `powersimdata.design.compare.generation`, since it is more robust than what we had before in `calculate_gen_inv_costs`. Everything else is just moved from elsewhere, with references updated.

### Time estimate
15-30 minutes.

### DIFF
--- a/powersimdata/design/compare/generation.py
+++ b/powersimdata/design/compare/generation.py
@@ -1,0 +1,25 @@
+from powersimdata.design.compare.helpers import _reindex_as_necessary
+from powersimdata.input.check import _check_data_frame
+
+
+def calculate_plant_difference(plant1, plant2):
+    """Calculate the capacity differences between two plant data frames. If capacity in
+    ``plant2`` is larger than capacity in ``plant1``, the return will be positive.
+
+    :param pandas.DataFrame plant1: first plant data frame.
+    :param pandas.DataFrame plant2: second plant data frame.
+    :return: (*pandas.DataFrame*) -- merged data frames with a new 'diff' column.
+    """
+    _check_data_frame(plant1, "plant1")
+    _check_data_frame(plant2, "plant2")
+    # Reindex so that we don't get NaN when calculating upgrades for new generators
+    plant1, plant2 = _reindex_as_necessary(plant1, plant2, ["bus_id", "type"])
+    plant_merge = plant1.merge(
+        plant2, how="outer", right_index=True, left_index=True, suffixes=(None, "_2")
+    )
+    plant_merge["diff"] = plant_merge.Pmax_2.fillna(0) - plant_merge.Pmax.fillna(0)
+    # Ensure that lats & lons get filled in as necessary from plant2 entries
+    for l in ["lat", "lon"]:
+        plant_merge[l].fillna(plant_merge[f"{l}_2"], inplace=True)
+
+    return plant_merge
--- a/powersimdata/design/investment/investment_costs.py
+++ b/powersimdata/design/investment/investment_costs.py
@@ -4,6 +4,11 @@ import warnings
 import numpy as np
 import pandas as pd
 
+from powersimdata.design.compare.generation import calculate_plant_difference
+from powersimdata.design.compare.transmission import (
+    calculate_branch_difference,
+    calculate_dcline_differences,
+)
 from powersimdata.design.investment import const
 from powersimdata.design.investment.create_mapping_files import (
     bus_to_neem_reg,
@@ -68,12 +73,11 @@ def calculate_ac_inv_costs(
     else:
         _check_grid_models_match(base_grid, grid)
 
-    # find upgraded AC lines
     grid_new = cp.deepcopy(grid)
-    # Reindex so that we don't get NaN when calculating upgrades for new branches
-    base_grid.branch = base_grid.branch.reindex(grid_new.branch.index).fillna(0)
-    grid_new.branch.rateA = grid.branch.rateA - base_grid.branch.rateA
-    grid_new.branch = grid_new.branch[grid_new.branch.rateA != 0.0]
+    # find upgraded AC lines
+    capacity_difference = calculate_branch_difference(base_grid.branch, grid.branch)
+    grid_new.branch = grid.branch.assign(rateA=capacity_difference.diff)
+    grid_new.branch = grid_new.branch.query("rateA != 0.0")
     if exclude_branches is not None:
         present_exclude_branches = set(exclude_branches) & set(grid_new.branch.index)
         grid_new.branch.drop(index=present_exclude_branches, inplace=True)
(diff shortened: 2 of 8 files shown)


powersimdata/scenario/check.py
```diff
@@ -0,0 +1,14 @@
+from powersimdata.scenario.analyze import Analyze
+from powersimdata.scenario.scenario import Scenario
+
+
+def _check_scenario_is_in_analyze_state(scenario):
+    """Ensure that scenario is a Scenario object in the analyze state.
```
rouille, line 6: I think we want an empty line after the description.

powersimdata/input/check.py
```diff
@@ -32,6 +35,15 @@ def check_grid(grid):
...
                 f"Exception during {check.__name__}: {sys.exc_info()[1]!r}"
             )
+    # Run checks which operate on a pandas data frame
+    for gencost_key in ("before", "after"):
+        try:
+            _check_gencost(grid.gencost[gencost_key], error_messages)
+        except Exception:
+            error_messages.append(
+                f"Exception during {check.__name__} {gencost_key}: "
+                "{sys.exc_info()[1]!r}"
```
jon-hagg, line 45: Not sure if this needs an "f" on the second line

powersimdata/input/check.py
```diff
@@ -210,3 +222,362 @@ def _check_connected_components(grid, error_messages):
...
+        raise ValueError(label + " must be a time series")
+
+
+def _check_grid_type(grid):
+    """Ensure that ``grid`` is a Grid object.
+    :param powersimdata.input.grid.Grid grid: a Grid instance.
+    :raises TypeError: if input is not a Grid instance.
+    """
+    if not isinstance(grid, Grid):
+        raise TypeError("grid must be a powersimdata.input.grid.Grid object")
```
jon-hagg, line 267: Could use `f"grid must be a {Grid} object"` to avoid hard coding the type. 
It will print out `<class 'powersimdata.input.grid.Grid'>`, so still readable.

jon-hagg: I guess a similar comment would apply to the various places we raise `TypeError`

powersimdata/input/check.py
danielolsen: You're right.

powersimdata/input/check.py
danielolsen: That's clever, I like it.

powersimdata/scenario/check.py
danielolsen: Done.

powersimdata/input/check.py
danielolsen: Done.

powersimdata/input/check.py
danielolsen: Done.

powersimdata/design/compare/generation.py
```diff
@@ -0,0 +1,31 @@
...
+    plant_merge = plant1.merge(
+        plant2, how="outer", right_index=True, left_index=True, suffixes=(None, "_2")
+    )
+    plant_merge["diff"] = plant_merge.Pmax_2.fillna(0) - plant_merge.Pmax.fillna(0)
+    # Ensure that lats & lons get filled in as necessary from plant2 entries
+    latlon_columns = ["lat", "lon"]
+    missing_lat_indices = plant_merge.query("lat.isnull()").index
+    plant_merge.loc[missing_lat_indices, latlon_columns] = plant_merge.loc[
+        missing_lat_indices, [f"{l}_2" for l in latlon_columns]
+    ]
```
BainanXia, line 28: Maybe we could do following here:
```python
for l in ["lat", "lon"]:
   plant_merge[l].fillna(plant_merge[l+"_2"], inplace=True)
```

powersimdata/design/investment/investment_costs.py
```diff
@@ -4,6 +4,8 @@
 import numpy as np
 import pandas as pd
 
+from powersimdata.design.compare import generation as compare_generation
+from powersimdata.design.compare import transmission as compare_transmission
```
BainanXia, line 8: Why don't we do `from powersimdata.design.compare.generation import calculate_plant_difference` directly here?

powersimdata/input/check.py
```diff
@@ -210,3 +222,362 @@ def _check_connected_components(grid, error_messages):
             f"but is specified as having {num_interconnects} interconnects: "
             f"{grid.interconnect}."
         )
+
+
+def _check_data_frame(df, label):
+    """Ensure that input is a pandas data frame.
```
BainanXia, line 228: Inspired by @rouille 's comments below, do we want an empty line here after the description as well? Same comments for the rest functions in this module.

powersimdata/design/investment/investment_costs.py
danielolsen: Originally I had called the generation function `calculate_difference`, but you are right that now that it has a more descriptive name, we can import the function directly rather than importing the module.

powersimdata/design/compare/generation.py
danielolsen: Nice, done.

powersimdata/design/investment/investment_costs.py
danielolsen: Done.

powersimdata/input/check.py
danielolsen: Done.

BainanXia (approved):
This is nice. Thanks.

rouille (approved):

danielolsen:
Per discussion at https://github.com/Breakthrough-Energy/PostREISE/pull/310, we probably also want to move `postreise.analyze.helpers` into `powersimdata.input.helpers`. Do we want to do this in this PR, or a follow-up?

rouille:
> Per discussion at https://github.com/Breakthrough-Energy/PostREISE/pull/310, we probably also want to move `postreise.analyze.helpers` into `powersimdata.input.helpers`. Do we want to do this in this PR, or a follow-up?

Your call

PR #508, opened 2021-06-24
ci: code coverage reporting

jon-hagg:
### Purpose
Upload code coverage report to [codecov.io](https://about.codecov.io/) for each merge to develop branch.

### What the code is doing
Created a separate workflow which is partly duplicated from the test.yml rather than make that more complicated with various conditionals. This keeps the stdout report for per-commit tests but uses the xml report required for uploading to codecov. 

### Testing
Ran the workflow on this branch, and got a report uploaded, which is available [here](https://codecov.io/gh/Breakthrough-Energy/PowerSimData/commit/434a083c7fcf6e75c64cae84a40850ce8ccede73/). The files and graphs tabs are probably the best place to explore.

### Time estimate
10 mins

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,36 @@
+name: Code coverage
+
+on:
+  push:
+    branches:
+      - 'develop'
+
+jobs:
+  coverage:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


rouille (approved):
LGTM

PR #510, opened 2021-06-28
feat: deduplicate interconnect passed to grid constructor

jon-hagg:
### Purpose
Improve experience of using the grid object by allowing users to provide interconnects as a `set`, and removing duplicates if a `list` is provided. Basically, eliminating the requirement of calling code to do `Grid(list(set(interconnect)))`.

### What the code is doing
Described above, and some test updates. Also, include tests in code coverage report.

### Testing
There are good unit tests already, just modified as needed.

### Usage Example/Visuals
```
In [1]: from powersimdata import Grid

In [2]: Grid({'Eastern', 'Texas'})
Reading bus.csv
Reading plant.csv
Reading gencost.csv
Reading branch.csv
Reading dcline.csv
Reading sub.csv
Reading bus2sub.csv
Reading zone.csv
Out[2]: <powersimdata.input.grid.Grid at 0x10f376cd0>

In [3]: Grid(['Texas', 'Texas'])
Reading bus.csv
Reading plant.csv
Reading gencost.csv
Reading branch.csv
Reading dcline.csv
Reading sub.csv
Reading bus2sub.csv
Reading zone.csv
Out[3]: <powersimdata.input.grid.Grid at 0x10f376df0>

In [4]: grid = _

In [5]: grid.interconnect
Out[5]: ['Texas']
```

### Time estimate
10 min

### DIFF
--- a/powersimdata/network/usa_tamu/model.py
+++ b/powersimdata/network/usa_tamu/model.py
@@ -77,7 +77,7 @@ class TAMU(AbstractGrid):
 def check_and_format_interconnect(interconnect):
     """Checks interconnect.
 
-    :param str/list/set interconnect: interconnect name(s).
+    :param str/iterable interconnect: interconnect name(s).
     :return: (*list*) -- interconnect(s)
     :raises TypeError: if parameter has wrong type.
     :raises ValueError: if interconnect not found or combination of interconnect is not
@@ -85,10 +85,11 @@ def check_and_format_interconnect(interconnect):
     """
     if isinstance(interconnect, str):
         interconnect = [interconnect]
-    if not isinstance(interconnect, (set, list)):
-        raise TypeError("interconnect must be one of str/list/set")
+    try:
+        interconnect = sorted(set(interconnect))
+    except:  # noqa
+        raise TypeError("interconnect must be either str or an iterable of str")
 
-    interconnect = list(set(interconnect))
     possible = ["Eastern", "Texas", "Western", "USA"]
     if any(i for i in interconnect if i not in possible):
         raise ValueError("Wrong interconnect. Choose from %s" % " | ".join(possible))
--- a/powersimdata/network/usa_tamu/model.py
+++ b/powersimdata/network/usa_tamu/model.py
@@ -77,7 +77,7 @@ class TAMU(AbstractGrid):
 def check_and_format_interconnect(interconnect):
     """Checks interconnect.
 
-    :param str/list interconnect: interconnect name(s).
+    :param str/list/set interconnect: interconnect name(s).
     :return: (*list*) -- interconnect(s)
     :raises TypeError: if parameter has wrong type.
     :raises ValueError: if interconnect not found or combination of interconnect is not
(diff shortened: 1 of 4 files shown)


BainanXia (approved):
Make sense. Thanks for taking care of the tests.

powersimdata/network/usa_tamu/model.py
```diff
@@ -77,26 +77,22 @@ def _drop_interconnect(self):
 def check_and_format_interconnect(interconnect):
     """Checks interconnect.
 
-    :param str/list interconnect: interconnect name(s).
+    :param str/list/set interconnect: interconnect name(s).
```
danielolsen, line 80: Should this be `str/iterable`? Since e.g. a tuple would also work properly.

powersimdata/network/usa_tamu/model.py
```diff
@@ -77,26 +77,22 @@ def _drop_interconnect(self):
...
         appropriate.
     """
     if isinstance(interconnect, str):
         interconnect = [interconnect]
-    if not isinstance(interconnect, list):
-        raise TypeError("interconnect must be a str or list of str")
+    if not isinstance(interconnect, (set, list)):
+        raise TypeError("interconnect must be one of str/list/set")
 
+    interconnect = list(set(interconnect))
```
danielolsen, line 91: Since we are creating a list internally, should we replace `list` with `sorted` to achieve predictable results, no matter the returned order of set elements?

powersimdata/network/usa_tamu/model.py
```diff
@@ -77,26 +77,22 @@ def _drop_interconnect(self):
...
     :raises TypeError: if parameter has wrong type.
     :raises ValueError: if interconnect not found or combination of interconnect is not
         appropriate.
     """
     if isinstance(interconnect, str):
         interconnect = [interconnect]
-    if not isinstance(interconnect, list):
-        raise TypeError("interconnect must be a str or list of str")
+    if not isinstance(interconnect, (set, list)):
+        raise TypeError("interconnect must be one of str/list/set")
```
danielolsen, line 89: I think we could raise the TypeError as the `except` block of a `try`/`except` block enclosing the `set` call below. Then anything that could be used to create a set will work, and anything else will raise the Exception with more context.

jon-hagg: Yeah that would make sense if we want to allow any iterable without having to list them all in the `isinstance` check. Hopefully anything realistic would have the expected behavior. E.g. we already handle `str`, but `set('abc')` returns `{'a', 'b', 'c'}` so if we didn't special case that we would probably get an error.

jon-hagg: One example: if we pass a `dict`, say `{'Texas': 'whatever'}`, it will pass this validation but end up failing when we try to construct a cache key, but that's a lucky coincidence. I think it's still worth it for the flexibility, just something to be aware of.

danielolsen: But we already pre-process a `str` into a `set`, right? So by the time we call `set(interconnect)`, any iterable or `str` should work?

jon-hagg: Yeah `str` works fine, just meant as an example where converting an iterable to a set might have weird behavior. I guess `dict` would actually work too - forgot the cache key stuff happens before this code is reached, so can disregard my previous comment.

danielolsen (approved):
Thanks

PR #512, opened 2021-06-29
refactor: move PostREISE helper functions to PowerSimData

danielolsen:

### Purpose
We are moving helper functions from PostREISE to PowerSimData. This is motivated by conversation in https://github.com/Breakthrough-Energy/PostREISE/pull/310; we were discussing the companion PR to #507, which did the same thing for check functions, and we noticed that there is another module that doesn't need to be PostREISE-specific.

### What the code is doing
Most of the changes are straight copy/paste from `postreise.analyze.helpers`. The exception is in **powersimdata/input/check.py**, in which we move the import of the Grid object into the `_check_grid_type` function to avoid a circular import (since the `powersimdata.input.grid` module imports `powersimdata.input.helpers` which now imports `powersimdata.input.check`).

### Testing
All unit tests, including the ones copied from PostREISE, still pass.

### Time estimate
15 minutes.

### DIFF
--- a/powersimdata/input/check.py
+++ b/powersimdata/input/check.py
@@ -5,7 +5,8 @@ import networkx as nx
 import numpy as np
 import pandas as pd
 
-from powersimdata.input.grid import Grid
+# Importing the module, not anything in it, to avid a circular import
+import powersimdata.input.grid as _grid
 from powersimdata.network.model import ModelImmutables
 
 
@@ -281,8 +282,8 @@ def _check_grid_type(grid):
     :param powersimdata.input.grid.Grid grid: a Grid instance.
     :raises TypeError: if input is not a Grid instance.
     """
-    if not isinstance(grid, Grid):
-        raise TypeError(f"grid must be a {Grid} object")
+    if not isinstance(grid, _grid.Grid):
+        raise TypeError(f"grid must be a {_grid.Grid} object")
 
 
 def _check_areas_and_format(areas, grid_model="usa_tamu"):
--- a/powersimdata/input/check.py
+++ b/powersimdata/input/check.py
@@ -41,7 +41,7 @@ def check_grid(grid):
             _check_gencost(grid.gencost[gencost_key], error_messages)
         except Exception:
             error_messages.append(
-                f"Exception during {check.__name__} {gencost_key}: "
+                f"Exception during _check_gencost: {gencost_key}: "
                 f"{sys.exc_info()[1]!r}"
             )
     if len(error_messages) > 0:
--- a/powersimdata/input/helpers.py
+++ b/powersimdata/input/helpers.py
@@ -1,6 +1,16 @@
 import os
+from collections import defaultdict
 
 import pandas as pd
+
+from powersimdata.input.check import (
+    _check_areas_are_in_grid_and_format,
+    _check_data_frame,
+    _check_grid_type,
+    _check_plants_are_in_grid,
+    _check_resources_are_in_grid_and_format,
+)
+from powersimdata.network.model import area_to_loadzone
 
 
 def csv_to_data_frame(data_loc, filename):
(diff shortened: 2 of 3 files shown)


powersimdata/input/helpers.py
```diff
@@ -120,3 +130,325 @@ def get_interconnect(idx):
...
+    :param dict areas: areas to use for decomposition. Keys are area types
+        ('*loadzone*', '*state*' or '*interconnect*'), values are
+        str/list/tuple/set of areas.
+    :param str/list/tuple/set resources: resource(s) to use for decomposition.
+    :param powersimdata.input.grid.Grid grid: Grid instance.
+    :return: (*dict*) -- keys are areas, values are dictionaries whose keys are
+        resources and values are data frames indexed by (datetime, plant) where plant
+        include only plants of matching type and located in area.
+    """
+    # areas_resources = defaultdict(dict)
```
jon-hagg, line 329: Looks like we can remove the commented out line

powersimdata/input/check.py
```diff
@@ -281,6 +280,9 @@ def _check_grid_type(grid):
     :param powersimdata.input.grid.Grid grid: a Grid instance.
     :raises TypeError: if input is not a Grid instance.
     """
+    # Import within this function to avoid circular import issue
+    from powersimdata.input.grid import Grid
+
```
rouille, line 285: What is going on here?

danielolsen: The `powersimdata.input.grid` module imports `powersimdata.input.helpers` which now imports `powersimdata.input.check` so if `powersimdata.input.check` imports `powersimdata.input.grid` in the module definition then we are in trouble.

jon-hagg: I don't really mind this way, but another option is to just import the module (keeping it at the top of the file) e.g. 

```
import powersimdata.input.grid as _grid
```

danielolsen: I don't think it matters what we call `powersimdata.input.grid` after we import it; if we import it at all, we get circular imports.

powersimdata/input/helpers.py
danielolsen: Done, with some additional simplification of the function as well.

powersimdata/input/check.py
jon-hagg: Difference is doing `import module` vs `from module import thing` - the former doesn't seem to execute the module contents so doesn't cause a circular import. So it's an option if we want to keep imports at the top.

danielolsen: TIL! Done.

danielolsen: I guess `import module` just needs to find the module, but `from module import thing` means that the module has to be inspected, which means that any imports from that module are executed?

jon-hagg:
Not part of this PR, but I noticed that `check` is unbound on [this line](https://github.com/Breakthrough-Energy/PowerSimData/blob/8b15e522053dafe609e7b1be1c8815ceb43ebec8/powersimdata/input/check.py#L44), should be `_check_gencost` I think.

powersimdata/input/check.py
rouille: Nice. I prefer to have imports at the top of the module

danielolsen:
> 
> 
> Not part of this PR, but I noticed that `check` is unbound on [this line](https://github.com/Breakthrough-Energy/PowerSimData/blob/8b15e522053dafe609e7b1be1c8815ceb43ebec8/powersimdata/input/check.py#L44), should be `_check_gencost` I think.

Good catch, fixed.

powersimdata/input/check.py
jon-hagg: Actually turns out either way needs to run all the lines in the module (tested by adding print statements). But nevertheless, `import module` still fixes the circular import, not sure why..

jon-hagg (approved):

PR #513, opened 2021-06-29
Refactor DataAccess classes to use fsspec

jon-hagg:
### Purpose
Consolidate code using fsspec api to handle both local and ssh implementations the same way. Get some initial feedback and validate if we are heading in the right direction, as well as possible suggestions for further improvements. 

### What the code is doing
The commits are broken down fairly small, so looking at those one at a time might be the easiest. 

We define a `self.fs` property on each `DataAccess` subclass which hold the filesystem object that does most of the work. Various `DataAccess` methods are either replaced or simplified using fsspec methods. To preserve progress bars and the current checksum functionality, we subclass the `SFTPFileSystem` and register our implementation to map to the `ssh` protocol. Most of the other changes are just cleaning up code or derived from the changes to data_access.py. 

### Testing
Some light testing using standalone and client/server container setup, but more needs to be done.

### Usage Example/Visuals
If everything is right, there should be no differences to the end user experience.

### Time estimate
30 min


jon-hagg:
Closing this one, will replace with PR based off new branch

PR #517, opened 2021-07-02
refactor: create class variables for supported models and engines

goccert25:

### Purpose
What is the larger goal of this change?
Create a single place where supported models and engines are listed for the `Grid` class. This will allow the download manager project to easily use those supported models and engines as choices for CLI parameters

### What the code is doing
How is the purpose executed?
Adding class variables to `Grid`

### Testing
unit tests

### Where to look
N/A

### Usage Example/Visuals
N/A

### Time estimate
How long will it take for reviewers and observers to understand this code change?
1 minute

### DIFF
--- a/powersimdata/input/grid.py
+++ b/powersimdata/input/grid.py
@@ -12,12 +12,18 @@ _cache = MemoryCache()
 
 
 class Grid:
+
+    SUPPORTED_MODELS = {"usa_tamu"}
+    SUPPORTED_ENGINES = {"REISE", "REISE.jl"}
+
     """Grid
 
     :param str/iterable interconnect: geographical region covered. Either *'USA'*, one of
         the three interconnections, i.e., *'Eastern'*, *'Western'* or *'Texas'* or a
         combination of two interconnections.
-    :param str source: model used to build the network.
+    :param str source: model used to build the network. Can be one of the
+        supported models ("usa_tamu"), or a .mat file that represents a
+        grid.
     :param str engine: engine used to run scenario, if using ScenarioGrid.
     :raises TypeError: if source and engine are not both strings.
     :raises ValueError: if source or engine does not exist.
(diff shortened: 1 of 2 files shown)


powersimdata/input/grid.py
```diff
@@ -27,9 +31,12 @@ def __init__(self, interconnect, source="usa_tamu", engine="REISE"):
         """Constructor."""
         if not isinstance(source, str):
             raise TypeError("source must be a str")
-        supported_engines = {"REISE", "REISE.jl"}
-        if engine not in supported_engines:
-            raise ValueError(f"Engine must be one of {','.join(supported_engines)}")
+        if source not in self.SUPPORTED_MODELS:
```
jon-hagg, line 34: I believe `source` can also be a path to a `.mat` file which contains a modified grid. See L#47 where this is handled.

goccert25: @jon-hagg updated the input validation to account for this. I think long term though the `source` parameter should be refactored to be an object that's passed into this class. In other words, this logic:
```
        if cached is not None:
            data = cached
        elif source == "usa_tamu":
            data = TAMU(interconnect)
        elif os.path.splitext(source)[1] == ".mat":
            if engine == "REISE":
                data = FromREISE(source)
            elif engine == "REISE.jl":
                data = FromREISEjl(source)
```
should exist outside of the Grid class. This makes the code more maintainable by using [dependency injection](https://en.wikipedia.org/wiki/Dependency_injection). As the code base grows, adding more source/engine options and tests to this class is going to become more of a PITA. Thoughts on adding a github issue/jira ticket/asana task (not sure how action items are tracked 😅 ) for refactoring this Grid class?

jon-hagg (approved):
LGTM

powersimdata/input/grid.py
danielolsen: A github issue is probably best. This idea doesn't quite fit into our definition of a 'feature', but I think the feature request issue template is probably the best to use for a refactor like this.

rouille: We also need to factor out a lot of code from the `TAMU` class since other grid models will have the same input files (CSV) with the same columns and header.

danielolsen: Part of that work was done in my branch `daniel/read_hifld_grid`, which is behind `develop` by a bit, but most of the relevant commits should be cherry-pickable without any merge conflicts.

PR #520, opened 2021-07-08
Simplify data access using pyfilesystem

jon-hagg:
### Purpose
Consolidate data access code by using the api provided by pyfilesystem which provides consistency across local, ssh, in memory, and future (blob storage) backed storage. 

### What the code is doing
The first group of commits is using fsspec to achieve the same goal, but then is replaced by pyfilesystem which I think matches our usage better. 

We define a `self.fs` property on each `DataAccess` subclass which hold the filesystem object that does most of the work. Additionally, `SSHDataAccess` also has a `self.local_fs`, and all file based operations are done through a pyfilesystem object, using pyfilesystem paths which are internally converted to the correct platform. Various `DataAccess` methods are either replaced or simplified using pyfilesystem methods. To preserve progress bars and the current checksum functionality, we customize the default `SSHFS` instance by wrapping it in a `SubFS` (pyfilesystem builtin meant for this purpose).

### Testing
I've done testing throughout the process, but need to run through them again with the latest changes. I will update this accordingly. It will include creating and launching a scenario over ssh, delete/move states, etc. 

### Where to look
There is a lot, but mostly originating with changes to `data_access.py`. The tests also now use an in memory filesystem which simplifies the setup, and might enable increasing coverage further.

### Time estimate
1+ hour(s). We should review in person


powersimdata/data_access/data_access.py
```diff
@@ -49,89 +46,58 @@ def move_to(self, file_name, to_dir, change_name_to=None):
...
-        _allowed = list(_dirs.keys())
-        if kind not in _allowed:
-            raise ValueError(f"Invalid 'kind', must be one of {_allowed}")
-
-        root = self.root if not backup else self.backup_root
-        return self.join(root, *_dirs[kind])
-
-    def match_scenario_files(self, scenario_id, kind, backup=False):
+    def tmp_folder(self, scenario_id):
         """Get path matching the given kind of scenario data
```
rouille, line 50: Description needs to be updated:  Get path to scenario temporary folder?

powersimdata/data_access/data_access.py
```diff
@@ -49,89 +46,58 @@ def move_to(self, file_name, to_dir, change_name_to=None):
...
-        :param str dest: destination path
-        :param bool recursive: create directories recursively
-        :param bool update: only copy if needed
-        """
-        raise NotImplementedError
+        return self.join(server_setup.EXECUTE_DIR, f"scenario_{scenario_id}")
 
-    def remove(self, target, recursive=False, confirm=True):
-        """Wrapper around rm command
+    def copy(self, src, dest):
```
rouille, line 57: Should we rename `src` to `path`? It is more intuitive and this is what is used in the `_check_file_exists` function

powersimdata/data_access/data_access.py
```diff
@@ -49,89 +46,58 @@ def move_to(self, file_name, to_dir, change_name_to=None):
...
+        """Copy file to new location
 
-        :param str target: path to remove
-        :param bool recursive: delete directories recursively
-        :param bool confirm: prompt before executing command
+        :param str src: path to file
+        :param str dest: destination folder
         """
-        raise NotImplementedError
+        if self.fs.exists(dest) and self.fs.isdir(dest):
```
rouille, line 63: Not sure I understand, `self.fs` is only define in the `LocalDataAccess` class

jon-hagg: It also refers to the property defined in `SSHDataAccess`. So it will be defined on any subclass instance, although not enforced by the base class (we could use the abc module to do this at some point).

rouille: Yeah. What would happen if someone instantiate the `DataClass` and call the `copy` method? Would the the `abc` module prevent that?

jon-hagg: Yeah, it would raise an error in that case. No one should be instantiating `DataAccess` directly though. I believe using `abc` would prevent that.

powersimdata/data_access/data_access.py
jon-hagg: Done

powersimdata/scenario/move.py
```diff
@@ -1,4 +1,9 @@
+from fs.copy import copy_dir
+from fs.walk import Walker
+
+from powersimdata.data_access.data_access import get_ssh_fs
 from powersimdata.scenario.state import State
+from powersimdata.utility import server_setup
 
 
 class Move(State):
```
rouille, line 9: `State` --> `Ready`?

rouille: I guess the `print_scenario_info` method is unnecessary

powersimdata/scenario/move.py
```diff
@@ -36,21 +41,11 @@ def move_scenario(self, target="disk", confirm=True):
             raise ValueError("scenario data can only be backed up to disk now")
 
         backup = BackUpDisk(self._data_access, self._scenario_info)
```
rouille, line 43: I think you can pass `self_scenario_info["id"]`. You could declare `sid` before creating this `BackUpDisk` instance and use it for the instantiation

powersimdata/scenario/move.py
```diff
@@ -65,27 +60,25 @@ def __init__(self, data_access, scenario_info):
         self._data_access = data_access
         self._scenario_info = scenario_info
```
rouille, line 61: I think you only need the id of the scenario. See previous [comment](https://github.com/Breakthrough-Energy/PowerSimData/pull/520/files?file-filters%5B%5D=.lock&file-filters%5B%5D=.py&file-filters%5B%5D=No+extension&hide-deleted-files=true#r677850959).

powersimdata/scenario/delete.py
```diff
@@ -38,29 +38,26 @@ def delete_scenario(self, confirm=True):
...
+        target = local_data_access.join("data", "**", f"{scenario_id}_*")
+        local_data_access.remove(target, confirm=confirm)
 
         # Delete attributes
         self._clean()
 
     def _clean(self):
         """Clean after deletion."""
-        self._data_access.close()
         self._scenario_info = None
```
rouille, line 63: Should the parent class be `Ready`. Should we remove the `print_scenario_info` method?

powersimdata/scenario/analyze.py
```diff
@@ -53,9 +53,7 @@ def refresh(self, scenario):
     def _set_allowed_state(self):
         """Sets allowed state."""
         if self._scenario_status == "extracted":
-            self.allowed = ["delete"]
-            if self._data_access.backup_root is not None:
-                self.allowed.append("move")
+            self.allowed = ["delete", "move"]
```
rouille, line 56: `self.allowed == ["delete", "move"] if self._scenario_status == "extracted" else []` ?

powersimdata/scenario/move.py
jon-hagg: Good catch, done

powersimdata/data_access/data_access.py
```diff
@@ -1,36 +1,33 @@
...
-}
+
+def get_ssh_fs(root=None):
+    if root is None:
+        root = ""
+    host = server_setup.SERVER_ADDRESS
+    port = server_setup.SERVER_SSH_PORT
+    username = server_setup.get_server_user()
+    base_fs = fs2.open_fs(f"ssh://{username}@{host}:{port}")
+    return WrapSSHFS(base_fs, root)
```
rouille, line 22: Is there any reason this function is not a method of the `SSHDataAccess` class?

powersimdata/data_access/data_access.py
```diff
@@ -279,101 +184,59 @@ def get_profile_version(self, grid_model, kind):
...
-        server_user = server_setup.get_server_user()
-        client.connect(
-            server_setup.SERVER_ADDRESS,
-            username=server_user,
-            port=server_setup.SERVER_SSH_PORT,
-            timeout=10,
-        )
-
-        self._ssh = client
+        return self._fs
```
rouille, line 222: Do we need to return the attribute?

jon-hagg: You mean `self._fs`? That is part of the `@property` defined above, so when we call `self.fs` it will create the filesystem if needed and return it.

powersimdata/data_access/data_access.py
jon-hagg: No reason. I found it convenient to be able to use interactively without creating an SSHDataAccess instance though

powersimdata/scenario/analyze.py
jon-hagg: I tried but it still gets reformatted to the following:
```
        self.allowed = (
            ["delete", "move"] if self._scenario_status == "extracted" else []
        )

```

rouille: So let's keep what we had then

powersimdata/data_access/data_access.py
rouille: Ok. Not super familiar with `@property`

powersimdata/scenario/analyze.py
BainanXia: Will this work?
```python
self.allowed = ["delete", "move"]*(self._scenario_status == "extracted")
```

jon-hagg: Whoa, what kind of magic is this. It does work, but I'd rather keep it explicit. That's pretty neat though.

powersimdata/scenario/delete.py
jon-hagg: Done - moved the try/catch to the `State` implementation so it works for the delete state as well

powersimdata/data_access/ssh_fs.py
```diff
@@ -0,0 +1,53 @@
...
+    def exec_command(self, command):
+        """Wrapper around paramiko exec_command
+
+        :param str command: the command to execute
+        :return: (*tuple*) -- standard streams
+        """
+        return self.client.exec_command(command)
+
+    def checksum(self, filepath):
+        """Return the checksum of the file path (using sha1sum)
```
rouille, line 46: I think we want an empty line here

powersimdata/data_access/ssh_fs.py
```diff
@@ -0,0 +1,53 @@
...
+    """Wrapper around another filesystem which is rooted at the given path and adds
+    progress bar for download
+    """
+
+    def __init__(self, parent_fs, path):
+        self.client = parent_fs._client
+        super().__init__(parent_fs, path)
+
+    def download(self, path, file, chunk_size=None, **options):
+        """Wrapper around pyfilesystem download with progress bar"""
```
rouille, line 32: I think we want an empty line here

powersimdata/data_access/ssh_fs.py
```diff
@@ -0,0 +1,53 @@
...
+        bar.total = int(b)
+        bar.update(int(a - last[0]))
+        last[0] = a
+
+    return show, bar
+
+
+class WrapSSHFS(SubFS):
+    """Wrapper around another filesystem which is rooted at the given path and adds
+    progress bar for download
```
rouille, line 24: should we add the parameter of the constructor in the docstring here?

powersimdata/data_access/data_access.py
```diff
@@ -1,36 +1,33 @@
...
 
-_dirs = {
-    "tmp": (server_setup.EXECUTE_DIR,),
-    "input": server_setup.INPUT_DIR,
-    "output": server_setup.OUTPUT_DIR,
-}
+
+def get_ssh_fs(root=None):
+    if root is None:
+        root = ""
```
BainanXia, line 16: Just curious why we default root to empty string, given `str` is immutable.

jon-hagg: Good point

rouille: I guess this could be applied to the `LocalDataAccess` and `SSHDataAccess` too

rouille:
@jon-hagg, the code looks good. Did you test it in the containerized framework? Should we test it in the server framework (@danielolsen, @BainanXia)?

jon-hagg:
I will test it again today with the latest changes.

Pipfile
```diff
@@ -18,3 +18,5 @@ psycopg2 = "~=2.8.5"
 scipy = "~=1.5"
 tqdm = "==4.29.1"
 requests = "~=2.25"
+fs = "*"
+"fs.sshfs" = "*"
```
BainanXia, line 22: Just curious why we have double quotes here. Is this file generated automatically via `pipenv`?

jon-hagg: Yeah it was added automatically, I believe using `pipenv install fs.sshfs`.

powersimdata/data_access/data_access.py
```diff
@@ -141,13 +105,12 @@ def execute_command_async(self, command):
...
     def checksum(self, relative_path):
-        """Return the checksum of the file path, and write the content if the
-        server is remote
+        """Return the checksum of the file path
 
         :param str relative_path: path relative to root
         :return: (*str*) -- the checksum of the file
         """
-        raise NotImplementedError
+        return "dummy_value"
```
BainanXia, line 113: The reason that we would like to return something here now is solely for simplification purpose for `LocalDataAccess` to inherit, right?

powersimdata/data_access/data_access.py
```diff
@@ -279,101 +181,59 @@ def get_profile_version(self, grid_model, kind):
...
-            bar.close()
-            os.close(tmp_file)
-        # wait for file handle to be available
-        shutil.move(tmp_path, to_path)
+        print(f"Transferring {file_name} from server")
+        with TempFS() as tmp_fs:
+            self.local_fs.makedirs(from_dir, recreate=True)
+            tmp_fs.makedirs(from_dir, recreate=True)
+            fs2.copy.copy_file(self.fs, from_path, tmp_fs, from_path)
+            fs2.move.move_file(tmp_fs, from_path, self.local_fs, from_path)
```
BainanXia, line 236: This is clean! I presume it is for the purpose if conflict protection, right?

jon-hagg: Yeah, this was probably my favorite part. And that is correct - the idea is to minimize the time we have the destination file open (e.g. ScenarioList.csv) to reduce collisions when running concurrently. Another benefit is if the download fails, e.g. for some large file, we don't end up with a partial copy.

powersimdata/data_access/data_access.py
```diff
@@ -442,9 +294,9 @@ def push(self, file_name, checksum, change_name_to=None):
         self.move_to(file_name, change_name_to=backup)
 
         values = {
-            "original": self.join(self.root, new_name),
-            "updated": self.join(self.root, backup),
-            "lockfile": self.join(self.root, "scenario.lockfile"),
+            "original": posixpath.join(self.root, new_name),
+            "updated": posixpath.join(self.root, backup),
+            "lockfile": posixpath.join(self.root, "scenario.lockfile"),
```
BainanXia, line 299: This would be something that needs to be changed via `os` so that our client and server setup will support non-linux servers?

powersimdata/data_access/data_access.py
```diff
@@ -456,86 +308,22 @@ def push(self, file_name, checksum, change_name_to=None):
...
-        """Create path on server
 
-        :param str full_path: the path, excluding filename
-        :raises IOError: if command generated stderr
-        """
-        _, _, stderr = self.ssh.exec_command(f"mkdir -p {full_path}")
-        errors = stderr.readlines()
-        if len(errors) > 0:
-            raise IOError(f"Failed to create {full_path} on server")
+class MemoryDataAccess(SSHDataAccess):
```
BainanXia, line 320: Shall we put down some doc strings here?

powersimdata/data_access/data_access.py
jon-hagg: Pretty much, yeah. I suppose it kinda makes sense to have as the default, for `DataAccess` subclasses that don't need to validate checksums. May move this around in the future though.

BainanXia (approved):
I've read through the code and learned a lot. This is definitely much cleaner. Thanks for the effort.

powersimdata/data_access/data_access.py
jon-hagg: Yeah there are several things we'd need to change in that case. As far as I know, non linux servers aren't anywhere on our roadmap.

Issue #522, opened 2021-07-13
Enable getting profiles from scenario in Execute state


danielolsen:
# :rocket:

- [ ] Is your feature request essential for your project?

### Describe the workflow you want to enable
I wish I could call `get_demand`, `get_solar`, etc. from a Scenario in Execute state.

### Describe your proposed implementation
Implement the functions the same as in Analyze. For demand, we can default to `original=True`, and raise an exception for `original=False`.

PR #523, opened 2021-07-13
feat: support getting profiles from execute state

jon-hagg:
### Purpose
Address #522

### What the code is doing
Define a new `State` subclass with methods to access transformed profiles, and have `Execute` and `Analyze` inherit from that. 

### Testing
Checked some of the methods for:
* an existing scenario (analyze state)
* a newly created scenario (create/execute state)

### Time estimate
10 min

### DIFF
--- a/powersimdata/scenario/ready.py
+++ b/powersimdata/scenario/ready.py
@@ -16,7 +16,7 @@ class Ready(State):
         "get_solar",
         "get_wind",
         "get_wind_offshore",
-    }
+    } | State.exported_methods
 
     def __init__(self, scenario):
         """Constructor."""
@@ -73,7 +73,6 @@ class Ready(State):
             turbines
         """
         wind = self.get_profile("wind")
-
         grid = self.get_grid()
         onshore_id = grid.plant.groupby(["type"]).get_group("wind").index
         return wind[onshore_id]
--- a/powersimdata/scenario/ready.py
+++ b/powersimdata/scenario/ready.py
@@ -6,6 +6,18 @@ from powersimdata.scenario.state import State
 
 
 class Ready(State):
+    exported_methods = {
+        "get_ct",
+        "get_grid",
+        "get_bus_demand",
+        "get_demand",
+        "get_hydro",
+        "get_grid",
+        "get_solar",
+        "get_wind",
+        "get_wind_offshore",
+    }
+
     def __init__(self, scenario):
         """Constructor."""
         super().__init__(scenario)
--- a/powersimdata/scenario/ready.py
+++ b/powersimdata/scenario/ready.py
@@ -1,0 +1,99 @@
+import copy
+
+from powersimdata.input.input_data import get_bus_demand
+from powersimdata.input.transform_profile import TransformProfile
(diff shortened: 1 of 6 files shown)


jon-hagg:
Working on reconfiguring the exported methods

rouille:
We should extend this to the `Create` class where we also do `get_solar`, etc.

powersimdata/scenario/ready.py
```diff
@@ -0,0 +1,127 @@
...
+        "get_wind_offshore",
+    }
+
+    def __init__(self, scenario):
+        """Constructor."""
+        super().__init__(scenario)
+
+    def _update_scenario_info(self):
+        """Updates scenario information."""
+        pass
```
rouille, line 28: How does it work? Where is the update happening?

jon-hagg: When called from analyze state, we default to this. But execute overrides the method, so it will still download the scenario list.

jon-hagg: BTW, I'm working on moving this to state.py instead, so it can also be used from the create state, which also defines its own version of `_update_scenario_info`

rouille: The `Scenario` class has a `print_scenario_info` method. Just wondering if the inheritance is not going to break somewhere.

jon-hagg: Yeah I was concerned about that. After testing again it seems to work fine. Changed it so `print_scenario_info` is exported from `scenario.state`, so the `Scenario` doesn't have to explicitly proxy.

jon-hagg:
> We should extend this to the `Create` class where we also do `get_solar`, etc.

Thinking we save that for a future PR. I don't want to do anything too crazy with inheritance without a bit more thought regarding the design.

BainanXia:
> > We should extend this to the `Create` class where we also do `get_solar`, etc.
> 
> Thinking we save that for a future PR. I don't want to do anything too crazy with inheritance without a bit more thought regarding the design.

Agree. Implementing capabilities of getting time series profiles in `create` state is a little tricky given the fact that it is possible the user hasn't set corresponding profiles yet or transformed profiles need to be generated based on an incomplete change table, which fits into a separate PR.

powersimdata/scenario/state.py
```diff
@@ -28,6 +29,19 @@ def refresh(self, scenario):
         """
         pass
 
+    def _update_scenario_info(self):
```
danielolsen, line 32: I think I see what's going on here, but maybe it would be clearer if we add a comment that this method is overwritten in the Execute class, and that's why we have a placeholder here.

danielolsen (approved):
Looks good, thanks.

BainanXia (approved):
LGTM. A presume a subsequent PR would be dealing with getting time series at create state.

powersimdata/scenario/state.py
jon-hagg: I changed the docstring to be more useful, but didn't mention specific classes so if things change later the description won't also need to change.

PR #525, opened 2021-07-15
feat: support passing extract_data for container and local simulations

jon-hagg:
### Purpose
Changes corresponding to https://github.com/Breakthrough-Energy/REISE.jl/pull/146. This exposes the extract-data parameter to a user who is interacting via the scenario object. 

### What the code is doing
The extract_data parameter is already part of the signature, but needed to be passed to REISE.jl. Also add implementations for `extract_simulation_output` so it can be called when using docker or local installation.

### Testing
Launched a simulation using docker and natively, with and without extract_data set.

### Time estimate
10 min

### DIFF
--- a/powersimdata/data_access/launcher.py
+++ b/powersimdata/data_access/launcher.py
@@ -45,6 +45,7 @@ class Launcher:
 
     def __init__(self, scenario):
         self.scenario = scenario
+        self.scenario_id = scenario.scenario_id
 
     def _launch(self, threads=None, solver=None, extract_data=True):
         """Launches simulation on target environment
@@ -102,7 +103,7 @@ class SSHLauncher(Launcher):
             "python3",
             "-u",
             path_to_script,
-            self.scenario.scenario_id,
+            self.scenario_id,
         ]
         cmd_io_redirect = ["</dev/null >/dev/null 2>&1 &"]
         cmd = cmd_pythonpath + cmd_pythoncall + extra_args + cmd_io_redirect
@@ -164,8 +165,7 @@ class HttpLauncher(Launcher):
         :return: (*dict*) -- contains "output", "errors", "scenario_id", and "status"
             keys which map to stdout, stderr, and the respective scenario attributes
         """
-        scenario_id = self.scenario.scenario_id
-        url = f"{self.BASE_URL}/launch/{scenario_id}"
+        url = f"{self.BASE_URL}/launch/{self.scenario_id}"
         params = {
             "threads": threads,
             "solver": solver,
@@ -184,8 +184,7 @@ class HttpLauncher(Launcher):
         :return: (*dict*) -- contains "output", "errors", "scenario_id", and "status"
             keys which map to stdout, stderr, and the respective scenario attributes
         """
-        scenario_id = self.scenario.scenario_id
-        url = f"{self.BASE_URL}/extract/{scenario_id}"
+        url = f"{self.BASE_URL}/extract/{self.scenario_id}"
         resp = requests.post(url)
         return resp.json()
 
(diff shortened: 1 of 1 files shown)


powersimdata/data_access/launcher.py
```diff
@@ -195,18 +218,28 @@ def _launch(self, threads=None, solver=None, extract_data=True):
         :return: (*dict*) -- contains "output", "errors", "scenario_id", and "status"
             keys which map to stdout, stderr, and the respective scenario attributes
         """
-        sys.path.append(server_setup.ENGINE_DIR)
         from pyreisejl.utility import app
```
rouille, line 221: We assume here that the engine is REISE.jl, which is fine for now. In the future, we might have other engines and we will have to access their own utility functions

powersimdata/data_access/launcher.py
```diff
@@ -195,18 +218,28 @@ def _launch(self, threads=None, solver=None, extract_data=True):
...
+            self.scenario.scenario_id, threads, solver, extract_data
+        )
+
+    def extract_simulation_output(self):
+        """Extracts simulation outputs {PG, PF, LMP, CONGU, CONGL}
+
+        :return: (*dict*) -- contains "output", "errors", "scenario_id", and "status"
+            keys which map to stdout, stderr, and the respective scenario attributes
+        """
+        from pyreisejl.utility import app
```
rouille, line 233: same comments here

powersimdata/data_access/launcher.py
```diff
@@ -195,18 +218,28 @@ def _launch(self, threads=None, solver=None, extract_data=True):
...
+        return app.extract_scenario(self.scenario.scenario_id)
 
     def check_progress(self):
         """Get the status of an ongoing simulation, if possible
 
         :return: (*dict*) -- contains "output", "errors", "scenario_id", and "status"
             keys which map to stdout, stderr, and the respective scenario attributes
         """
-        sys.path.append(server_setup.ENGINE_DIR)
         from pyreisejl.utility import app
```
rouille, line 243: and here

powersimdata/data_access/launcher.py
```diff
@@ -151,6 +151,8 @@ def check_progress(self):
 
 
 class HttpLauncher(Launcher):
+    BASE_URL = f"http://{server_setup.SERVER_ADDRESS}:5000"
+
```
rouille, line 155: should we define `scenario_id` here since it is used in `_launch`, `extract_simulation_output` and `check_progress`?

rouille: We could also have a constructor for the class where we set `scenario_id` and `base_url` as attributes of the class

powersimdata/data_access/launcher.py
```diff
@@ -163,27 +165,48 @@ def _launch(self, threads=None, solver=None, extract_data=True):
...
         scenario_id = self.scenario.scenario_id
-        url = f"http://{server_setup.SERVER_ADDRESS}:5000/status/{scenario_id}"
+        url = f"{self.BASE_URL}/status/{scenario_id}"
         resp = requests.get(url)
         return resp.json()
 
 
 class NativeLauncher(Launcher):
+    def __init__(self, scenario):
+        sys.path.append(server_setup.ENGINE_DIR)
```
rouille, line 206: Do we still need that?

powersimdata/data_access/launcher.py
rouille: Can we move the import statement at the top of the file and just do `app.launch_simulation(...)`, `app.extract_scenario(...)` and `app.check_progress(...)` ?

powersimdata/data_access/launcher.py
jon-hagg: Done for `scenario_id`. Left the base url since it's static, so only needs to be defined for the class (not per instance), even though we access it using `self`.

powersimdata/data_access/launcher.py
jon-hagg: We could use `importlib.import_module` to assign it? Not sure if there is another way. I tried just moving the import statement to the constructor as-is, but that resulted in an import error.

rouille: I was trying to avoid to have the same import statement in each function. If it is too complicated or what you propose generate more lines, let's move forward with the current implementation.

jon-hagg: Done. Usually prefer the standard imports since they can be detected by refactoring tools, etc, but this use case is not too bad, and already a bit "non standard".

rouille (approved):
Thanks

PR #527, opened 2021-07-17
refactor: use dependency injection on Grid class

goccert25:
Opening this PR for the purpose of kicking off discussion around it and to talk over the code. I'll clean up this description if this ever ends up actually merging.

### DIFF
--- a/powersimdata/design/investment/investment_costs.py
+++ b/powersimdata/design/investment/investment_costs.py
@@ -67,7 +67,8 @@ def calculate_ac_inv_costs(
     """
     grid = scenario.state.get_grid()
     if base_grid is None:
-        base_grid = Grid(
+        base_grid = Grid( # Why are we repeatedly regenerating the base_grid?
+                          # (There are 3 other instances of this in this file)
             scenario.info["interconnect"].split("_"), source=scenario.info["grid_model"]
         )
     else:
--- a/powersimdata/design/investment/investment_costs.py
+++ b/powersimdata/design/investment/investment_costs.py
@@ -16,6 +16,7 @@ from powersimdata.design.investment.create_mapping_files import (
 )
 from powersimdata.design.investment.inflation import calculate_inflation
 from powersimdata.input.check import _check_grid_models_match
+from powersimdata.input.abstract_grid import AbstractGridFactory
 from powersimdata.input.grid import Grid
 from powersimdata.utility.distance import haversine
 
@@ -67,9 +68,8 @@ def calculate_ac_inv_costs(
     """
     grid = scenario.state.get_grid()
     if base_grid is None:
-        base_grid = Grid( # Why are we repeatedly regenerating the base_grid?
-                          # (There are 3 other instances of this in this file)
-            scenario.info["interconnect"].split("_"), source=scenario.info["grid_model"]
+        base_grid = Grid(
+            model_data=AbstractGridFactory.get_or_create(scenario.info["grid_model"], scenario.info["interconnect"].split("_"))
         )
     else:
         _check_grid_models_match(base_grid, grid)
@@ -289,7 +289,7 @@ def calculate_dc_inv_costs(scenario, sum_results=True, base_grid=None):
     grid = scenario.state.get_grid()
     if base_grid is None:
         base_grid = Grid(
-            scenario.info["interconnect"].split("_"), source=scenario.info["grid_model"]
+            model_data=AbstractGridFactory.get_or_create(scenario.info["grid_model"], scenario.info["interconnect"].split("_"))
         )
     else:
         _check_grid_models_match(base_grid, grid)
@@ -382,7 +382,7 @@ def calculate_gen_inv_costs(
     grid = scenario.state.get_grid()
     if base_grid is None:
         base_grid = Grid(
-            scenario.info["interconnect"].split("_"), source=scenario.info["grid_model"]
+            model_data=AbstractGridFactory.get_or_create(scenario.info["grid_model"], scenario.info["interconnect"].split("_"))
         )
     else:
         _check_grid_models_match(base_grid, grid)
--- a/powersimdata/input/abstract_grid.py
+++ b/powersimdata/input/abstract_grid.py
@@ -1,6 +1,7 @@
 import pandas as pd
 
 from powersimdata.input import const
+from powersimdata.utility.helpers import MemoryCache, cache_key
 
 
 class AbstractGrid:
@@ -43,3 +44,19 @@ def storage_template():
         "terminal_max": None,
     }
     return storage
+
+
+class AbstractGridFactory:
+    _cache = MemoryCache()
+
+    @classmethod
+    def get_or_create(cls, abstract_grid_class, *args):
+        assert(issubclass(abstract_grid_class, AbstractGrid))
+        key = cache_key(abstract_grid_class.__name__, *args)
+        cached = cls._cache.get(key)
+        if cached is not None:
+            return cached
+
+        new_abstract_grid = abstract_grid_class.__init__(*args)
+        cls._cache.put(key, new_abstract_grid)
+        return new_abstract_grid
--- a/powersimdata/input/grid.py
+++ b/powersimdata/input/grid.py
@@ -29,44 +29,20 @@ class Grid:
     :raises ValueError: if source or engine does not exist.
     """
 
-    def __init__(self, interconnect, source="usa_tamu", engine="REISE"):
+    def __init__(self, model_data):
         """Constructor."""
-        if not isinstance(source, str):
-            raise TypeError("source must be a str")
-        if source not in self.SUPPORTED_MODELS and not source.endswith(".mat"):
-            raise ValueError(
-                f"Source must be one of {','.join(self.SUPPORTED_MODELS)} "
-                "or the path to a .mat file that represents a grid "
-            )
-        if engine not in self.SUPPORTED_ENGINES:
-            raise ValueError(
-                f"Engine must be one of {','.join(self.SUPPORTED_ENGINES)}"
-            )
-
-        key = cache_key(interconnect, source)
-        cached = _cache.get(key)
-        if cached is not None:
-            data = cached
-        elif source == "usa_tamu":
-            data = TAMU(interconnect)
-        elif os.path.splitext(source)[1] == ".mat":
-            if engine == "REISE":
-                data = FromREISE(source)
-            elif engine == "REISE.jl":
-                data = FromREISEjl(source)
-
-        self.data_loc = data.data_loc
-        self.interconnect = data.interconnect
-        self.zone2id = data.zone2id
-        self.id2zone = data.id2zone
-        self.sub = data.sub
-        self.plant = data.plant
-        self.gencost = data.gencost
-        self.dcline = data.dcline
-        self.bus2sub = data.bus2sub
-        self.bus = data.bus
-        self.branch = data.branch
-        self.storage = data.storage
+        self.data_loc = model_data.data_loc
+        self.interconnect = model_data.interconnect
+        self.zone2id = model_data.zone2id
+        self.id2zone = model_data.id2zone
+        self.sub = model_data.sub
+        self.plant = model_data.plant
+        self.gencost = model_data.gencost
+        self.dcline = model_data.dcline
+        self.bus2sub = model_data.bus2sub
+        self.bus = model_data.bus
+        self.branch = model_data.branch
+        self.storage = model_data.storage
 
         _cache.put(key, self)
 
--- a/powersimdata/input/grid.py
+++ b/powersimdata/input/grid.py
@@ -8,8 +8,6 @@ from powersimdata.network.usa_tamu.constants import storage as tamu_storage
 from powersimdata.network.usa_tamu.model import TAMU
 from powersimdata.utility.helpers import MemoryCache, cache_key
 
-_cache = MemoryCache()
-
 
 class Grid:
 
@@ -43,8 +41,6 @@ class Grid:
         self.bus = model_data.bus
         self.branch = model_data.branch
         self.storage = model_data.storage
-
-        _cache.put(key, self)
 
         self.grid_model = self._get_grid_model()
         self.model_immutables = ModelImmutables(self.grid_model)
--- a/powersimdata/scenario/create.py
+++ b/powersimdata/scenario/create.py
@@ -9,11 +9,14 @@ from powersimdata.input.change_table import ChangeTable
 from powersimdata.input.grid import Grid
 from powersimdata.input.input_data import InputData, get_bus_demand
 from powersimdata.input.transform_grid import TransformGrid
+from powersimdata.input.scenario_grid import FromREISEjl
 from powersimdata.input.transform_profile import TransformProfile
 from powersimdata.network.model import ModelImmutables
 from powersimdata.scenario.execute import Execute
 from powersimdata.scenario.state import State
 from powersimdata.utility import server_setup
+from powersimdata.input.abstract_grid import AbstractGridFactory
+from powersimdata.network.usa_tamu.model import TAMU
 
 
 class Create(State):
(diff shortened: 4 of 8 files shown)


powersimdata/input/grid.py
```diff
@@ -29,44 +29,20 @@ class Grid:
...
-        cached = _cache.get(key)
-        if cached is not None:
-            data = cached
-        elif source == "usa_tamu":
-            data = TAMU(interconnect)
-        elif os.path.splitext(source)[1] == ".mat":
-            if engine == "REISE":
-                data = FromREISE(source)
-            elif engine == "REISE.jl":
-                data = FromREISEjl(source)
```
goccert25, line 56: Ultimately, I'd like to move the initialization of the `data` variable _outside_ of this `Grid` class. This will make the class more maintainable on the long run. If you add 30 more models, you'll have 30 branching if statements here, vs if you initialize the `data` variable outside of the class and pass it in, you would just have
`Grid(TAMU(interconnect))` or `Grid(OTHERMODEL(interconnect))`.

powersimdata/design/investment/investment_costs.py
```diff
@@ -67,7 +67,8 @@ def calculate_ac_inv_costs(
     """
     grid = scenario.state.get_grid()
     if base_grid is None:
-        base_grid = Grid(
+        base_grid = Grid( # Why are we repeatedly regenerating the base_grid?
+                          # (There are 3 other instances of this in this file)
```
goccert25, line 71: This repeated initialization (3 times in this file) of the `Grid` is pretty rough. What do you all think about adding `base_grid` to the scenario class? It seems to make sense that the user would always want the `base_grid` associated with the scenario they're creating

powersimdata/input/grid.py
```diff
@@ -29,44 +29,20 @@ class Grid:
...
+        self.sub = model_data.sub
+        self.plant = model_data.plant
+        self.gencost = model_data.gencost
+        self.dcline = model_data.dcline
+        self.bus2sub = model_data.bus2sub
+        self.bus = model_data.bus
+        self.branch = model_data.branch
+        self.storage = model_data.storage
 
         _cache.put(key, self)
```
goccert25, line 47: I'm curious, why was this `_cache` system put in and what problem did it solve?

goccert25:
Hey all, I wanted to open this PR early to get some eyes on the approach before committing 100% and trying to refactor everything. What do you all think of the approach? Where could it be improved? Would this negatively affect the current workflow of researchers using the simulator?

The primary goals of the refactor right now is:
- Improve code quality, usability, and readability
- Make it easier to add additional types of models (like TAMU), and many more of them

Are there any goals that should be considered as well?

powersimdata/input/grid.py
rouille: It can take some time to instantiate a USA `Grid` class so we save the object in memory. To illustrate, in **PostREISE**, we do analysis/plotting based on the scenario id and we often create a `Grid` object in these functions. Instead of creating a `Grid` object for each analysis/plots of the same scenario, we load it from memory.

powersimdata/design/investment/investment_costs.py
rouille: I guess it would not hurt to have it as an attribute of the `Grid` class. What do you thing @BainanXia and @danielolsen?

powersimdata/input/grid.py
rouille: I agree. What do you think @jon-hagg, @danielolsen, @BainanXia and @ahurli? I think that it could also solve the issue where a grid model (e.g., a single small country) has no interconnection.

jon-hagg: +1

danielolsen: I think every grid model should have at least one interconnection.

powersimdata/design/investment/investment_costs.py
danielolsen: The point of the Grid caching is so that the repeated initialization doesn't cost much extra time. Initializing another grid object every time we load a Scenario seems like it would hurt performance, not help it.

powersimdata/input/grid.py
jon-hagg: One place it is useful is running tests - we end up using one of the base grids multiple times, but since it's the same each time (not caching transformed grids here), we can reuse a copy from memory. If I recall this led to a slight improvement in the time it takes to run the test suite.

powersimdata/design/investment/investment_costs.py
jon-hagg: I imagine we are talking about adding this to the scenario class:
```
def get_base_grid(self):
    return Grid(self.info["interconnect"].split("_"), source=self.info["grid_model"])
```

where the result would be cached internally. However if the `Grid` constructor takes in `data` instead, we'd have to move the cache elsewhere, maybe a "factory" that essentially does the if/else logic currently in the grid class.

BainanXia: There are two things we are discussing here:
- Having a grid attribute `self.grid = Grid(self.info["interconnect"].split("_"), source=self.info["grid_model"])` in the scenario class, so that we could call `scenario.grid` instead of `scenario.get_grid()`.
- Having a base grid attribute or a method that returns the base grid in scenario class.

For the first one, we could evaluate whether there are situations that we load a scenario object without querying `get_grid()` method. However, it might add more overheads when load a scenario, which is probably okay I presume.

For the second one, I think it is obvious that we don't need the base grid for many situations when loading a scenario object. I would not set the base grid as an attribute and I'm with @danielolsen at this point, that's the point of caching.

powersimdata/input/grid.py
BainanXia: An interconnection defines a power grid that is connected by ac branches synchronized with same frequency when solving OPF. If every bus is connected by a path with only ac lines with every other bus in the network, it is one interconnection, no matter it is big or small.

rouille: So an interconnection can be a country name then?

danielolsen: There's no reason why not.

rouille: I thought it would be weir to have, e.g., a `Australia` class that we would instantiate through `Australia("Australia")` (I am sure Australia has several interconnections and it is a bad example). In that case I would rather do `Australia()`

BainanXia: Should it be model name instead, something like `Good_Australia_Model("Australia")`, just like `TAMU("USA")`?

goccert25: > I think every grid model should have at least one interconnection.
@danielolsen just to be clear, this is in response to @rouille's statement "I think that it could also solve the issue where a grid model (e.g., a single small country) has no interconnection.", or was it for/against the proposal I made?

powersimdata/input/grid.py
goccert25: What are your opinions on pulling the cache out of the Grid class? This would:
- make the pulling out of the initialization of the `data` variable easier
- make the code simpler - the idea of a Cache class that manages Grid objects (or any other objects) makes more sense than a Grid class that holds an internal cache imo
Cons:
- I can see user experience going down. I imagine data scientists who work with Grid everyday in jupyter notebooks don't want to have to also remember to import a Cache class to put Grid into. I could see this potentially being alleviated by providing helper functions such as `cache.get_or_create(Grid, TAMU, ["Western"))` but it could potentially mess with people's workflows
- Finding an intuitive key for a Cache that is pulled out of the Grid class seems difficult to say the least and eludes me currently

powersimdata/input/grid.py
goccert25: Also, just to verify my understanding, it's `TAMU(interconnect)`, `FromREISE(source)`, and `FromREISEjl(source)` that are the very expensive time consuming function calls right?

danielolsen: @goccert25 yes, I was referring to @rouille's comment.

`TAMU(interconnect)` involves reading and parsing a few hundred thousand lines of CSVs, and `FromREISE(source)` (almost never used anymore) and `FromREISEjl(source)` involve reading and parsing a matfile, and for these I think the lengthiest part of the operation is the `powersimdata.input.scenario_grid.format_gencost` function.

danielolsen: @rouille from the user side, we would still have `Grid(interconnect="Australia", source="Good_Australia_Model")`, right? If we define the `Good_Australia_Model` class appropriately, we should be able to get the hypothetical single interconnect via `Good_Australia_Model()` I think. I guess we could make `interconnect` an optional parameter in some cases, and then it would be possible to call `Grid(source="Good_Australia_Model")`.

rouille: Yeah. My comment was related to the new instantiation that is discussed here: `Grid(Australia("Australia"))`. If model = country name = interconnection, it seems better to do `Grid(Australia())` or even `Grid(Australia())` if there is only one interconnection and the grid model is explicit.

danielolsen: Ah, I see now, I got a little lost in the conversation.

I guess if we assume that the only grid model that can be instantiated with just a interconnect name (or a list of them) and nothing else is the `usa_tamu` model, then we can take a single parameter for a `Grid` init: which can be either an `AbstractGrid` instance (i.e. `TAMU("Eastern")` or `OTHERMODEL(interconnect)` or `FromREISEjl(filepath)`) or a list/str, which will get passed to a `TAMU` object, so `Grid("Eastern")` will become `Grid(TAMU("Eastern"))`, etc.

powersimdata/design/investment/investment_costs.py
```diff
@@ -381,7 +382,7 @@ def calculate_gen_inv_costs(
     grid = scenario.state.get_grid()
     if base_grid is None:
         base_grid = Grid(
-            scenario.info["interconnect"].split("_"), source=scenario.info["grid_model"]
+            model_data=AbstractGridFactory.get_or_create(scenario.info["grid_model"], scenario.info["interconnect"].split("_"))
```
jon-hagg, line 385: I still think it wouldn't hurt to have a `get_base_grid` method in the scenario class to simplify usages like this.

jon-hagg: Unrelated - looks like we still need to handle the case where we pass a string for the grid model, right? In `get_or_create`, we call `__init__` directly on the argument. Same comment for passing a mat file path.

powersimdata/input/abstract_grid.py
```diff
@@ -43,3 +44,19 @@ def storage_template():
         "terminal_max": None,
     }
     return storage
+
+
+class AbstractGridFactory:
```
jon-hagg, line 49: Consider naming this simply `GridFactory`. Other than being more concise, I think the `Abstract` prefix implies the class itself is abstract, vs producing an `AbstractGrid`.

ToddG: +1 on renaming to `GridFactory` rather than `AbstractGridFactory`.

ahurli (commented):
It seems to me like there's 2 things going on in this PR:
1) Updating the calls in `design` to be able to support arbitrary grids (though I still see in some places that `TAMU` is still hard-coded in). I like Jon's suggestion of adding a `get_base_grid` method to scenarios so that we can remove grid initiation altogether from `design`. I think this might be something easier to do before refactoring the grid itself, and it would also make the grid refactor easier if we remove this interaction.
2) Adding a factory to the grid initiation to support arbitrary grid data. I definitely like the idea of adding a factory so that we have more flexibility in how we initiate it, but as the grid is such a central part of the framework, there's a lot to consider, especially if we're not maintaining backwards compatibility.

powersimdata/design/investment/investment_costs.py
goccert25: >Unrelated - looks like we still need to handle the case where we pass a string for the grid model, right? In get_or_create, we call __init__ directly on the argument. Same comment for passing a mat file path.

@jon-hagg In the proposed changes, it's no longer possible to pass a string for the grid model. Specifically, this is why `set_grid` was changed to take in an `AbstractGrid` object instead of a string (you'll notice the default value for `grid_model` is now `TAMU` instead of `"usa_tamu"`). The `scenario.info["grid_model"]` will then always be an `AbstractGrid` object

powersimdata/input/abstract_grid.py
goccert25: I'm not opposed to renaming, but I think `GridFactory` is a bit confusing as well, as in the current implementation `AbstractGridFactory` actually caches objects like `TAMU`, `FromREISE`, and `FromREISEjl`, which are all `AbstractGrid`'s, and doesn't actually cache the `Grid` class itself. What about `GridDataFactory`?

powersimdata/scenario/create.py
```diff
@@ -145,7 +148,7 @@ def set_builder(self, *args, **kwargs):
         )
         self.set_grid(*args, **kwargs)
 
-    def set_grid(self, grid_model="usa_tamu", interconnect="USA"):
+    def set_grid(self, grid_model=TAMU, interconnect="USA"):
```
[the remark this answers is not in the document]
goccert25: Good call out on UX. Code as a user interface isn't something I usually see, but it seems to be core to this project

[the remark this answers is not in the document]
goccert25: One thing I can think of is modifying `AbstractGridFactory` to take either a `class` as the first object or a `string`, and the `AbstractGridFactory` can have logic that chooses a `class` based off of the passed in `string`

powersimdata/scenario/create.py
```diff
@@ -356,9 +358,9 @@ def set_engine(self, engine):
 
         :param str engine: simulation engine
         """
-        possible = ["REISE.jl"]
+        possible = [FromREISEjl]
```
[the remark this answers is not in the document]
goccert25: I updated this to `FromREISEjl` because I changed the code here https://github.com/Breakthrough-Energy/PowerSimData/pull/527/files#diff-39ffb11520dc297abc3cccc10da29d4d4a09d27c9655de99a31635c19c054b32R73-R75 to take in `engine` as a class, so I consequently had to update this `set_engine` code to make the possible value a class. If people gave approval for this direction I would have dove further in and changed all downstream dependencies

goccert25:
@ahurli 

> It seems to me like there's 2 things going on in this PR:
> 
> 1. Updating the calls in design to be able to support arbitrary grids (though I still see in some places that TAMU is still hard-coded in). I like Jon's suggestion of adding a get_base_grid method to scenarios so that we can remove grid initiation altogether from design. I think this might be something easier to do before refactoring the grid itself, and it would also make the grid refactor easier if we remove this interaction.
> 2. Adding a factory to the grid initiation to support arbitrary grid data. I definitely like the idea of adding a factory so that we have more flexibility in how we initiate it, but as the grid is such a central part of the framework, there's a lot to consider, especially if we're not maintaining backwards compatibility.

Great call out. With respect to 1, I'm in support of adding a `get_base_grid` method and reducing copy and pasted code, but @BainanXia and @danielolsen both have opposing opinions. Ultimately I think you all are closer to the end user and have a better sense of what would be the best call in this case, so could I ask you @ahurli to get a decision that the team agrees upon, and I'd be happy to crack a PR for that.

With respect to 2, I'll take the feedback provided so far and see what I can come up with. I think what I might end up doing is what I commented previously, which is
>One thing I can think of is modifying AbstractGridFactory to take either a class as the first object or a string, and the AbstractGridFactory can have logic that chooses a class based off of the passed in string, thoughts?

BainanXia:
@goccert25 Reading through the conversations, I think we have been drifting from the original ideas:
- @danielolsen and I were opposing to having `grid`/`base_grid` as attributes of `Scenario` class, the reasons are explained [above](https://github.com/Breakthrough-Energy/PowerSimData/pull/527#discussion_r672680513). As for adding a method, `get_base_grid()`, so that all the initializations are encapsulated under the hood, I like it. 
- Regarding the second point that @ahurli raised above, we have discussed this in an internal meeting. We would like to hold off the refactor/implementation for now but keep the conversation open. The main reason is `Grid` class is one of the main user interfaces, which potentially involves major refactor all over the framework, but without salient immediate benefit at this point. I believe we will have a better idea about the design once we have another grid model except for `TAMU` ready to use in our software.

goccert25:
@BainanXia great, thanks for taking the time to drive a decision with respect to the refactoring, I'll close this PR then and crack a quick one for the `get_base_grid()` method

PR #528, opened 2021-08-07
refactor: add get_base_grid() to states

goccert25:

### Purpose
What is the larger goal of this change?
To pay down some tech debt and hopefully make PowerSimData codebase more maintainable

### What the code is doing
How is the purpose executed?
Removing many similar/copy and pasted calls to initialize the original Grid that a scenario was derived from via information from the scenario object

### Testing
Existing unit tests

### Time estimate
How long will it take for reviewers and observers to understand this code change?
20 minutes

### DIFF
--- a/powersimdata/design/investment/investment_costs.py
+++ b/powersimdata/design/investment/investment_costs.py
@@ -1,4 +1,3 @@
-import copy as cp
 import warnings
 
 import numpy as np
@@ -16,7 +15,6 @@ from powersimdata.design.investment.create_mapping_files import (
 )
 from powersimdata.design.investment.inflation import calculate_inflation
 from powersimdata.input.check import _check_grid_models_match
-from powersimdata.input.grid import Grid
 from powersimdata.utility.distance import haversine
 
 
@@ -65,24 +63,27 @@ def calculate_ac_inv_costs(
         float if ``sum_results``, or pandas Series indexed by branch ID.
         Whether summed or not, values are $USD, inflation-adjusted to today.
     """
-    grid = scenario.state.get_grid()
+    grid_differences = scenario.get_grid()
     if base_grid is None:
-        base_grid = Grid(
-            scenario.info["interconnect"].split("_"), source=scenario.info["grid_model"]
-        )
+        base_grid = scenario.get_base_grid()
     else:
-        _check_grid_models_match(base_grid, grid)
+        _check_grid_models_match(base_grid, grid_differences)
 
-    grid_new = cp.deepcopy(grid)
     # find upgraded AC lines
-    capacity_difference = calculate_branch_difference(base_grid.branch, grid.branch)
-    grid_new.branch = grid.branch.assign(rateA=capacity_difference.diff)
-    grid_new.branch = grid_new.branch.query("rateA != 0.0")
+    capacity_difference = calculate_branch_difference(
+        base_grid.branch, grid_differences.branch
+    )
+    grid_differences.branch = grid_differences.branch.assign(
+        rateA=capacity_difference.diff
+    )
+    grid_differences.branch = grid_differences.branch.query("rateA != 0.0")
     if exclude_branches is not None:
-        present_exclude_branches = set(exclude_branches) & set(grid_new.branch.index)
-        grid_new.branch.drop(index=present_exclude_branches, inplace=True)
+        present_exclude_branches = set(exclude_branches) & set(
+            grid_differences.branch.index
+        )
+        grid_differences.branch.drop(index=present_exclude_branches, inplace=True)
 
-    costs = _calculate_ac_inv_costs(grid_new, sum_results)
+    costs = _calculate_ac_inv_costs(grid_differences, sum_results)
     return costs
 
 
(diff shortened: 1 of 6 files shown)


powersimdata/design/investment/investment_costs.py
```diff
@@ -50,39 +48,29 @@ def calculate_ac_inv_costs(
...
     """
     grid = scenario.state.get_grid()
-    if base_grid is None:
-        base_grid = Grid(
-            scenario.info["interconnect"].split("_"), source=scenario.info["grid_model"]
-        )
-    else:
-        _check_grid_models_match(base_grid, grid)
-
-    grid_new = cp.deepcopy(grid)
```
goccert25, line 76: Note, I removed this deepcopy because looking at the `get_grid` functions, there's already a deepcopy being made, so this call seemed useless, correct me if I'm wrong though

jon-hagg:
Nice, the implementation looks good to me. One other thing I would consider is updating the usages of `scenario.state.get_base_grid()` to be simply `scenario.get_base_grid()`.

powersimdata/design/investment/investment_costs.py
```diff
@@ -50,39 +48,29 @@ def calculate_ac_inv_costs(
     scenario,
     sum_results=True,
     exclude_branches=None,
-    base_grid=None,
```
BainanXia, line 53: I think we should keep this. The `base_grid` here is any grid that the user would like to compare against, which is different from the base grid we have been talking about, i.e. a fresh instance of the `Grid` class without modifications by change table.

powersimdata/design/investment/investment_costs.py
```diff
@@ -50,39 +48,29 @@ def calculate_ac_inv_costs(
...
-        corresponding unmodified Grid.
     :return: (*dict*) -- keys are {'line_cost', 'transformer_cost'}, values are either
         float if ``sum_results``, or pandas Series indexed by branch ID.
         Whether summed or not, values are $USD, inflation-adjusted to today.
     """
     grid = scenario.state.get_grid()
-    if base_grid is None:
-        base_grid = Grid(
-            scenario.info["interconnect"].split("_"), source=scenario.info["grid_model"]
-        )
```
BainanXia, line 72: The logic `if base_grid is None` can be replaced by `base_grid = scenario.get_base_grid()`

powersimdata/design/investment/investment_costs.py
BainanXia: I agree. Then we don't need l.64: `grid = cp.deepcopy(grid)`, do we?

powersimdata/design/investment/investment_costs.py
```diff
@@ -272,34 +260,24 @@ def get_branch_mult(x, bus_reg, ac_reg_mult, branch_lookup_alerted=set()):
         return {"line_cost": lines.cost, "transformer_cost": transformers.cost}
 
 
-def calculate_dc_inv_costs(scenario, sum_results=True, base_grid=None):
```
BainanXia, line 275: Same comment as above.

powersimdata/design/investment/investment_costs.py
```diff
@@ -354,7 +332,6 @@ def calculate_gen_inv_costs(
     year,
     cost_case,
     sum_results=True,
-    base_grid=None,
```
BainanXia, line 357: Same here.

powersimdata/design/transmission/mwmiles.py
```diff
@@ -15,7 +14,7 @@ def calculate_mw_miles(scenario, exclude_branches=None):
     :return: (*dict*) -- Upgrades to the branches.
     """
 
-    original_grid = Grid(scenario.info["interconnect"].split("_"))
```
BainanXia, line 18: Nice catch.

powersimdata/design/investment/investment_costs.py
goccert25: good catch. done

powersimdata/design/transmission/mwmiles.py
goccert25: :)

powersimdata/design/investment/investment_costs.py
goccert25: done

powersimdata/design/investment/investment_costs.py
goccert25: done

goccert25:
>Nice, the implementation looks good to me. One other thing I would consider is updating the usages of scenario.state.get_base_grid() to be simply scenario.get_base_grid().

@jon-hagg I used `scenario.state.get_base_grid()` because in the code there was already `scenario.state.get_grid()` throughout. I'm happy with switching over to `scenario.get_base_grid()` but if so I would ask if I could also do `scenario.get_grid()` to be consistent. Let me know!

danielolsen:
The code is clear, and it simplifies the logic within the `powersimdata.design` module. I support a general change from `scenario.state.foo` to `scenario.foo`, which is available since #503. Previously we simplified the user-facing example code but kept the more verbose calls within the internal code so that the tests would still pass with the mock objects, but now we should be able to simplify everywhere.

One comment is that with the variable renaming from `grid_new` to `grid` within the investment cost functions, we should be sure to remember that what's in this object named `grid` is not really a grid in the usual sense of the word, but a grid _difference_ for one table within the grid (e.g. the `branch` table or the `plant` table or the `dcline` table), which we use a Grid object for because it's convenient to have those other Grid attributes around (i.e. the `bus` table, `sub` table, other non-table attributes, etc.) within the investment cost calculation functions. One day we might decide that we want a cleaner implementation, e.g. maybe with a new `GridDelta` class that could have one or more tables that are really tables of element _differences_, not of elements themselves, but I think that's outside of the scope of this PR.

jon-hagg:
@goccert25 good call - I would also support switching to `scenario.get_grid()`.

goccert25:
@jon-hagg done

@danielolsen renamed the `grid` variable to `grid_differences` to be more appropriate

Feel free to approve and merge whenever, I rebased on develop as of the time of this comment

jon-hagg (approved):
LGTM

danielolsen (approved):
Looks good, thanks!

PR #530, opened 2021-08-17
chore: use setup.cfg

jon-hagg:
### Purpose
Switch to setup.cfg for package metadata since it's the [recommended](https://packaging.python.org/tutorials/packaging-projects/#configuring-metadata) approach. 

### What the code is doing
The details are the same as in setup.py, but this also adds some classifiers, project urls, and sets the readme as the `long_description` which will appear on the PyPI page once we publish.

### Testing
First ran `pip install build`, followed by `python -m build`, which creates a `dist/` folder with `.whl` package file. I installed this in a new environment and checked that I can still import stuff and instantiate a grid (to verify the `package_data` is included).

Also published to the test pypi server, check it out [here](https://test.pypi.org/project/powersimdata/)

### Time estimate
5 min

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,38 @@
+[metadata]
+name = powersimdata
+version = 0.4.3
+author = Kaspar Mueller
+author_email = kaspar@breakthroughenergy.org
+description = Power Simulation Data
+long_description = file: README.md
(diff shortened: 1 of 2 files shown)


danielolsen (approved):
LGTM

PR #531, opened 2021-08-18
ci: add workflow to build and upload package

jon-hagg:
### Purpose
New workflow to handle building and uploading the package to PyPI whenever we push a new git tag with the specified version scheme. It can also be run manually, so we can test it out on the existing `v0.4.3` release.

### What the code is doing
Usual workflow stuff. We install the `build` package to create the `dist/` folder which will be uploaded subsequently using the github action provided by pypa.

### Testing
Tested the release to [TestPyPI](https://test.pypi.org/project/powersimdata/) initially by omitting the tags filter in the `on:` section. Once this is merged I will run the workflow on our latest release, which should also upload to the real PyPI.

### Time estimate
10 min

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,28 @@
+name: Publish Python 🐍 package 📦 to PyPI and TestPyPI
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - v*.*.*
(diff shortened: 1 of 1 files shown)


kasparm (approved):
LGTM thanks.

PR #537, opened 2021-09-01
ci: combine workflow triggers

jon-hagg:
### Purpose
Use the generic workflow trigger action to kick off the docs build and docker build in the respective repositories. Related issue https://github.com/Breakthrough-Energy/plug/issues/26

### What the code is doing
Removed obsolete docker build action, replaced it, and the existing docs trigger, with the workflow-trigger action.

### Testing
Tested the action manually

### Time estimate
5 min


jon-hagg:
Note: I've manually disabled the 'Publish docker image' workflow and removed the `CR_PAT` secret ahead of time, since they will be obsolete when this is merged.

PR #543, opened 2021-09-07
testing, please ignore

jon-hagg:


jon-hagg:
Nice - looks like a PR from a fork will use the fork's workflows, including any configured secrets (rather than secrets from the target repo). So unless the fork has a valid secret, the checks will fail (which is what we want).

PR #546, opened 2021-09-14
fix: correct generation of duplicated bus entries in bus-to-NEEM-region mapping

### LINKED ISSUE #545: Duplicate entries in buses_NEEMregion.csv can cause ValueError when calculating AC investment costs
# :beetle:

- [x] I have checked that this issue has not already been reported.

### Bug summary
For _some_ scenarios, if the upgrades branches touch a bus for which there are duplicate entries in the **powersimdata/design/investment/databuses_NEEMregion.csv** file, a ValueError will be thrown. This is because the duplicated index causes the lookup to grab more values than it should, and therefore we're trying to insert more data than the data frame has room for.

### Code for reproduction
A minimum code snippet required to reproduce the bug. Please make sure to minimize the
number of dependencies required.
```python
from powersimdata.design.investment.investment_costs import calculate_ac_inv_costs
scenario = Scenario(3287)
calculate_ac_inv_costs(scenario)
```

### Actual outcome
The output produced by the above code, which may be a screenshot, console output, etc.
```shell
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\design\investment\investment_costs.py", line 86, in calculate_ac_inv_costs
    costs = _calculate_ac_inv_costs(grid_differences, sum_results)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\design\investment\investment_costs.py", line 211, in _calculate_ac_inv_costs
    branch.loc[:, "to_region"] = bus_reg.loc[branch.to_bus_id, "name_abbr"].tolist()
  File "C:\Python39\lib\site-packages\pandas\core\indexing.py", line 670, in __setitem__
    iloc._setitem_with_indexer(indexer, value)
  File "C:\Python39\lib\site-packages\pandas\core\indexing.py", line 1601, in _setitem_with_indexer
    self._setitem_with_indexer(new_indexer, value)
  File "C:\Python39\lib\site-packages\pandas\core\indexing.py", line 1666, in _setitem_with_indexer
    raise ValueError(
ValueError: cannot set using a multi-index selection indexer with a different length than the value
```

### Expected outcome
No error.

### Environment
Please specify your platform and versions of the relevant libraries you are using:
* Operating system: Windows
* PowerSimData revision: ab21118359fadbb6c4150c50c17e7a34b26a329b
* Python version: 3.9.6

### Additional context
We can see the root cause directly by inspecting the CSV as it is loaded within the investment cost module:
```python
>>> import pandas as pd
>>> from powersimdata.design.investment import const
>>> bus_reg = pd.read_csv(const.bus_neem_regions_path, index_col="bus_id")
>>> bus_reg.loc[bus_reg.index.duplicated(False)]
             name_abbr  dist      lat      lon
bus_id
12359          PJM ROM   0.0  39.9515 -75.8260
12359            PJM E   0.0  39.9515 -75.8260
12459            PJM E   0.0  39.9024 -75.8392
12459          PJM ROM   0.0  39.9024 -75.8392
12460          PJM ROM   0.0  39.9024 -75.8392
12460            PJM E   0.0  39.9024 -75.8392
36055   NonRTO Midwest   0.0  39.6566 -83.5378
36055          PJM ROR   0.0  39.6566 -83.5378
42494   NonRTO Midwest   0.0  37.9335 -87.5617
42494          MISO IN   0.0  37.9335 -87.5617
43085        MISO WUMS   0.0  42.4947 -88.6446
43085          PJM ROR   0.0  42.4947 -88.6446
47988        MISO WUMS   0.0  42.5008 -89.1299
47988          PJM ROR   0.0  42.5008 -89.1299
47989        MISO WUMS   0.0  42.5008 -89.1299
47989          PJM ROR   0.0  42.5008 -89.1299
48134        MISO WUMS   0.0  42.4960 -88.5134
48134          PJM ROR   0.0  42.4960 -88.5134
```

The code had not minded before #450, since it would grab however many entries for both buses and average them naively.

Suggested solution: fix `powersimdata.design.investment.create_mapping_files.write_bus_neem_map` to never cause this issue (either fix the bug that causes one bug to map to multiple regions, or select one somehow and toss the other), and re-generate the **buses_NEEMregion.csv** file.

danielolsen:

### Purpose
Closes #545.

### What the code is doing
- Within **create_mapping_files.py**, after we detect bus coordinates that lay within region polygons, we identify duplicated buses. For each, we find the closest neighboring bus within non-duplicated buses. We assume that that is the best guess for the 'true' region of the duplicated bus, and mark any other regions to be dropped.
- There is also an unrelated fix for the code to make directories.
- We update the CSV as a result of running the new code.

### Testing
Tested manually to generate a new CSV:
```python
from powersimdata import Grid
from powersimdata.design.investment.create_mapping_files import write_bus_neem_map
grid = Grid("USA")
write_bus_neem_map(grid)
```

Confirming that the new CSV has no duplicates:
```python
>>> from powersimdata.design.investment import const
>>> mapping = pd.read_csv(const.bus_neem_regions_path)
>>> mapping.index.duplicated().sum()
0
```

We can also see that the newly-generated CSV now has exactly nine fewer rows, and no other changes, which follows from the nine duplicate pairs detected earlier.

Confirming that the Scenario that triggered the bug before no longer does:
```python
>>> from powersimdata.design.investment.investment_costs import calculate_ac_inv_costs
>>> scenario = Scenario(3287)
>>> calculate_ac_inv_costs(scenario)
{'line_cost': 387768866295.44775, 'transformer_cost': 9334839565.245375}
```

### Time estimate
15-30 minutes.


powersimdata/design/investment/create_mapping_files.py
```diff
@@ -67,6 +68,29 @@ def _find_nearest(series, polygons, search_dist):
...
+        entries = duplicated.query("bus_id == @bus")
+        coords = entries["geometry"].iloc[0].coords[0]  # First duped entry, only point
+        regions = set(entries["name_abbr"])  # noqa: F841
+        candidates = points_in_regions.query(
+            "index not in @duplicated.index and name_abbr in @regions"
+        )
+        neighbor = candidates.apply(
+            lambda x: haversine((x.geometry.x, x.geometry.y), coords), axis=1
+        ).idxmin()
+        closest_region = candidates.loc[neighbor, "name_abbr"]  # noqa: F841
```
BainanXia, line 84: If I'm understanding correctly, the logic here builds on the assumption that the `closest_region` must be one of the regions for the duplicated bus. Just curious what if that's not the case, i.e. we have two entries as following
```
bus_1 region A
bus_1 region B
```
and the closest neighbor, `bus_2` is in `region C`, i.e. no other buses in `region A` and `region B`, what should we do?

If we are safe in that situation, i.e. that will never happen. Would it be easier to simply drop all the duplications and reassign the regions of those entries by looking up the closest neighbor region directly instead of figuring out which regions should be dropped (as implemented currently)?

danielolsen: My initial thinking of only looking at buses within the overlapping regions was to limit computation time. However, since the number of duplicate is fairly small, and this process is done only infrequently, computation time is probably not the main consideration.

Thinking a bit more, I could definitely draw a picture where a bus lands within two overlapping region shapes, but the closest neighbor bus is within a third region. I think the chances of this are fairly low though, since the bus would have to be within an intersection of two regions (0.01% incidence in this data set) and also in close vicinity to a three-region border. The assumption that one of the regions is correct seems reasonable enough to me.

BainanXia: Yeah, regarding the second point, three region overlapping is one case. What I was thinking about previously is for buses within two overlapping regions, what if there is no bus within `S(region_a) - S(region_b)` or `S(region_b) - S(region_a)`, in such situation, we only have buses within the overlapping zone for the two regions, i.e. candidates list is empty in l.78. However, I agree, certain possibility is low and probably we can ignore that.

Given the assumption that we can always find a nearest neighbor in either region for any bus within overlapping zones, what I was proposing is (still only looking at those duplicated buses) we can simply reassign region for those buses based on the nearest neighbor instead of figuring out which region should drop, i.e. 
```python
duplicated_flag = points_in_regions.index.duplicated(keep=False)
non_duplicated = points_in_regions.loc[~duplicated_flag]
duplicated = points_in_regions.loc[~points_in_regions.index.duplicated(keep='first')]
# simply look at all regions with duplicated buses here, sacrifice some computational time here comparing with only look at duplicated regions for corresponding buses
regions = points_in_regions.loc[duplicated_flag, "name_abbr"].unique()
candidates = non_duplicated.query("name_abbr in @regions")
duplicated["name_abbr"] = duplicated.apply(lambda x: candidates.iloc[find_closest_neighbor(x.geometry, candidates.geometry)]["name_abbr"])
points_in_regions = non_duplicated.append(duplicated)
```

BainanXia: Quick comment on this resolved thread: I think we are good to go with the current implementations, just clarifying the corner cases that we don't cover and providing an alternative of implementation here.

BainanXia (approved):
Good catch.

PR #553, opened 2021-09-24
Remove psycopg2 dependency

jon-hagg:
### Purpose
Reduce the installation overhead and avoid related errors due compilation, etc.

### What the code is doing
We aren't using this package except in tests, and don't really need to be running those right now - keeping them around in case they are useful when we switch to sqlite or alternate data store. For now they are moved to a sql folder and we tell pytest to ignore anything in there. As a result, we can simplify the ci workflow to just use `pytest-local`.

### Testing
Ran `tox -e pytest-local` and also built the docs locally - as expected there is an error generating docs for modules that import psycopg2, but those are simply omitted, and the rest of the docs look fine.

### Time estimate
5 min


kasparm (approved):
That works and looks good to me.
Thanks @jon-hagg

PR #554, opened 2021-09-28
feat: allow the user to remove lines, buses, plants, and DC lines

### LINKED ISSUE #423: Add ability to remove grid elements from change table
We can already scale generators Pmax and Pmin down to 0 to turn them off and effectively remove them, but AFAIK we can't remove branches. This may be something we want to do in the future, so that we could do things like replacing a low-voltage line with a higher-voltage one. It would also be nice if generators can be actually removed, rather than just scaled down to 0.

danielolsen:

### Purpose
~Partial fulfilment of~ Closes #423. Plants and DC lines can already be _effectively_ removed by scaling them down to zero, this allows branches and buses to be removed as well. If we like the structure of the branch and bus removal, and want to enable full removal of plants, DC lines, storage, etc., then I can amend this PR to include those analogous methods as well.

### What the code is doing
Within the `ChangeTable` class:
- We add new methods `remove_bus` and `remove_branch`, which check that the elements to be removed are actually in the grid before adding to the change table dictionary. `remove_bus` has a check to ensure that a bus cannot be removed if there's at least one plant with non-zero capacity at it, and to implement this we modify `_get_df_with_new_elements` so that if we request the `plant` table, we apply any plant scaling.
- We add a user notification method, to ensure that the user is aware if they've accidentally islanded some load buses. This will not break the simulation, but load shedding will be enabled which could slow simulation.

Within the `TransformGrid` class:
- We interpret the `"remove_bus"` and `"remove_branch"` change table keys.

### Testing
New unit tests added.

### Time estimate
15-30 minutes.

### DIFF
--- a/powersimdata/input/change_table.py
+++ b/powersimdata/input/change_table.py
@@ -1,6 +1,7 @@
 import copy
 import os
 import pickle
+from itertools import chain
 
 from powersimdata.design.transmission.upgrade import (
     scale_congested_mesh_branches,
@@ -124,6 +125,14 @@ class ChangeTable:
         keys in the dictionary are: *'lat'*, *'lon'*, one of *'zone_id'*/*'zone_name'*,
         and optionally *'Pd'*, specifying the location of the bus, the demand zone, and
         optionally the nominal demand at that bus (defaults to 0).
+    * *'remove_branch'*:
+        value is a set. Each entry in this set is a branch ID to be removed.
+    * *'remove_bus'*:
+        value is a set. Each entry in this set is a bus ID to be removed.
+    * *'remove_dcline'*:
+        value is a set. Each entry in this set is a DC line ID to be removed.
+    * *'remove_plant'*:
+        value is a set. Each entry in this set is a plant ID to be removed.
     """
 
     def __init__(self, grid):
@@ -133,7 +142,9 @@ class ChangeTable:
         """
         self.grid = grid
         self.ct = {}
-        self._new_element_caches = {k: {} for k in {"branch", "bus", "dcline", "plant"}}
+        self._new_element_caches = {
+            k: {} for k in {"branch", "bus", "dcline", "plant", "storage_gen"}
+        }
 
     @staticmethod
     def _check_resource(resource):
@@ -199,7 +210,7 @@ class ChangeTable:
             self.ct.clear()
             return
         # Input validation
-        allowed = {"branch", "dcline", "demand", "plant", "storage"}
+        allowed = {"branch", "bus", "dcline", "demand", "plant", "storage"}
         if isinstance(which, str):
             which = {which}
         if not isinstance(which, set):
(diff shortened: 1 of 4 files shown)


powersimdata/input/change_table.py
```diff
@@ -789,17 +794,92 @@ def _get_df_with_new_elements(self, table):
...
+                    f"new_{table}",
+                    tuple(tuple(sorted(b.items())) for b in self.ct.get(table, {})),
+                ]
+                + [
+                    tuple([subkey] + sorted(subdict.items()))
+                    for k in modification_keys[1:]
+                    for subkey, subdict in self.ct.get(k, {}).items()
+                ]
+            )
+        )
```
rouille, line 814: you lost me here. What if `modification` is a single element, i.e., `table` is not `plant`? Won't `modification_keys[1:]` raise an error?

danielolsen: No, we'll just get an empty list:
```python
>>> foo = ["a"]
>>> foo[1:]
[]
```

The current implementation is clunky for sure. If `table == "plant"`, then we want to reliably create a tuple that includes all plant additions and plant scaling information. If `table != "plant"`, we should be creating a tuple as before, just with one new top-level string entry for `"new_bus"` or `"new_branch"` or whatever.

powersimdata/input/change_table.py
```diff
@@ -789,17 +794,92 @@ def _get_df_with_new_elements(self, table):
...
+        if not any(m in self.ct for m in modification_keys):
             return getattr(self.grid, table)
-        new_elements_tuple = tuple(tuple(sorted(b.items())) for b in self.ct[add_key])
-        if new_elements_tuple in self._new_element_caches[table]:
-            return self._new_element_caches[table][new_elements_tuple]
+        modified_elements_tuple = tuple(
+            chain.from_iterable(
+                [
+                    f"new_{table}",
+                    tuple(tuple(sorted(b.items())) for b in self.ct.get(table, {})),
```
jon-hagg, line 806: We need the extra `tuple`? I know it's already there, just curious what the data looks like.. kinda thought it was idempotent.

danielolsen: I think the outermost `tuple` in this line is needed, otherwise we end up with a generator instead of something that can be hashed (unless you're talking about the `tuple` call wrapping `chain.from_iterable`).

I realize how clunky this block is, if you come up with something cleaner I'll happily take it. The test case that I added should activate this block to validate any refactoring attempts.

jon-hagg: Makes sense, I misread the parenthesis..

powersimdata/input/change_table.py
```diff
@@ -124,6 +125,10 @@ class ChangeTable:
         keys in the dictionary are: *'lat'*, *'lon'*, one of *'zone_id'*/*'zone_name'*,
         and optionally *'Pd'*, specifying the location of the bus, the demand zone, and
         optionally the nominal demand at that bus (defaults to 0).
+    * *'remove_branch'*:
```
rouille, line 128: do we want `remove_branch` (`remove_bus`) following `new_branch` (`new_bus`) or all `new_x` and `remove_x` altogether? I have no preference.

powersimdata/input/change_table.py
```diff
@@ -789,17 +794,92 @@ def _get_df_with_new_elements(self, table):
             'branch', 'bus', 'dcline', or 'plant'.
         :return: (*pandas.DataFrame*) -- the post-transformation table.
         """
```
rouille, line 796: It is kind of confusing that this function is named `_get_df_with_new_elements` now that we use it to not only add new objects but also remove some.

Issue #538, opened 2021-09-03
Consolidate Data Access Code


ahurli:
### Description
The proposed change is to refactor the code such that read/write operations are handled in a consolidated place (i.e. either input_data or data_access).

### Initiative / Goal (Business)
This will help us move fully to `pyfilesystem` as currently `pyfilesystem` is only used to move or delete existing files. By consolidating the read/write parts of data access, we should be able to then update these functions to use `pyfilesystem` more easily and thus be able to realize the full benefits of having a fully-abstracted file system (e.g. easier to switch between different filesystems, support for multiple data sources and/or scenario lists, etc).

### Success (Optional)
Using `pyfilesystem` for all file system operations.

### Acceptance criteria
Remove direct read/writes of data to local file system (i.e. pandas, pickle, or scipy.io) and consolidate path building from anywhere outside of data access/input data.

- [x] Define expected read/write functionality outside of powersimdata workflow (e.g. export_case_mat in powersimdata/input/export_data was refactored out so that ext. collaborators could manually save a case.mat to examine/test without starting a simulation). How and where do we need to expose these functions, both for now and in the future?
- [x] Decide design for where the new file objects will be created and what their lifecycle looks like
- [x] Update direct-to-local-filesystem read/writes of data (i.e. pandas, pickle, or scipy.io)
    - [ ] input/change_table.py: pickle dump
    - [ ] input/export_data.py: savemat, pandas csv
    - [ ] data_access/csv_store.py: pandas csv
    - [ ] output/output_data.py: pandas pickle
- [ ] Consolidate path building from anywhere outside of data access/input_data (in addition to above)
    - [ ] create.py
    - [ ] execute.py
    - [ ] analyze.py
    - [ ] (bonus) move.py
    - [ ] (bonus) delete.py

[Dropbox Paper with more in-depth notes about this epic](https://www.dropbox.com/scl/fi/uwynx5md5ut6ks0od9dsu/Data-Consolidation.paper?dl=0&rlkey=cc5ct4harvc0b2qw8jiloyfw7)

### Owner
@ahurli 

### Support
@rouille @jon-hagg @BainanXia @danielolsen

PR #555, opened 2021-09-29
Feat: Add Read to Data_Access

ahurli:

### Purpose
Adding a `read` method to the `Data_Access` class to be used later by later refactors as part of epic #538.

### What the code is doing
Added code to `data_access.py` to read in .pkl, .csv, and .mat files. It is roughly the same as the existing code in `_read_data` in `input_data.py`, except it is now using pyfilesystem and returns the mat file contents instead of just returning a filepath.

### Testing
I tested both `LocalDataAccess` to read from my local machine and `SSHDataAccess` from the dev server. 

### Where to look
`data_access.py`

### Usage Example/Visuals
from powersimdata.data_access.data_access import LocalDataAccess
test_fs = LocalDataAccess('~/ScenarioData')
test_fs.read('test.pkl')
test_fs.read('test.csv')
test_fs.read('test.mat')

### Time estimate
5 min


powersimdata/data_access/data_access.py
```diff
@@ -191,6 +217,7 @@ def __init__(self, root=server_setup.DATA_ROOT_DIR):
         """Constructor"""
         super().__init__(root)
         self._fs = None
+        self._base_fs = None
```
rouille, line 220: Will we need this attribute inn the future?

powersimdata/data_access/data_access.py
```diff
@@ -27,6 +30,29 @@ def __init__(self, root):
         self.root = root
         self.join = fs2.path.join
 
+    def read(self, filepath):
+        """Reads data from local machine.
```
rouille, line 34: I guess this could be updated to `Read data from data store`

rouille:
I am just wondering if the `read` method should not be defined into the `LocalDataAccess` class. Currently, we always read profiles, change table and grid MAT-file located in the user space. We need to fetch the data first from other data store but the reading is done locally, no?

powersimdata/data_access/tests/test_data_access.py
```diff
@@ -1,3 +1,5 @@
+import pickle
```
rouille, line 1: Unused (see `flake8` error)

powersimdata/data_access/data_access.py
```diff
@@ -27,6 +30,29 @@ def __init__(self, root):
...
+        :return: (*pandas.DataFrame* or *dict*) -- pkl and csv files will be returned as
+            a data frame, while a mat file will be returned as a dictionary
+        :raises ValueError: if extension is unknown.
+        """
+        ext = os.path.basename(filepath).split(".")[-1]
+        if ext == "pkl":
+            file_object = self.fs.open(filepath, mode="rb")
+            data = pd.read_pickle(file_object)
+        elif ext == "csv":
+            file_object = self.fs.open(filepath)
```
jon-hagg, line 46: Any reason we don't use `mode="rb"` here as well?

jon-hagg: To be safe, I would close the `file_object` when we're done with it. This could be just before the return, or if we can standardize the mode, we could move the call outside the if/else and use a `with` statement.

powersimdata/data_access/data_access.py
```diff
@@ -27,6 +30,29 @@ def __init__(self, root):
         self.root = root
         self.join = fs2.path.join
 
+    def read(self, filepath):
+        """Reads data from local machine.
+
+        :param str filepath: path to file, with extension either 'pkl', 'csv', or 'mat'.
+        :return: (*pandas.DataFrame* or *dict*) -- pkl and csv files will be returned as
+            a data frame, while a mat file will be returned as a dictionary
```
jon-hagg, line 38: I'm not 100% sure about the return type - e.g. if we load a change table from .pkl, would we return a `ChangeTable` instance? I'd say if there is no static guarantee on the type, best we could do is `object` (with some description about the possible values).

danielolsen: Unpickling a change table returns a `dict`.

powersimdata/data_access/data_access.py
ahurli: Nope! Thanks for catching that.

powersimdata/data_access/data_access.py
ahurli: Good call! I had to use `rb` mode for the `mat` file, and didn't realize it would work for `csv` as well. Updated.

ahurli:
Keeping `read` in the top-level `DataAccess` class gives us the ability to use it for both local and ssh implementations, and it will also mean it should work for other data access instances like an `AzureBlob` data access when that's ready. Even for existing methods that are only in one type but not the other (i.e. `execute_command_async` exists but is not implemented in the parent `DataAccess` class and only implemented in the `SSHDataAccss`), it's still defined at the top level.

danielolsen:
Do we expect an `SSHDataAccess` object to ever want to do a direct `read`? So far we've always been downloading to a local cache and then doing local file operations from there, or writing with a local file operation and then using the `SSHDataAccess` only to upload. Would we follow the same pattern when we're interacting with blob storage?

rouille:
> Keeping `read` in the top-level `DataAccess` class gives us the ability to use it for both local and ssh implementations, and it will also mean it should work for other data access instances like an `AzureBlob` data access when that's ready. Even for existing methods that are only in one type but not the other (i.e. `execute_command_async` exists but is not implemented in the parent `DataAccess` class and only implemented in the `SSHDataAccss`), it's still defined at the top level.

My thinking was declaring the method in the parent class with a `NotImplentedError` or a `pass` and defining the function within the `LocalDataAccess`. I think we will always read in the userspace (`LocalDataAccess`) that could be laptop or a remote server. I believe it will never be blob storage since Python is not installable there.

BainanXia:
> Do we expect an `SSHDataAccess` object to ever want to do a direct `read`? So far we've always been downloading to a local cache and then doing local file operations from there, or writing with a local file operation and then using the `SSHDataAccess` only to upload. Would we follow the same pattern when we're interacting with blob storage?

Reading through this track, I'm thinking about this same thing. Right now, downloading/uploading (could interact with remote file system) and reading/writing (local file operations) are separated, which means we will never do a direct reading/writing on the remote file system, unless we refactor the logic to have reading/writing be the universal interface to interact with any file system, both local and remote. In that case, `read` method within `SSHDataAccess` will need to take care of the whole process, `check_cache` -> `copy_from` -> `cache` -> `read`, whereas in `LocalDataAccess`, the `read` method will be simply as implemented here.

ahurli:
I was thinking that the logic for not writing to a remote data store needs to be contained at the `Input` or `OutputData` layer anyway, and as this solution should work for any pyfilesystem, I'm not sure it makes sense to prematurely restrict it to just `LocalDataAccess`.

Ultimately, I want to move towards Bainan's suggestion of having something like a `SSHDataAccess` with a configured local cache pyfilesystem, but that's definitely outside the scope of this current work. In that case, I would want to overwrite the default `read` and `write` operations to use the local caching once that's added.

jon-hagg:
I think this is good to go after amending the commit message

PR #556, opened 2021-10-02
feat: add write feature to Data Access

ahurli:

### Purpose
Adding a `write` method to the `Data_Access` class to be used later by later refactors as part of epic #538.

### What the code is doing
Added code to `data_access.py` to write out .pkl, .csv, and .mat files.

### Testing
I tested both `LocalDataAccess` to write to my local machine and `SSHDataAccess` to the dev server. 

### Where to look
`data_access.py`

### Usage Example/Visuals
from powersimdata.data_access.data_access import LocalDataAccess
test_fs = LocalDataAccess('~/ScenarioData')
test_fs.write('test.pkl', data)
test_fs.write('test.csv', data)
test_fs.write('test.mat', data)

### Time estimate
5 min


powersimdata/data_access/data_access.py
```diff
@@ -27,6 +31,27 @@ def __init__(self, root):
...
+        ext = os.path.basename(filepath).split(".")[-1]
+        if ext == "pkl":
+            self.fs.writebytes(filepath, pickle.dumps(data))
+        elif ext == "csv":
+            self.fs.writetext(filepath, data.to_csv())
+        elif ext == "mat":
+            tmp = io.BytesIO()
+            savemat(tmp, data, appendmat=False)
+            tmp.seek(0)
+            self.fs.writefile(filepath, tmp)
```
jon-hagg, line 51: I think we could standardize a bit. What I mean is, each of `pickle.dump`, `data.to_csv` and `savemat` can take a file object to write to. So if we did `with self.fs.openbin(filepath) as f:` before the if/else, then inside could just be `savemat(f, data, appendmat=False)`, etc. Of course, this only works if those calls don't assume it's an os file, and stick with the standard file api. So I see the advantage to the current approach, just thought we might try eliminating the extra write to the temp file here.

ahurli: Thanks! I didn't see the `openbin` option when looking through pyfilesystem, so thanks for pointing that out!

powersimdata/data_access/data_access.py
```diff
@@ -27,6 +30,25 @@ def __init__(self, root):
         self.root = root
         self.join = fs2.path.join
 
+    def write(self, filepath, data):
+        """Write a file from data store,
```
jon-hagg, line 34: Should it be "Write a file to data store"?

powersimdata/data_access/data_access.py
jon-hagg: This looks good. Did you run through usage examples again? Just want to make sure the 3 methods don't have any hidden assumption about the file object being specifically an os file.

PR #561, opened 2021-10-18
Feat: Saving Change Table through Pyfilesystem

ahurli:

### Purpose
Previously, the change table was being saved directly to the local disk in `change_table.py`. Now, the change table is being saved through the configured pyfilesystem.

### What the code is doing
Instead of saving via the change table directly to the local file system, the change table data to be saved is instead passed to `InputData` so that it can be saved via pyfilesystem.

### Testing
Tested the change table upload in both the serve setup and dockerized version

### Where to look
Code has been removed from both `change_table.py` and `create.py`. Most of the change is in `input_data.py`. I also updated the `write` method in `data_access.py` to add print lines that match behavior previously present in `create.py`.

### Usage Example/Visuals
How the code can be used and/or images of any graphs, tables or other visuals (not always applicable). 

### Time estimate
10 min

### DIFF
--- a/powersimdata/input/input_data.py
+++ b/powersimdata/input/input_data.py
@@ -1,10 +1,8 @@
 import os
 
-import fs
 import pandas as pd
 
 from powersimdata.data_access.context import Context
-from powersimdata.data_access.data_access import LocalDataAccess
 from powersimdata.data_access.profile_helper import ProfileHelper
 from powersimdata.utility import server_setup
 from powersimdata.utility.helpers import MemoryCache, cache_key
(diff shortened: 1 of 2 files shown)


powersimdata/input/input_data.py
```diff
@@ -114,6 +120,25 @@ def get_profile_version(self, grid_model, kind):
         """
         return self.data_access.get_profile_version(grid_model, kind)
 
+    def save_change_table(self, ct, scenario_id):
+        """Saves change table to the data store.
+
+        :param dict ct: a change table
+        :raises IOError: if file already exists on local machine.
```
rouille, line 127: Missing entry for `scenario_id`. Raised `IOError` can be removed.

rouille (changes requested):

powersimdata/input/input_data.py
```diff
@@ -66,6 +67,10 @@ class InputData:
     def __init__(self, data_loc=None):
         """Constructor."""
         self.data_access = Context.get_data_access(data_loc)
+        if type(self.data_access) != LocalDataAccess:
+            self.local_data_access = LocalDataAccess(server_setup.LOCAL_DIR)
+        else:
+            self.local_data_access = self.data_access
```
jon-hagg, line 73: I believe after the recent update, `self.local_data_access`  is not needed anymore

Issue #552, opened 2021-09-22
Cleanup


ahurli:
- [ ] ~remove copy_from (in input_data, csv_store, output_data)~ not essential
- [x] remove move_to (in create CT, execute prepare*) (see further comment in PR #565)
- [x] cleanup _read_data in input_data

PR #565, opened 2021-11-03
Feat: Update Saving REISE.jl Inputs with Pyfilesystem

ahurli:

### Purpose
Update how the inputs to `REISE.jl` get saved to go through exclusively through pyfilesystem instead of saving to local and using pyfilesystem to move to the expected folders for `REISE.jl`

### What the code is doing
The write function in `data_access` was updated to reflect the fact that the `REISE.jl` inputs don't get saved locally if in the server mode. `export_data` was updated to return the mpc and mpc_storage data instead of saving directly to the local filesystem, though that option was kept in for researchers who may want to use that function to save and investigate the mpc files outside of a scenario workflow (e.g. without having to create a whole scenario and finding it in the `/tmp` folder). Finally, the `execute` state was updated to reflect these new changes and save the input files (mpc and profiles) via `pyfilesystem` and `data_access` instead.

### Testing
Tested locally, as well as with the dev server.

### Where to look
Only three files were updated. Additionally, I left `SimulationInput` within `execute` to be moved around at a later date.

### Time estimate
10 min

### DIFF
--- a/powersimdata/data_access/data_access.py
+++ b/powersimdata/data_access/data_access.py
@@ -61,33 +61,39 @@ class DataAccess:
 
         :param str filepath: path to save data to, with extension either 'pkl', 'csv', or 'mat'.
         :param (*pandas.DataFrame* or *dict*) data: data to save
-        :raises ValueError: if extension is unknown.
+        :param bool save_local: whether a copy should also be saved to the local filesystem, if
+            such a filesystem is configured. Defaults to True.
         """
-
-        ext = os.path.basename(filepath).split(".")[-1]
 
         self._check_file_exists(filepath, should_exist=False)
 
         print("Writing %s" % filepath)
 
-        if save_local:
-            f = self.local_fs.openbin(filepath, mode="w")
-        else:
-            f = self.fs.openbin(filepath, mode="w")
+        self._write(self.fs, filepath, data)
 
-        if ext == "pkl":
-            pickle.dump(data, f)
-        elif ext == "csv":
-            data.to_csv(f)
-        elif ext == "mat":
-            savemat(f, data, appendmat=False)
-        else:
-            raise ValueError("Unknown extension! %s" % ext)
+        if save_local and self.local_fs is not None:
+            self._write(self.local_fs, filepath, data)
 
-        f.close()
+    def _write(self, fs, filepath, data):
+        """Write a file to given data store.
 
-        if save_local and self.local_fs != self.fs:
-            fs2.copy.copy_file(self.local_fs, filepath, self.fs, filepath)
+        :param fs fs: pyfilesystem to which to write data
+        :param str filepath: path to save data to, with extension either 'pkl', 'csv', or 'mat'.
+        :param (*pandas.DataFrame* or *dict*) data: data to save
+        :raises ValueError: if extension is unknown.
+        """
+
+        ext = os.path.basename(filepath).split(".")[-1]
+
+        with fs.openbin(filepath) as f:
+            if ext == "pkl":
+                pickle.dump(data, f)
+            elif ext == "csv":
+                data.to_csv(f)
+            elif ext == "mat":
+                savemat(f, data, appendmat=False)
+            else:
+                raise ValueError("Unknown extension! %s" % ext)
 
     def copy_from(self, file_name, from_dir):
         """Copy a file from data store to userspace.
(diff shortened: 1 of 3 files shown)


powersimdata/input/export_data.py
```diff
@@ -119,9 +125,12 @@ def export_case_mat(grid, filepath, storage_filepath=None):
             }
         }
 
-        savemat(storage_filepath, mpc_storage, appendmat=False)
+    if filepath:
```
jon-hagg, line 128: In general I recommend `if x is not None:` instead of `if x:`. The latter works fine in this case, but I think is less idiomatic. On a more practical level, `if x` depends on the type of `x` (including whether it has a `__bool__` method), so adds extra complexity.

powersimdata/data_access/data_access.py
```diff
@@ -70,17 +70,23 @@ def write(self, filepath, data):
...
-            elif ext == "csv":
-                data.to_csv(f)
-            elif ext == "mat":
-                savemat(f, data, appendmat=False)
-            else:
-                raise ValueError("Unknown extension! %s" % ext)
+        if save_local:
+            f = self.local_fs.openbin(filepath, mode="w")
+        else:
+            f = self.fs.openbin(filepath, mode="w")
```
jon-hagg, line 76: I feel like the `if/else` could be built into the inheritance (or some other) structure, but not sure how it would look. I'll see if I can come up with something unless you get to it first.

powersimdata/input/export_data.py
ahurli: Sounds good! Thanks!

powersimdata/data_access/data_access.py
ahurli: I agree; I don't think this is the cleanest solution, but I also think it might be a little premature to try and design it now as I'm imagining a bigger change that would be outside the scope of this PR. I'm still not sure that having the local filesystem as part of the data access class makes the most sense since I'm envisioning a scenario object with multiple data access objects (e.g. SSH, blob, local).

But I would love to see what you're talking about!

jon-hagg:
Just noticed that after this change, we will have eliminated all usages of `DataAccess.move_to` outside of `data_access.py` (and there is only one usage there). So either as part of this PR or another one I would suggest we rename it to `_move_to` to indicate that it's not part of the public api anymore.

powersimdata/scenario/execute.py
```diff
@@ -205,17 +203,16 @@ def prepare_mpc_file(self):
...
-        )
-        if len(self.grid.storage["gen"]) > 0:
-            self._data_access.move_to(
-                storage_file_name, self.REL_TMP_DIR, change_name_to="case_storage.mat"
-            )
+        mpc, mpc_storage = export_case_mat(self.grid)
+
+        self._data_access.write(file_path, mpc, save_local=False)
+
+        if mpc_storage:
```
jon-hagg, line 214: One more here - `if mpc_storage is not None`

powersimdata/data_access/data_access.py
jon-hagg: What about something like this? Regarding a scenario having multiple data access objects, I'm starting to feel we can eliminate `LocalDataAccess` pending a few more changes, and combine both blob and ssh filesystems into a `RemoteDataAccess`, or even just the base class and exposing some configurability. We should discuss more, but I think it is separate from this PR.

```
    def write(self, filepath, data, save_local=True):
        self._check_file_exists(filepath, should_exist=False)
        print("Writing %s" % filepath)

        self._write(self.fs, filepath, data)
        if save_local:
            self._write(self.local_fs, filepath, data)


    def _write(self, fs, filepath, data):
        ext = os.path.basename(filepath).split(".")[-1]

        with fs.openbin(filepath) as f:
            if ext == "pkl":
                pickle.dump(data, f)
            elif ext == "csv":
                data.to_csv(f)
            elif ext == "mat":
                savemat(f, data, appendmat=False)
            else:
                raise ValueError("Unknown extension! %s" % ext)
```

ahurli:
> Just noticed that after this change, we will have eliminated all usages of `DataAccess.move_to` outside of `data_access.py` (and there is only one usage there). So either as part of this PR or another one I would suggest we rename it to `_move_to` to indicate that it's not part of the public api anymore.

Good call out! I was planning on going through and cleaning up the various data access methods as part of Issue #552, so I'll add it to the list.

powersimdata/scenario/execute.py
jon-hagg: last one - just in case this got lost among the other comments. i'll go ahead and approve since the rest looks good

jon-hagg (approved):
Thanks, LGTM

PR #566, opened 2021-11-22
feat: add functionality to load Grid objects with HIFLD grid model

### LINKED ISSUE #551: Enable HIFLD grid model to instantiate Grid objects
# :rocket:

- [x] Is your feature request essential for your project?

### Describe the workflow you want to enable
I wish I could instantiate a Grid object with `grid_model="hifld"`.

### Describe your proposed implementation
We need to:
- Create a module `powersimdata.network.hifld`.
- Define a class `HIFLD` within `powersimdata.network.hifld.model`
- Create files `powersimdata.network.hifld.constants.plants`, `powersimdata.network.hifld.constants.storage`, `powersimdata.network.hifld.constants.zones`
- Update `powersimdata.network.model._check_model` to allow `hifld` as a possible grid model.
- Import the `HIFLD` class into `powersimdata.input.grid` and instantiate these objects as necessary.
- More?

### Additional context
There's a **rough** implementation currently in the `daniel/read_hifld_grid` branch, that worked when it was written but has not been tested lately, so may have broken since then.

danielolsen:

### Purpose
Allows the user to instantiate HIFLD grid objects by adding the necessary folder structure and sub-modules. Closes #551.

### What the code is doing
- In `powersimdata.network.hifld.model`: defining an HIFLD class, and populating it with necessary methods. Most of this code is copied directly from `powersimdata.network.usa_tamu.model`
- In `powersimdata.input.grid`: adding `"hifld"` as an allowable model/source input, and instantiating an `HIFLD` object as appropriate.
- In `powersimdata.network.model`: adding `"hifld"` as an acceptable model for instantiating a `ModelImmutables` object.
- Mostly-empty `__init__.py` files as necessary within `powersimdata.network.hifld.constants`. These will be filled in before this PR gets promoted out of draft stage.

### Testing
Current testing relies on https://github.com/Breakthrough-Energy/PreREISE/pull/236 ~and https://github.com/Breakthrough-Energy/PreREISE/pull/240. Through their combination in the PreREISE testing branch `daniel/hifld_top_level_rebased`, and with the branch for this PR~ (EDIT: #240 has been merged). Using the branch for #236, we can:
```python
# Building the hifld grid
# This step currently takes 20-30 minutes to download/process all data
from prereise.gather.griddata.hifld import create_csvs
create_csvs("path/to/powersimdata/network/hifld/data")

# Loading and checking the hifld grid
from powersimdata import Grid
from powersimdata.input.check import check_grid
check_grid(Grid("USA", "hifld"))
check_grid(Grid("Eastern", "hifld"))
check_grid(Grid("Western", "hifld"))
check_grid(Grid("ERCOT", "hifld"))
check_grid(Grid(["Eastern", "Western"], "hifld"))
check_grid(Grid(["Eastern", "ERCOT"], "hifld"))
check_grid(Grid(["ERCOT", "Western"], "hifld"))
```

Checks should all pass, returning `None` without raising any exceptions.

Other entries within `powersimdata.network.hifld.constants` still need to be populated, but those will be just straightforward data-filling and shouldn't affect code functionality, besides enabling the relevant information to be looked up for hifld Grid object by other functions/methods.

### Time estimate
15-30 minutes. Most new code isn't actually new, just copied with slight tweaks.

### DIFF
--- a/powersimdata/network/hifld/constants/plants.py
+++ b/powersimdata/network/hifld/constants/plants.py
@@ -1,0 +1,105 @@
+_exports = [
+    "all_resources",
+    "carbon_per_mmbtu",
+    "carbon_per_mwh",
+    "carbon_resources",
+    "clean_resources",
+    "label2type",
+    "nox_per_mwh",
+    "renewable_resources",
+    "so2_per_mwh",
+    "type2color",
+    "type2hatchcolor",
+    "type2label",
+]
+
+type2color = {
+    "wind": "xkcd:green",
+    "solar": "xkcd:amber",
+    "hydro": "xkcd:light blue",
+    "ng": "xkcd:orchid",
+    "nuclear": "xkcd:silver",
+    "coal": "xkcd:light brown",
+    "geothermal": "xkcd:hot pink",
+    "dfo": "xkcd:royal blue",
+    "biomass": "xkcd:dark green",
+    "other": "xkcd:melon",
+    "storage": "xkcd:orange",
+    "wind_offshore": "xkcd:teal",
+    "solar_curtailment": "xkcd:amber",
+    "wind_curtailment": "xkcd:green",
+    "wind_offshore_curtailment": "xkcd:teal",
+}
+
+type2label = {
+    "nuclear": "Nuclear",
+    "geothermal": "Geo-thermal",
+    "coal": "Coal",
+    "dfo": "DFO",
+    "hydro": "Hydro",
+    "ng": "Natural Gas",
+    "solar": "Solar",
+    "wind": "Wind",
+    "wind_offshore": "Wind Offshore",
+    "biomass": "Biomass",
+    "other": "Other",
+    "storage": "Storage",
+    "solar_curtailment": "Solar Curtailment",
+    "wind_curtailment": "Wind Curtailment",
+    "wind_offshore_curtailment": "Offshore Wind Curtailment",
+}
+
+type2hatchcolor = {
+    "solar_curtailment": "xkcd:grey",
+    "wind_curtailment": "xkcd:grey",
+    "wind_offshore_curtailment": "xkcd:grey",
+}
+
+label2type = {value: key for key, value in type2label.items()}
+
+renewable_resources = {"solar", "wind", "wind_offshore"}
+carbon_resources = {"coal", "ng", "dfo"}
+clean_resources = renewable_resources | {"geothermal", "hydro", "nuclear"}
+all_resources = carbon_resources | {"other"} | clean_resources
+
+
+# MWh to kilograms of CO2
+# Source: IPCC Special Report on Renewable Energy Sources and Climate Change
+# Mitigation (2011), Annex II: Methodology, Table A.II.4, 50th percentile
+# http://www.ipcc-wg3.de/report/IPCC_SRREN_Annex_II.pdf
+carbon_per_mwh = {
+    "coal": 1001,
+    "dfo": 840,
+    "ng": 469,
+}
+
+# MMBTu of fuel per hour to kilograms of CO2 per hour
+# Source: https://www.epa.gov/energy/greenhouse-gases-equivalencies-calculator-calculations-and-references
+# = (Heat rate MMBTu/h) * (kg C/mmbtu) * (mass ratio CO2/C)
+carbon_per_mmbtu = {
+    "coal": 26.05,
+    "dfo": 20.31,
+    "ng": 14.46,
+}
+
+# MWh to kilograms of NOx
+# Source: EPA eGrid 2018, tab 'US18' (U.S. summary), columns AN to AP
+# https://www.epa.gov/egrid/egrid-questions-and-answers
+nox_per_mwh = {
+    "coal": 0.658,
+    "dfo": 1.537,
+    "ng": 0.179,
+}
+
+# MWh to kilograms of SO2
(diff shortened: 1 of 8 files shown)


rouille:
I know we made a mistake to use Texas in place of ERCOT for the TAMU grid but should right our mistake for HIFLD? We have a lot of code that check/analyze/plot interconnect(s), I presume these expect Texas.

danielolsen:
> I know we made a mistake to use Texas in place of ERCOT for the TAMU grid but should right our mistake for HIFLD? We have a lot of code that check/analyze/plot interconnect(s), I presume these expect Texas.

I think any code that's not model-specific should look to the grid model for the appropriate interconnections. If our check/analyze/plot code isn't working, then we should change it to look for the right information in the right place (and make sure that information is there). We'll need this sooner or later, either for the new USA grid model or for models of other grids (e.g. Europe).

danielolsen:
After tweaking the output plant CSV a bit more within https://github.com/Breakthrough-Energy/PreREISE/pull/236 (adding type, adding heat rate columns, translating to PowerSimData naming expectations), and updating the placeholders within in `constants` sub-module, we can also successfully call `powersimdata.input.export_data.export_case_mat` on the resulting grids.

danielolsen:
The latest commits add the missing constants required to fully populate the `ModelImmutables` object which should enable the full suite of Grid analyses:
```python
>>> from powersimdata.network.model import ModelImmutables
>>> mi_tamu = ModelImmutables("usa_tamu")
>>> mi_hifld = ModelImmutables("hifld")
>>> mi_tamu.plants.keys() == mi_hifld.plants.keys()
True
>>> mi_tamu.zones.keys() == mi_hifld.zones.keys()
True
>>> mi_tamu.storage.keys() == mi_hifld.storage.keys()
True
```
`constants.plants` and `constants.storage` are identical between the `usa_tamu` and `hifld` grid models, and `constants.zones` only differs by using `"ERCOT"` in place of `"Texas"`.

Therefore, this is now ready for a full review

powersimdata/network/hifld/model.py
```diff
@@ -0,0 +1,119 @@
...
+    add_coord_to_grid_data_frames,
+    add_zone_to_grid_data_frames,
+    csv_to_data_frame,
+)
+from powersimdata.network.csv_reader import CSVReader
+from powersimdata.network.usa_tamu.constants.storage import defaults
+
+
+class HIFLD(AbstractGrid):
+    def __init__(self, interconnect):
```
jon-hagg, line 14: I think we could move all the instance methods to `AbstractGrid`, then just declare:
```
class HIFLD(AbstractGrid):
    pass
```
Same for the TAMU grid

powersimdata/network/hifld/constants/plants.py
```diff
@@ -0,0 +1,105 @@
...
+    "solar": "Solar",
+    "wind": "Wind",
+    "wind_offshore": "Wind Offshore",
+    "biomass": "Biomass",
+    "other": "Other",
+    "storage": "Storage",
+    "solar_curtailment": "Solar Curtailment",
+    "wind_curtailment": "Wind Curtailment",
+    "wind_offshore_curtailment": "Offshore Wind Curtailment",
+}
```
rouille, line 50: I know that PostREISE plotting functions rely on this dictionary and it is convenient to group generator type and curtailment but it is ugly. Curtailment color label/color should be moved in a dedicated dictionary. This completely out of the scope of this PR.

powersimdata/network/hifld/constants/plants.py
```diff
@@ -0,0 +1,105 @@
...
+
+type2hatchcolor = {
+    "solar_curtailment": "xkcd:grey",
+    "wind_curtailment": "xkcd:grey",
+    "wind_offshore_curtailment": "xkcd:grey",
+}
```
rouille, line 56: Same here curtailment is not a generator type

powersimdata/network/hifld/model.py
rouille: It makes sense. Is that still an abstract class then. It would look like a Parent class. Do we need to rename it?

jon-hagg:
> It makes sense. Is that still an abstract class then. It would look like a Parent class. Do we need to rename it?

An abstract class can have some behavior defined. Typically it would also declare abstract methods, to be implemented by derived classes. So if there are only "concrete" methods, a rename could make sense, e.g. `BaseGrid`. I don't feel strongly about it though since python doesn't enforce such things. Compared to e.g. java/c# where `abstract` is a keyword and enforced by the compiler.

powersimdata/network/hifld/model.py
jon-hagg: Oops, didn't see this one so replied below

danielolsen: Could we do within the child classes:
```python
def __init__(self):
    self.top_dirname = os.path.dirname(__file__)
    super().__init__()
```
and then within the parent class:
````python
def __init__(self):
    self._set_data_loc()
    self.interconnect = check_and_format_interconnect(interconnect)
    self._build_network()

def _set_data_loc(self):
    data_loc = os.path.join(self.top_dirname, "data")
    if os.path.isdir(data_loc) is False:
        raise IOError("%s directory not found" % data_loc)
    else:
        self.data_loc = data_loc
```
?
So the child inits set the directory based on their `__file__`s, and the parent init uses the already-assigned directory within `_set_data_loc`.

jon-hagg (approved):

PR #568, opened 2022-01-06
feat: add mapping from state abbreviations to load zone IDs

lanesmith:

### Purpose
Add a mapping from state abbreviations to load zone IDs to `zones.py`, where other similar mappings are found.

### What the code is doing
This mapping essentially inverts the mapping from load zone IDs to state abbreviations, which is also found in `zones.py`.

### Testing
This new mapping was tested locally and works as expected.

### Where to look
The new mapping can be found in `powersimdata\network\usa_tamu\constants\zones.py`.

### Time estimate
This should be very quick. Let me know if you think other reviewers should be added.

### DIFF
--- a/powersimdata/network/usa_tamu/constants/zones.py
+++ b/powersimdata/network/usa_tamu/constants/zones.py
@@ -1,9 +1,11 @@
 import os
+from collections import defaultdict
 
 import pandas as pd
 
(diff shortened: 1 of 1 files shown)


BainanXia (approved):
Thanks

PR #573, opened 2022-01-12
Drop python 3.7 and install fs-azureblob

jon-hagg:

### Purpose
Install the fs-azureblob package for future use and drop support for python 3.7. 

### Where to look
Take a look at the package updates in Pipfile.lock, since regenerating the file pulled in new versions of other packages as well (including pandas). 

### Time estimate
5 min


danielolsen:
Pandas 1.3.3 vs. 1.3.5 shouldn't make too much of a difference I think. It's great that we're not on 1.1.x in PowerSimData, we should definitely upgrade across the other repos as well.

Should we also test against Python 3.10?

danielolsen (approved):
LGTM

BainanXia (approved):

PR #575, opened 2022-01-14
List profile versions from blob storage

jon-hagg:

### Purpose
Currently the way we do this involves parsing a `version.json` file located in blob storage which must match the versions that are actually available. However to list locally available profiles, we check the files directly, which is more reliable than using the json file. This PR makes it so we use the same code for both local and blob storage, with the only difference being which filesystem is provided. 

### What the code is doing
See above. It is a refactor to be more consistent and eliminate the need for the `version.json` file.

### Testing
Ran the `_local` and `_cloud` functions interactively. Added new test for the shared logic (which was already in use, but untested). 

### Time estimate
10 min

### DIFF
--- a/powersimdata/data_access/profile_helper.py
+++ b/powersimdata/data_access/profile_helper.py
@@ -7,6 +7,39 @@ from tqdm.auto import tqdm
 from powersimdata.utility import server_setup
 
 
+def _get_profile_version(_fs, kind):
+    """Returns available raw profiles from the give filesystem
+    :param fs.base.FS _fs: filesystem instance
+    :param str kind: *'demand'*, *'hydro'*, *'solar'* or *'wind'*.
+    :return: (*list*) -- available profile version.
+    """
+    matching = [f for f in _fs.listdir(".") if kind in f]
+    return [f.lstrip(f"{kind}_").rstrip(".csv") for f in matching]
+
+
+def get_profile_version_cloud(grid_model, kind):
+    """Returns available raw profile from blob storage.
+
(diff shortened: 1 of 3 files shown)


powersimdata/data_access/profile_helper.py
```diff
@@ -7,6 +7,39 @@
...
+def get_profile_version_local(grid_model, kind):
+    """Returns available raw profile from local file.
+
+    :param str grid_model: grid model.
+    :param str kind: *'demand'*, *'hydro'*, *'solar'* or *'wind'*.
+    :return: (*list*) -- available profile version.
+    """
+    profile_dir = fs.path.join(server_setup.LOCAL_DIR, "raw", grid_model)
+    lfs = fs.open_fs(profile_dir)
+    return _get_profile_version(lfs, kind)
```
BainanXia, line 40: Not sure whether there is an easy way to combine these two functions together and make location of the profiles a parameter.

BainanXia (approved):
LGTM.

powersimdata/data_access/profile_helper.py
jon-hagg: There might be a simplification we could do, say if we had the local and blob filesystems already created as part of some `DataAccess` object. This seemed like the simplest approach for now though.

rouille:
Very nice. I have never been a fan of the json file

jon-hagg:
> Very nice. I have never been a fan of the json file

Same. I've been looking forward to this PR ever since adding the json file.

PR #576, opened 2022-01-26
Enable Export from Scenario/Grid to pypsa.Network

FabianHofmann:

### Purpose

This PR introduces a new function `export_to_pypsa` in the module `powersimdata.input.export_data`. It may take a `Grid` object or a `Scenario` object. 

As I see it, the only information which is not translated is the `storages`. As discussed the `storages` structure in the `Grid` object will likely be changed in near*ish* future. A warning is given when the `Grid` object contains storages. 

  
### What the code is doing

Here a short example

```python 
from powersimdata.input.export_data import export_to_pypsa
from powersimdata import Grid

interconnect='USA'
g = Grid("USA")
n = export_to_pypsa(g)
```

The function has a flag `preserve_all_columns` which, when set to True, will store all columns given by the `Grid` object to the static `pypsa.Network` dataframes. If False, only columns which pypsa can make use of (and some important columns like `name`, `interconnect_sub_id`) will be stored. Default is False.  

Some constants are taken from `powersimdata.network.usa_tamu` and added to the network such as color coding and co2 emissions. This part could be more generalized, as it only works for the USA model (but I think this is the main use case for near future).  
 
### Testing

A unit test was added. It checks roughly that components were added. 

### Where to look

As discussed: Currently everything is in the `export_data` module. In the long term. one could create a dedicated module `translations` for any exports or imports to other packages.  

### Time estimate

~ 30-60 min

# Note

As for the capacity expansion, additional information could be translated to the `pypsa.Network` such as maximal capacity expansion. Further, I did not find any information of possible co2 targets. 

PyPSA is not yet able to store [Categorical data](https://pandas.pydata.org/pandas-docs/stable/user_guide/categorical.html)  columns in the static dataframe. I will likely look into this in the near future. This should reduce the memory usage significantly.

### DIFF
--- a/powersimdata/input/export_data.py
+++ b/powersimdata/input/export_data.py
@@ -1,9 +1,12 @@
 import copy
 
 import numpy as np
+import pandas as pd
 from scipy.io import savemat
 
 from powersimdata.input.transform_profile import TransformProfile
+from powersimdata.input.grid import Grid
+from powersimdata.network.usa_tamu.constants import plants as plantconstants
 
 
 def export_case_mat(grid, filepath=None, storage_filepath=None):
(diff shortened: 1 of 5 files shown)


powersimdata/input/export_data.py
```diff
@@ -1,9 +1,13 @@
 import copy
+import warnings
 
 import numpy as np
+import pandas as pd
 from scipy.io import savemat
 
 from powersimdata.input.transform_profile import TransformProfile
+from powersimdata.input.grid import Grid
```
danielolsen, line 9: This can be simplified to `from powersimdata import Grid`. For the most common user-facing objects, we export them from the module root for brevity.

powersimdata/input/export_data.py
```diff
@@ -1,9 +1,13 @@
...
+from powersimdata.network.usa_tamu.constants import plants as plantconstants
```
danielolsen, line 10: We don't want to assume that every grid will be a `usa_tamu` grid, since we're in the process of expanding our model coverage. To get the same set of information from an arbitrary grid, we can look at `Grid.model_immutables.plants`.

FabianHofmann:
@danielolsen thanks for the comments, I modified the imports. As to expect the tests fail as pypsa is not installed. I added a the requirement to "dev-tools" in the Pipfile. But I am not sure whether this is the right spot.

FabianHofmann:
Just as a follow-up: I noticed adding pypsa to the dev-packages does not do the work (sorry I have to get used to `tox`). In meanwhile: the pypsa export test is now skip if pypsa is not available, which is the case for the CI. The test passes locally though.

danielolsen:
@FabianHofmann I have an idea of how to get everything working properly, I sent you a message on slack.

powersimdata/input/export_data.py
```diff
@@ -140,3 +149,281 @@ def export_transformed_profile(kind, scenario_info, grid, ct, filepath):
     profile = tp.get_profile(kind)
     print(f"Writing scaled {kind} profile to {filepath} on local machine")
     profile.to_csv(filepath)
+
+
+def export_to_pypsa(scenario, preserve_all_columns=False):
```
danielolsen, line 154: Can we change the call signature to something more intuitive? Something like:
```
export_to_pypsa(scenario_or_grid, preserve_all_columns=False)
export_to_pypsa(scenario=None, Grid=None, preserve_all_columns=False)
export_to_pypsa(*, scenario=None, Grid=None, preserve_all_columns=False)
```
Having an input named `scenario` that can equivalently take a `Grid` object seems handy, but potentially confusing.

powersimdata/input/tests/test_export_to_pypsa.py
```diff
@@ -0,0 +1,44 @@
...
+    assert "ramp_10" not in n.generators
+    assert "rateB" not in n.lines
+    assert "QminF" not in n.links
+
+
+@pytest.mark.skipif(not pypsa_available, reason="Package PyPSA not available.")
+def test_export_grid_to_pypsa():
+    grid = Grid("USA")
+
+    n = export_to_pypsa(grid)
```
danielolsen, line 32: Are there any tests in PyPSA for whether a Network object is internally consistent, without trying to instantiate the optimization problem itself? E.g. that branch endpoints are present in the bus table, etc.

powersimdata/input/export_data.py
```diff
@@ -140,3 +149,281 @@ def export_transformed_profile(kind, scenario_info, grid, ct, filepath):
...
+    buses = grid.bus.rename(columns=bus_rename)
+    buses.control.replace([1, 2, 3, 4], ["PQ", "PV", "slack", ""], inplace=True)
+    buses["zone_name"] = buses.zone_id.map({v: k for k, v in grid.zone2id.items()})
+    buses["substation"] = "sub" + grid.bus2sub["sub_id"].astype(str)
+    buses["zone_name"] = buses.substation.map({v: k for k, v in grid.zone2id.items()})
+
+    # ensure compatibility with substations (these are imported later)
+    buses["is_substation"] = False
+    buses["interconnect_sub_id"] = -1
+    buses["name"] = ""
```
danielolsen, line 222: This function is pretty large. What do you think about separating constants from logic to keep the logic part shorter? E.g.
putting into the module main:
```
pypsa_const = {
    "bus": {
        "rename": foo,
        "rename_t": bar,
        "default_drop_cols": baz,
    },
    "generator": {...},
    "branch": {...},
    ...
}
```
and then just accessing these within the function rather than defining them? Alternatively, we could chunk the main function into smaller functions for `bus`, `generator`, `branch` etc. and then use the main function to delegate logic to these.

powersimdata/input/tests/test_export_to_pypsa.py
FabianHofmann: Unfortunately there are no dedicated tests. However some tests are run when the components are added to the network. For example when adding lines, it check that all `bus0` and `bus1` entries are existent buses in the network.

powersimdata/input/export_data.py
FabianHofmann: Yes! That is a good idea.

FabianHofmann:
FYI: When updating the dependencies including the `Pipfile.lock`, I had to replace a `unittest.TestCase.assertWarns` by a `pytest.warns` in the `design/tests/test_scenario_info.py` due a nasty bug reported in https://bugs.python.org/issue29620. The bad guy is the `pyomo` package which defines additional import handlers in `https://github.com/Pyomo/pyomo/blob/main/pyomo/common/dependencies.py`. I hope that replacement is suitable.

danielolsen:
> FYI: When updating the dependencies including the `Pipfile.lock`, I had to replace a `unittest.TestCase.assertWarns` by a `pytest.warns` in the `design/tests/test_scenario_info.py` due a nasty bug reported in https://bugs.python.org/issue29620. The bad guy is the `pyomo` package which defines additional import handlers in `https://github.com/Pyomo/pyomo/blob/main/pyomo/common/dependencies.py`. I hope that replacement is suitable.

Not a problem at all. I think we generally prefer `pytest` to `unittest`, so you're doing us a favor.

danielolsen:
I'm running into problems trying to optimize a Network object that's been created from a Scenario object with `pyomo=False`:
```python
>>> from powersimdata.input.export_data import export_to_pypsa
>>> scenario = Scenario()
>>> scenario.set_grid("usa_tamu", "Texas")
>>> scenario.set_base_profile("demand", "vSep2021")
>>> scenario.set_base_profile("hydro", "vJan2021")
>>> scenario.set_base_profile("solar", "vJan2021")
>>> scenario.set_base_profile("wind", "vJan2021")
>>> scenario.set_time("2016-01-01 00:00", "2016-01-03 23:00", "24H")
>>> n = export_to_pypsa(scenario)
>>> n.lopf(pyomo=False, solver_name="gurobi")
WARNING:pypsa.components:You have defined one or more shunt impedances. Shunt impedances are ignored by the linear optimal power flow (LOPF).
INFO:pypsa.linopf:Prepare linear problem
  File "C:\Python39\lib\site-packages\pypsa\components.py", line 650, in lopf
    return network_lopf_lowmem(self, **args)
  File "C:\Python39\lib\site-packages\pypsa\linopf.py", line 1199, in network_lopf
    fdp, problem_fn = prepare_lopf(n, snapshots, keep_files, skip_objective,
  File "C:\Python39\lib\site-packages\pypsa\linopf.py", line 906, in prepare_lopf
    define_kirchhoff_constraints(n, snapshots)
  File "C:\Python39\lib\site-packages\pypsa\linopf.py", line 418, in define_kirchhoff_constraints
    con = write_constraint(n, cycle_sum, '=', 0)
  File "C:\Python39\lib\site-packages\pypsa\linopt.py", line 258, in write_constraint
    exprs = 'c' + _str_array(cons, True) + ':\n' + lhs + sense + ' ' + rhs + '\n\n'
TypeError: can only concatenate str (not "int") to str
```

Trying to call `.lopf()` on a Network created from a Grid object seems to fail for both `pyomo=True` and `pyomo=False`, but that's more expected since we aren't supplying load profiles:

```python
>>> from powersimdata import Grid
>>> from powersimdata.input.export_data import export_to_pypsa
>>> grid = Grid("Texas")
>>> n = export_to_pypsa(grid)
>>> n.lopf()
WARNING:pypsa.components:You have defined one or more shunt impedances. Shunt impedances are ignored by the linear optimal power flow (LOPF).
INFO:pypsa.opf:Performed preliminary steps
INFO:pypsa.opf:Building pyomo model using `kirchhoff` formulation
C:\Python39\lib\site-packages\pypsa\pf.py:854: RuntimeWarning: divide by zero encountered in true_divide
  b = 1./np.concatenate([(c.df.loc[c.ind, attribute]).values \
C:\Python39\lib\site-packages\pypsa\pf.py:871: RuntimeWarning: invalid value encountered in multiply
  sub_network.p_branch_shift = -b*np.concatenate([(c.df.loc[c.ind, "phase_shift"]).values*np.pi/180. if c.name == "Transformer"
  File "C:\Python39\lib\site-packages\pypsa\components.py", line 648, in lopf
    return network_lopf(self, **args)
  File "C:\Python39\lib\site-packages\pypsa\opf.py", line 1653, in network_lopf
    network_lopf_build_model(network, snapshots, skip_pre=skip_pre,
  File "C:\Python39\lib\site-packages\pypsa\opf.py", line 1468, in network_lopf_build_model
    define_nodal_balance_constraints(network,snapshots)
  File "C:\Python39\lib\site-packages\pypsa\opf.py", line 1058, in define_nodal_balance_constraints
    l_constraint(network.model, "power_balance", power_balance,
  File "C:\Python39\lib\site-packages\pypsa\opt.py", line 208, in l_constraint
    v._data[i] = _GeneralConstraintData(constr_expr, v)
  File "C:\Python39\lib\site-packages\pyomo\core\base\constraint.py", line 293, in __init__
    self.set_value(expr)
  File "C:\Python39\lib\site-packages\pyomo\core\base\constraint.py", line 534, in set_value
    expr, self.name))
  File "C:\Python39\lib\site-packages\pyomo\core\base\component.py", line 277, in name
    return self.getname(fully_qualified=True)
  File "C:\Python39\lib\site-packages\pyomo\core\base\component.py", line 924, in getname
    raise RuntimeError("Fatal error: cannot find the component data in "
RuntimeError: Fatal error: cannot find the component data in the owning component's _data dictionary.
>>> n.lopf(pyomo=False)
WARNING:pypsa.components:You have defined one or more shunt impedances. Shunt impedances are ignored by the linear optimal power flow (LOPF).
INFO:pypsa.linopf:Prepare linear problem
  File "C:\Python39\lib\site-packages\pypsa\components.py", line 650, in lopf
    return network_lopf_lowmem(self, **args)
  File "C:\Python39\lib\site-packages\pypsa\linopf.py", line 1199, in network_lopf
    fdp, problem_fn = prepare_lopf(n, snapshots, keep_files, skip_objective,
  File "C:\Python39\lib\site-packages\pypsa\linopf.py", line 906, in prepare_lopf
    define_kirchhoff_constraints(n, snapshots)
  File "C:\Python39\lib\site-packages\pypsa\linopf.py", line 418, in define_kirchhoff_constraints
    con = write_constraint(n, cycle_sum, '=', 0)
  File "C:\Python39\lib\site-packages\pypsa\linopt.py", line 258, in write_constraint
    exprs = 'c' + _str_array(cons, True) + ':\n' + lhs + sense + ' ' + rhs + '\n\n'
TypeError: can only concatenate str (not "int") to str
```

Am I missing a step somewhere?

FabianHofmann:
@danielolsen I could reproduce the error, I had some wrong conversion of the power potentials (see report). It should be fixed now. However I would like to do some more tests before merging and let you review again. The tests with the optimization shows issues that I did not consider before. I let you know when I think every thing is alright.

danielolsen:
> @danielolsen I could reproduce the error, I had some wrong conversion of the power potentials (see report). It should be fixed now. However I would like to do some more tests before merging and let you review again. The tests with the optimization shows issues that I did not consider before. I let you know when I think every thing is alright.

Sounds good. When testing, I noticed a problem with the behavior of `Scenario.get_bus_demand()`: it would return the full year's profile, no matter the time period that the Scenario was defined over. Since the conversion code uses `get_bus_demand`, we also ended up with a full year in the `Network` object every time (the other profiles also behaved the same way). This was fixed in https://github.com/Breakthrough-Energy/PowerSimData/pull/581, so after you rebase onto the latest `develop` it should be easier to test the conversion and optimization on a Scenario with a narrow time window.

danielolsen:
When the latest version of this branch is rebased onto the latest `develop` commit (important because there's a new commit today fixing a bug with `get_demand` in `Create`-state Scenarios, see #587), calling `lopf` on a Network created via a Scenario object completes successfully! Tested with both `pyomo=True` and `pyomo=False`. Converting a Grid object to a Network object seems to work as well, although in some limited testing it seems that the `Texas` network is infeasible using the `Pd` values in the bus table (which is probably due to data issues on our end).

danielolsen (approved):
Looks great, thanks!

FabianHofmann:
I cannot really make sense of the error, but it seems not to be related with the changes of the PR...

danielolsen:
> I cannot really make sense of the error, but it seems not to be related with the changes of the PR...

I agree, this seems like possibly a side-effect of #586, but I don't know why the tests passed there but don't pass now.

FabianHofmann:
Hey @danielolsen, the PR #589 should fix it. After merging, the tests in this PR should also pass.

PR #578, opened 2022-01-26
test: add grid check for loop branches

danielolsen:

### Purpose
Add a check for branches which connect to the same bus.

### What the code is doing
Simple pandas comparisons, extending the existing `check_grid` function.

### Testing
Tested manually using the grid generated by the current `hifld` branch of PreREISE: the test fails. When we use the data generated by the updated branch from https://github.com/Breakthrough-Energy/PreREISE/pull/263, the test passes.

### Time estimate
5 minutes.


BainanXia (approved):
Make sense.

jon-hagg (approved):

PR #581, opened 2022-02-04
fix: when retrieving profiles for partial-year Scenario, slice to Scenario time range

danielolsen:

### Purpose
Currently, when we retrieve profiles for a partial-year Scenario, we still return the entire year's profile. This can cause problems, since the time-series results are only defined over the Scenario's time range (see https://github.com/Breakthrough-Energy/PostREISE/issues/327 as an example). This corrects that issue, by slicing both the raw profiles (demand, solar, hydro, wind) and the bus-distributed profiles (via the method/function `get_bus_demand`).

### What the code is doing
Within `input_data.py`, we decompose the current `get_bus_demand` function into a scenario-info-retrieval step and a decomposition step. The scenario-info-retrieval step now slices the demand profile by time before the decomposition step. The decomposition step is now refactored to avoid mutating the `bus` input.

Within `create.py` and `ready.py`: we slice profiles before returning them to the user.

### Testing
Tested manually: ~for now, unit tests to be added:~
**EDIT**: unit tests are also added.
```python
>>> scenario = Scenario()
>>> scenario.set_grid("usa_tamu", "Texas")
>>> scenario.set_base_profile("demand", "vJan2021")
>>> scenario.set_base_profile("solar", "vJan2021")
>>> # Before we set the time, the default is the full year
>>> scenario.get_demand().shape
--> Loading demand
(8784, 8)
>>> scenario.get_bus_demand().shape
--> Loading demand
(8784, 2000)
>>> scenario.get_solar().shape
--> Loading solar
(8784, 36)
>>> # After we set the time, we get just the slice
>>> scenario.set_time("2016-01-01 00:00", "2016-01-03 23:00", "24H")
>>> scenario.get_demand().shape
--> Loading demand
(72, 8)
>>> scenario.get_bus_demand().shape
--> Loading demand
(72, 2000)
>>> scenario.get_solar().shape
--> Loading solar
(72, 36)
```

### Time estimate
15 minutes.

### DIFF
--- a/powersimdata/input/input_data.py
+++ b/powersimdata/input/input_data.py
@@ -140,20 +140,28 @@ def _read_data(filepath):
     return data
 
 
-def get_bus_demand(scenario_info, grid):
-    """Returns demand profiles by bus.
+def distribute_demand_from_zones_to_buses(zone_demand, bus):
+    """Decomposes zone demand to bus demand based on bus 'Pd' column.
 
-    :param dict scenario_info: scenario information.
-    :param powersimdata.input.grid.Grid grid: grid to construct bus demand for.
-    :return: (*pandas.DataFrame*) -- data frame of demand.
+    :param pandas.DataFrame zone_demand: demand by zone. Index is timestamp, columns are
+        zone IDs, values are zone demand (MW).
+    :param pandas.DataFrame bus: table of bus data, containing at least 'zone_id' and
+        'Pd' columns.
+    :return: (*pandas.DataFrame*) -- data frame of demand. Index is timestamp, columns
+        are bus IDs, values are bus demand (MW).
+    :raises ValueError: if the columns of ``zone_demand`` don't match the set of zone
+        IDs within the 'zone_id' column of ``bus``.
     """
-    bus = grid.bus.copy()
-    demand = InputData().get_data(scenario_info, "demand")[bus.zone_id.unique()]
-    bus["zone_Pd"] = bus.groupby("zone_id")["Pd"].transform("sum")
-    bus["zone_share"] = bus["Pd"] / bus["zone_Pd"]
-    zone_bus_shares = pd.DataFrame(
-        {z: bus.groupby("zone_id").get_group(z).zone_share for z in demand.columns}
-    ).fillna(0)
-    bus_demand = demand.dot(zone_bus_shares.T)
+    if set(bus["zone_id"].unique()) != set(zone_demand.columns):
+        raise ValueError("zones don't match between zone_demand and bus dataframes")
+    grouped_buses = bus.groupby("zone_id")
+    bus_zone_pd = grouped_buses["Pd"].transform("sum")
+    bus_zone_share = pd.concat(
+        [pd.Series(bus["Pd"] / bus_zone_pd, name="zone_share"), bus["zone_id"]], axis=1
+    )
+    zone_bus_shares = bus_zone_share.pivot_table(
+        index="bus_id", columns="zone_id", values="zone_share", fill_value=0
+    )
+    bus_demand = zone_demand.dot(zone_bus_shares.T)
 
     return bus_demand
--- a/powersimdata/input/transform_profile.py
+++ b/powersimdata/input/transform_profile.py
@@ -6,16 +6,20 @@ from powersimdata.input.input_data import InputData
 class TransformProfile:
     """Transform profile according to operations listed in change table."""
 
-    def __init__(self, scenario_info, grid, ct):
+    _default_dates = {"start_date": "2016-01-01 00:00", "end_date": "2016-12-31 23:00"}
+
+    def __init__(self, scenario_info, grid, ct, slice=True):
         """Constructor.
 
         :param dict scenario_info: scenario information.
         :param powersimdata.input.grid.Grid grid: a Grid object previously
             transformed.
         :param dict ct: change table.
+        :param bool slice: whether to slice the profiles by the Scenario's time range.
         """
+        self.slice = slice
         self._input_data = InputData()
-        self.scenario_info = scenario_info
+        self.scenario_info = {**self._default_dates, **scenario_info}
 
         self.ct = copy.deepcopy(ct)
         self.grid = copy.deepcopy(grid)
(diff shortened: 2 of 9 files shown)


rouille:
Did you think about slicing in the `TransformProfile` class so it will be done in one location?

danielolsen:
> Did you think about slicing in the `TransformProfile` class so it will be done in one and only one location?

I don't think `TransformProfile` would cover everything, since `get_bus_demand` for the `Create` state goes through `InputData` instead. Most of the refactor code covers this use case.

rouille:
> > Did you think about slicing in the `TransformProfile` class so it will be done in one and only one location?
> 
> I don't think `TransformProfile` would cover everything, since `get_bus_demand` for the `Create` state goes through `InputData` instead. Most of the refactor code covers this use case. **EDIT**: now that I think about it a bit more, maybe `get_bus_demand` should be going through `TransformProfile` at some point? Since we could be scaling up demand for certain zones, but if that never gets applied then we're just getting the base bus demand.

Yeah, I was not sure. Do you need to slice the `get_demand` method of the `Analyze` class?

powersimdata/input/input_data.py
```diff
@@ -145,15 +145,42 @@ def get_bus_demand(scenario_info, grid):
...
+        IDs within the 'zone_id' column of ``bus``.
+    """
+    if set(bus["zone_id"].unique()) != set(zone_demand.columns):
+        raise ValueError("zones don't match between zone_demand and bus dataframes")
+    grouped_buses = bus.groupby("zone_id")
+    bus_zone_pd = grouped_buses["Pd"].transform("sum")
+    bus_zone_share = bus["Pd"] / bus_zone_pd
+    zone_bus_shares = pd.DataFrame(0, index=bus.index, columns=zone_demand.columns)
+    for zone, zone_buses in grouped_buses:
+        zone_bus_shares[zone].update(bus_zone_share.loc[zone_buses.index])
```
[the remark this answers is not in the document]
danielolsen: Nice! Adjusting for line lengths, we end up with 6 lines to replace the 4 from before, but we get rid of that ugly loop. Done (with slight tweaks).

danielolsen:
This has been refactored so that everything is done within TransformProfile. This should solve two issues:
- Ensuring that profiles are sliced to match the Scenario.
- Ensuring that `get_bus_demand` uses the scaled demand, not the raw demand.

In the process, we've removed `powersimdata.input.input_data.get_bus_demand`, which PostREISE currently relies on to calculate congestion surplus (see https://github.com/Breakthrough-Energy/PostREISE/pull/328), so I think we'll want to make a v0.5.0 release of PowerSimData, and then a v0.5.1 release of PostREISE which requires `PowerSimData ~=0.5.0`.

powersimdata/scenario/create.py
```diff
@@ -236,7 +240,7 @@ def get_profile(self, kind):
                 self.get_grid(),
                 self.get_ct(),
             )
-            return profile.get_profile(kind)
+            return profile.get_profile(kind).loc[self.start_date : self.end_date]
```
jon-hagg, line 243: Do we still need the `.loc` here?

powersimdata/scenario/ready.py
```diff
@@ -55,7 +55,9 @@ def get_profile(self, kind):
         :return: (*pandas.DataFrame*) -- profile.
         """
         profile = TransformProfile(self._scenario_info, self.get_grid(), self.get_ct())
-        return profile.get_profile(kind)
+        return profile.get_profile(kind).loc[
+            self._scenario_info["start_date"] : self._scenario_info["end_date"]
+        ]
```
jon-hagg, line 60: Still need `.loc`?

powersimdata/input/transform_profile.py
```diff
@@ -125,6 +135,6 @@ def get_profile(self, name):
         if name not in possible:
             raise ValueError("Choose from %s" % " | ".join(possible))
         elif name == "demand":
-            return self._get_demand_profile()
+            return self._slice(self._get_demand_profile())
         else:
-            return self._get_renewable_profile(name)
+            return self._slice(self._get_renewable_profile(name))
```
jon-hagg, line 140: I think we use this to write the profiles that are uploaded to the server, which I assume are used by REISE.jl directly. Is that fine?

powersimdata/scenario/create.py
danielolsen: Good catch: no, it's now redundant (here and in `Ready`). Fixed.

jon-hagg:
> I think we'll want to make a v0.5.0 release of PowerSimData, and then a v0.5.1 release of PostREISE

I think it's fine to have breaking changes in `develop` branch, we would just bump the version as you suggest whenever we do the next release. Any reason it would need to be done now?

danielolsen:
> I think it's fine to have breaking changes in `develop` branch, we would just bump the version as you suggest whenever we do the next release. Any reason it would need to be done now?

In order to fix https://github.com/Breakthrough-Energy/PostREISE/issues/327, we're updating PowerSimData, but PostREISE will only grab the latest numbered version, right?

jon-hagg:
Hmm I see. Depends on which images are used - we publish a `stable` image with the pypi packages, and a `latest` image with `develop` versions of both PowerSimData and PostREISE. But yeah we would need a release for the `stable` image to be fixed.

powersimdata/input/transform_profile.py
danielolsen: Oh yeah, good catch. We don't want to change the files that are prepared for REISE.jl, at least not right now.

jon-hagg: Perhaps a bool parameter that defaults to this behavior, but in `SimulationInput.prepare_profile` we pass the value that says to preserve the full profile?

jon-hagg: I guess in develop branch that would be in input/export_data.py (I have a different branch checked out where it's called directly)

danielolsen: Added in the latest commit. `SimulationInput.prepare_profile` now has a `slice` flag (defaulting to False) which gets passed to `input.export_data.export_transformed_profile` (default is True), which gets passed to `TransformProfile` (default is True).

When we refactor REISE.jl, I think we can just flip the `slice` flag in the `SimulationInput.prepare_profile` call, and everything else should work as expected.

danielolsen:
> Hmm I see. Depends on which images are used - we publish a `stable` image with the pypi packages, and a `latest` image with `develop` versions of both PowerSimData and PostREISE. But yeah we would need a release for the `stable` image to be fixed.

Right, and if anyone's working outside of the docker environment, we want `pip install postreise` to produce something working.

BainanXia (approved):
LGTM. Thanks for adding the tests. We will refactor REISE.jl to take sliced profiles instead, then remove the flag.

jon-hagg (approved):

Issue #489, opened 2021-05-26
Mock profiles for testing


jon-hagg:
See https://github.com/Breakthrough-Energy/PowerSimData/pull/485#issuecomment-848311743 for additional context.

We are using truncated profiles for tests that require them, but this requires hard coding the `test_usa_tamu` grid model, downloading from blob storage, and is limited to the plant ids matching the current grid. It would be more flexible if we could generate a mock profile based on an arbitrary grid (either real or mock), and use that for tests.

PR #586, opened 2022-02-14
test: Use fake profile data instead of test_usa_tamu

c-voegele:

### Purpose
Hi Breakthrough Energy Team! Thank you for your work on this interesting and important project!

This PR is an attempt to fix https://github.com/Breakthrough-Energy/PowerSimData/issues/489.

The  `test_usa_tamu` profiles are no longer needed for `test_transform_profile` and no Internet connection is required to run the tests.

### What the code is doing
To avoid the Azure blob call, I mocked `InputData` to return random profiles. I just used uniform random data in [0, 1) as suggested [here](https://github.com/Breakthrough-Energy/PowerSimData/pull/485#issuecomment-848331723)

### Testing
I ran `tox -e pytest-local -- --cov=powersimdata` and `tox -e checkformatting` and both are passing.

To make sure that `test_usa_tamu` was not required, I wiped `~/ScenarioData` and rebuilt it with `LocalConfig().initialize()`. After running the tests, `test_usa_tamu` is no longer loaded.

### Where to look
`powersimdata/input/tests/test_transform_grid.py`

### Usage Example/Visuals
n/a 

### Time estimate
15 minutes

### DIFF
--- a/powersimdata/input/tests/test_transform_profile.py
+++ b/powersimdata/input/tests/test_transform_profile.py
@@ -1,19 +1,17 @@
+from unittest.mock import patch
+
 import numpy as np
 import pytest
 from numpy.testing import assert_almost_equal
 
 from powersimdata.input.change_table import ChangeTable
 from powersimdata.input.grid import Grid
-from powersimdata.input.input_data import InputData
 from powersimdata.input.transform_grid import TransformGrid
 from powersimdata.input.transform_profile import TransformProfile
+from powersimdata.tests.mock_input_data import MockInputData
 
 interconnect = ["Western"]
 param = {
-    "demand": "vJan2021",
-    "hydro": "vJan2021",
-    "solar": "vJan2021",
-    "wind": "vJan2021",
     "n_zone_to_scale": 6,
     "n_plant_to_scale": 50,
     "n_plant_to_add": 100,
@@ -88,7 +86,7 @@ def get_change_table_for_new_plant_addition(base_grid, resource):
     return ct.ct
 
 
-def _check_plants_are_scaled(ct, base_grid, profile_info, raw_profile, resource):
+def _check_plants_are_scaled(ct, base_grid, raw_profile, resource):
     plant_id_type = get_plant_with_resource(base_grid, resource)
 
     base_profile = (
@@ -98,7 +96,8 @@ def _check_plants_are_scaled(ct, base_grid, profile_info, raw_profile, resource)
     tg = TransformGrid(base_grid, ct)
     transformed_grid = tg.get_grid()
 
-    tp = TransformProfile(profile_info, transformed_grid, ct)
+    empty_scenario_info = {}  # scenario_info not needed since input_data is mocked
+    tp = TransformProfile(empty_scenario_info, transformed_grid, ct)
     transformed_profile = tp.get_profile(resource)
 
     scaled_plant_id = []
@@ -134,7 +133,7 @@ def _check_plants_are_scaled(ct, base_grid, profile_info, raw_profile, resource)
     return transformed_profile
 
 
-def _check_new_plants_are_added(ct, base_grid, profile_info, raw_profile, resource):
+def _check_new_plants_are_added(ct, base_grid, raw_profile, resource):
     n_plant = param["n_plant_to_add"]
     plant_id_type = (
         base_grid.plant.isin(profile_type[resource]).query("type == True").index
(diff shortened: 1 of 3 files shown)


powersimdata/input/tests/test_transform_profile.py
```diff
@@ -36,6 +34,11 @@ def get_plant_with_resource(base_grid, resource):
     return plant_id
 
 
+def _get_plant_ids_for_type(grid, type):
+    plant_ids = list(grid.plant.isin(profile_type[type]).query("type == True").index)
```
danielolsen, line 38: I think `grid.plant.isin(profile_type[type]).query("type == True")` can be simplified to `grid.plant.query("type in @profile_type[@type]")`.

c-voegele: Done, thanks!

BainanXia:
This is nice! I'm wondering whether it is possible to pull `mock_input_data` and corresponding raw profile functions into a separate module within `powersimdata/tests/` together with other mocks so that it could potentially be used elsewhere, then we could import and patch it in `test_transform_profile.py`.

BainanXia:
@jon-hagg With the `mock_input_data` implemented, do we still need truncated test profiles in blob storage under `test_usa_tamu` folder?

jon-hagg:
> @jon-hagg With the `mock_input_data` implemented, do we still need truncated test profiles in blob storage under `test_usa_tamu` folder?

I would keep them for now - they are used [here](https://github.com/Breakthrough-Energy/PowerSimData/blob/develop/powersimdata/scenario/tests/test_scenario.py#L63) to test scenario preparation

c-voegele:
> This is nice! I'm wondering whether it is possible to pull `mock_input_data` and corresponding raw profile functions into a separate module within `powersimdata/tests/` together with other mocks so that it could potentially be used elsewhere, then we could import and patch it in `test_transform_profile.py`.

Thanks for your suggestion @BainanXia! I reorganized the code as you suggested. Let me know what you think!

powersimdata/tests/mock_input_data.py
```diff
@@ -0,0 +1,67 @@
+import numpy as np
+import pandas as pd
+
+from powersimdata.input.grid import Grid
+
+
+class MockInputData:
```
BainanXia, line 7: It would be great if we can have some doc strings here given it is now in a user facing `tests` module.

powersimdata/tests/mock_input_data.py
```diff
@@ -0,0 +1,67 @@
...
+        "wind": {"wind", "wind_offshore"},
+        "solar": {"solar"},
+        "hydro": {"hydro"},
+    }
+
+    def __init__(
+        self,
+        grid: Grid,
+        start_time="2016-01-01 00:00:00",
+        periods=24,
```
BainanXia, line 18: Given this mock class can potentially be used in a general sense, maybe we can make this flexible by specifying the `end_time` or we could keep `periods` argument as well just in case people don't want to bother.

BainanXia:
> > This is nice! I'm wondering whether it is possible to pull `mock_input_data` and corresponding raw profile functions into a separate module within `powersimdata/tests/` together with other mocks so that it could potentially be used elsewhere, then we could import and patch it in `test_transform_profile.py`.
> 
> Thanks for your suggestion @BainanXia! I reorganized the code as you suggested. Let me know what you think!

Thanks for the contribution! I think we are good to go except for some minor concerns above.

powersimdata/tests/mock_input_data.py
c-voegele: Sure, added docs and tests!

powersimdata/tests/mock_input_data.py
c-voegele: Done!

c-voegele:
> > > This is nice! I'm wondering whether it is possible to pull `mock_input_data` and corresponding raw profile functions into a separate module within `powersimdata/tests/` together with other mocks so that it could potentially be used elsewhere, then we could import and patch it in `test_transform_profile.py`.
> > 
> > 
> > Thanks for your suggestion @BainanXia! I reorganized the code as you suggested. Let me know what you think!
> 
> Thanks for the contribution! I think we are good to go except for some minor concerns above.

Thanks! I addressed your comments. Let me know if you find anything else!

powersimdata/input/tests/test_transform_profile.py
```diff
@@ -197,39 +191,42 @@ def base_grid():
...
+    return mock_input_data
+
+
+@pytest.fixture(scope="module", autouse=True)
+def mock_input_data_class(input_data):
+    with patch(
+        "powersimdata.input.transform_profile.InputData"
+    ) as mock_input_data_class:
+        mock_input_data_class.return_value = input_data
+        yield
```
BainanXia, line 206: This is clean.

BainanXia (approved):
Great! Thanks for improving test coverage and the patience @c-voegele

BainanXia:
> Thanks for the approval!
> 
> It doesn't look like I have write permissions so this is as far as I can go.

No worries, we will take care of it.

PR #584, opened 2022-02-09
Merge data access refactor feature branch 

jon-hagg:
[Pull Request doc](https://breakthrough-energy.github.io/docs/user/git_guide.html#d-pull-request)

### Purpose
Data access refactor and blob storage integration. Someone needs to approve it but I can leave open if folks haven't had a chance to review the recent PR, just lemme know.

### Testing
This was tested in #570 

### Usage Example/Visuals
Here is the output I get when I change the local directory from `ScenarioData` to an empty directory, then load a scenario from blob storage.
```
In [1]: from powersimdata import Scenario

In [2]: s = Scenario(1270)
Could not connect to ssh server
Initialized remote filesystem with profile_fs,scenario_fs
Transferring ScenarioList.csv.2 from scenario_fs
Transferring ExecuteList.csv.2 from scenario_fs
SCENARIO: Julia | USA2030HVDC_Design3_OB1_Mesh500x38

--> State
analyze
--> Loading grid
data/input/1270_grid.mat not found on local machine
Could not connect to ssh server
Initialized remote filesystem with profile_fs,scenario_fs
Transferring 1270_grid.mat from scenario_fs
Loading bus
Loading plant
Loading heat_rate_curve
Loading gencost_before
Loading gencost_after
Loading branch
Loading dcline
Loading sub
Loading bus2sub
Could not connect to ssh server
Initialized remote filesystem with profile_fs,scenario_fs
Transferring ScenarioList.csv.2 from scenario_fs
--> Loading ct
data/input/1270_ct.pkl not found on local machine
Transferring 1270_ct.pkl from scenario_fs
```

### Time estimate
n/a - code has been reviewed

PR #588, opened 2022-02-14
Fix data access integration tests

danielolsen:

### Purpose
Fix failing integration tests. This is a clean-up from #584.

### What the code is doing
- `SSHDataAccess` gets an `exec_command` method which passes the command through to the SSH filesystem.
- Two obsolete tests are removed from the ScenarioListManager and ExecuteListManager test sets. @jon-hagg said 'they were testing fallback logic that moved from `csv_store.py` to the `copy_from` method in `DataAccess`, so that is tested elsewhere'.

### Testing
`pytest .` passes all tests when connected to the VPN, and when not connected to the VPN the only test failure is `test_setup_server_connection`.

### Time estimate
5 minutes to understand these changes, longer if you want to go back to #584 and trace the evolution of tests there.

### DIFF
--- a/powersimdata/data_access/data_access.py
+++ b/powersimdata/data_access/data_access.py
@@ -257,6 +257,10 @@ class SSHDataAccess(DataAccess):
             self._fs = get_multi_fs(self.root)
         return self._fs
 
+    def exec_command(self, command):
+        ssh_fs = self.fs.get_fs("ssh_fs")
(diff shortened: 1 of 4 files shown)


kasparm (approved):
Works.
Thanks for the fix!

BainanXia (approved):

PR #589, opened 2022-02-24
mock_input_data: fix eval computation for new pandas version

FabianHofmann:

### Purpose
With the up-to-date pandas version the eval string in `powersimdata/tests/mock_input_data.py` fails due to a chained call of the `@` accessor. This PR provides a fix for that the error. 

### What the code is doing
Instead of the `eval` call, use a lambda function for the selection.  

### Testing
It fixes the tests with the new pandas version.

### Time estimate
3 min

### DIFF
--- a/powersimdata/tests/mock_input_data.py
+++ b/powersimdata/tests/mock_input_data.py
@@ -90,9 +90,8 @@ class MockInputData:
         :param str resource_type: Can be any of *'hydro'*, *'solar'*, or *'wind'*.
         :return: (*list*) -- list of plant_ids
         """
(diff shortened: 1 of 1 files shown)


danielolsen (approved):
The lambda here is pretty cool! Tests pass:
- locally in an environment with `numexpr`.
- when the branch from #576 is rebased onto this one, in a docker environment that emulates the CI environment where the #576 tests had been failing.

jon-hagg (approved):

BainanXia (approved):
Nice. Haven't tried using `lambda` for indexing directly before.

Issue #511, opened 2021-06-29
Integrate demand flexibility in the scenario framework


lanesmith:
# :rocket:

- [x] Is your feature request essential for your project?

### Describe the workflow you want to enable
I wish I could specify demand flexibility profiles, costs, and parameters from the scenario framework in `PowerSimData`. Rather than the current method for testing demand flexibility, which involves pairing demand flexibility profiles and parameters with old scenario files, this would allow proper test scenarios with demand flexibility to be created and considered.

This feature is not essential in that it is preventing test simulations that include demand flexibility to be run. However, including this feature will allow specific test scenarios to be created, allowing smaller simulations to be created (e.g., ERCOT). This will be useful for testing since demand flexibility still has a higher computational cost than a simulation without demand flexibility. Finally, including demand flexibility in the scenario framework is important as `PowerSimData` is where the feature reduction functionality will likely live.

### Describe your proposed implementation
Many files within `powersimdata/input`, `powersimdata/output`, and `powersimdata/scenario` will need to be updated to include this functionality. Overall, I propose implementing something similar to storage, where flexible demand will be viewed as a resource that is added rather than a resource that is always there (e.g., the base configuration that is seen with demand, solar, wind, etc.) and then scaled. This should also hopefully help with backward compatibility of existing scenarios, a concern that was raised during a conversation with @BainanXia. Flexible demand will still differ from the storage implementation in that it will need to contain both static information and time series data.

Some other issues raised in conversation with @BainanXia that will need to be addressed include (1) how specific demand flexibility nodes will be passed through the framework for use in `REISE.jl` and (2) how the demand flexibility profiles are able to be temporally reduced for inclusion in `SwitchWrapper`. For (1), this is something that will be important, though probably not as important initially. Currently, the assumption that demand flexibility occurs at every demand node should allow the time series data to be passed similar to how the regular demand data is passed, with mapping the demand flexibility to the buses occurring within `REISE.jl`. Eventually, the flexible demand feature reduction will identify a set of flexible demand nodes and those nodes will need to be passed similarly to what appears to be done for storage. This functionality will require changes to how demand flexibility is handled in `REISE.jl` too. For (2), considering how demand flexibility is used within `SwitchWrapper` will rely on a couple things. First, it will need to be determined if Switch's representation of demand flexibility is amenable to our method for handling demand flexibility. Second, it will probably depend on the type of temporal reduction that is used. The integration to `SwitchWrapper` is a much more open question.

[...]

PR #590, opened 2022-02-28
Specify demand flexibility in the scenario framework

lanesmith:

### Purpose
The goal of this PR is to allow users to specify demand flexibility profiles and parameters in `PowerSimData`. This PR closes Issue #511.

### What the code is doing
Demand flexibility profiles and parameters are specified using the change table (via the `add_demand_flexibility` function). Demand flexibility profiles are accessed using the same core functionality as the other profile data. Accessing the demand flexibility profiles does differ slightly in that the profiles are not scaled (as you might see with demand, solar, wind, and hydro) and different area granularity can potentially be included (some combination of zone IDs and bus IDs, consistent with the allowed inputs in `REISE.jl`). This functionality is located in the `_get_demand_flexibility_profile` function. The final main addition is that the demand flexibility parameters file is created (via the `prepare_demand_flexibility_parameters` function) using the parameters input by the user. Other changes are less significant, such as including demand flexibility as a recognizable profile type and differentiating between demand profiles and demand flexibility profiles.

### Testing
Tests were completed locally using the Native set-up specified [here](https://breakthrough-energy.github.io/docs/user/installation_guide.html#natively). As has been discussed, there is an error when trying to run the specified scenario in `REISE.jl` from `PowerSimData`. Since the scenario creation works as expected (i.e., all input files are created in a single directory) and those files can be used in the standalone version of `REISE.jl`, I believe this problem lies not with the feature presented in this PR, but with the Native set-up (it seems to be an environment issue). This will be explored in a separate PR.

I have also started to include pytests (located in `test_change_table.py`), but I ran into an issue with what I was able to implement. Many of the demand-flexibility-related functions rely on checking or including the names of the different versions of the different demand flexibility profiles (e.g., version `EFS_x_y_z` of the demand flexibility up profile). Since none of these profiles are included in blob storage, the only location they are looked for is locally; unless these profiles are included locally, the functions will fail (e.g., as would be the case when pushing to GitHub). I was trying to implement something like what is seen in `mock_input_data.py` and `test_transform_profile.py`, but I couldn't figure out how to overcome the functions looking for the existence of the profiles. Any suggestions are welcome.

### Where to look
Most of the main changes are located in:
- `powersimdata/input/change_table.py`
- `powersimdata/input/transform_profile.py`
- `powersimdata/scenario/execute.py`

Smaller ancillary changes were made to:
- `powersimdata/data_access/profile_helper.py`
- `powersimdata/data_access/scenario_list.py`
- `powersimdata/input/input_data.py`

`powersimdata/output/output_data.py` and `powersimdata/scenario/analyze.py` include some changes that allow the demand-flexibility-related data to be accessed following the simulation. The test is located in `powersimdata/input/tests/test_change_table.py`.

### Time estimate
This will probably take a bit longer since it is a fairly big new feature.

### DIFF
--- a/powersimdata/data_access/scenario_list.py
+++ b/powersimdata/data_access/scenario_list.py
@@ -75,7 +75,7 @@ class ScenarioListManager(CsvStore):
         scenario_info.move_to_end("id", last=False)
         table.reset_index(inplace=True)
         entry = pd.DataFrame({k: [v] for k, v in scenario_info.items()})
-        table = table.append(entry)
+        table = pd.concat([table, entry])
         table.set_index("id", inplace=True)
 
         print("--> Adding entry in %s" % self._FILE_NAME)
--- a/powersimdata/input/tests/test_change_table.py
+++ b/powersimdata/input/tests/test_change_table.py
@@ -438,13 +438,26 @@ def test_add_new_elements_at_new_buses(ct):
 
 def test_change_table_clear_success(ct):
     fake_scaling = {"demand", "branch", "solar", "ng_cost", "coal_pmin", "dcline"}
-    fake_additions = {"storage", "new_dcline", "new_branch", "new_plant"}
+    fake_additions = {
+        "storage",
+        "new_dcline",
+        "new_branch",
+        "new_plant",
+        "demand_flexibility",
+    }
     all_fakes = fake_scaling | fake_additions
     original_dict_object = ct.ct
     for fake in all_fakes:
         ct.ct[fake] = {}
     # Test that each individual clear makes a change, and the ct ends up empty
-    clear_keys = {"branch", "dcline", "demand", "plant", "storage"}
+    clear_keys = {
+        "branch",
+        "dcline",
+        "demand",
+        "plant",
+        "storage",
+        "demand_flexibility",
+    }
     for key in clear_keys:
         old_keys = set(ct.ct.keys())
         ct.clear(key)
@@ -492,3 +505,21 @@ def test_remove_bus(ct):
         ct.remove_bus({845})
     ct.scale_plant_capacity(resource="ng", plant_id={0: 0})
     ct.remove_bus({845})
+
+
+def test_add_demand_flexibility(ct):
+    with pytest.raises(Exception):
+        # Fails because "demand_flexibility_dn", a required key, is not included
+        ct.add_demand_flexibility(
+            {"demand_flexibility_up": "test", "demand_flexibility_duration": 6}
+        )
+
+    with pytest.raises(Exception):
+        # Fails because
+        ct.add_demand_flexibility(
+            {
+                "demand_flexibility_up": "test",
+                "demand_flexibility_dn": "test",
+                "demand_flexibility_duration": 6,
+            }
+        )
--- a/powersimdata/input/tests/test_change_table.py
+++ b/powersimdata/input/tests/test_change_table.py
@@ -513,13 +513,13 @@ def test_remove_bus(ct):
 
 
 def test_add_demand_flexibility(ct):
-    with pytest.raises(Exception):
+    with pytest.raises(ValueError):
         # Fails because "demand_flexibility_dn", a required key, is not included
         ct.add_demand_flexibility(
             {"demand_flexibility_up": "Test", "demand_flexibility_duration": 6}
         )
 
-    with pytest.raises(Exception):
+    with pytest.raises(ValueError):
         # Fails because there is a key that should not be there
         ct.add_demand_flexibility(
             {
@@ -530,7 +530,7 @@ def test_add_demand_flexibility(ct):
             }
         )
 
-    with pytest.raises(Exception):
+    with pytest.raises(ValueError):
         # Fails because there are no profiles available that match the specified version
         ct.add_demand_flexibility(
             {
--- a/powersimdata/input/tests/test_change_table.py
+++ b/powersimdata/input/tests/test_change_table.py
@@ -1,12 +1,10 @@
-import os
-import shutil
-from pathlib import Path
-
 import pandas as pd
 import pytest
 
+from powersimdata.data_access.context import Context
 from powersimdata.input.change_table import ChangeTable
 from powersimdata.input.grid import Grid
+from powersimdata.tests.mock_context import MockContext
 
 grid = Grid(["USA"])
 
@@ -512,7 +510,7 @@ def test_remove_bus(ct):
     ct.remove_bus({845})
 
 
-def test_add_demand_flexibility(ct):
+def test_add_demand_flexibility(ct, monkeypatch):
     with pytest.raises(ValueError):
         # Fails because "demand_flexibility_dn", a required key, is not included
         ct.add_demand_flexibility(
@@ -540,15 +538,18 @@ def test_add_demand_flexibility(ct):
             }
         )
 
+    monkeypatch.setattr(Context, "get_data_access", MockContext().get_data_access)
+    data_access = Context.get_data_access()
+
     # Create fake files in the expected directory path
-    exp_path = os.path.join(Path.home(), "ScenarioData", "raw", str(grid.grid_model))
-    dir_exists_prev = True
-    if not os.path.isdir(exp_path):
-        os.makedirs(exp_path)
-        dir_exists_prev = False
-    fake_df = pd.DataFrame()
-    fake_df.to_csv(os.path.join(exp_path, "demand_flexibility_up_Test.csv"))
-    fake_df.to_csv(os.path.join(exp_path, "demand_flexibility_dn_Test.csv"))
+    exp_path = f"raw/{grid.grid_model}"
+
+    for csv_file in (
+        "demand_flexibility_up_Test.csv",
+        "demand_flexibility_dn_Test.csv",
+    ):
+        with data_access.write(exp_path + "/" + csv_file) as f:
+            pd.DataFrame().to_csv(f)
 
     # Add a test instance of demand flexibility to the change table
     ct.add_demand_flexibility(
@@ -566,10 +567,3 @@ def test_add_demand_flexibility(ct):
         },
     }
     assert ct.ct == exp_dict
-
-    # Delete the created directory and fake data
-    if dir_exists_prev:
-        os.remove(os.path.join(exp_path, "demand_flexibility_up_Test.csv"))
-        os.remove(os.path.join(exp_path, "demand_flexibility_dn_Test.csv"))
-    else:
-        shutil.rmtree(os.path.join(Path.home(), "ScenarioData"))
--- a/powersimdata/input/tests/test_change_table.py
+++ b/powersimdata/input/tests/test_change_table.py
@@ -1,3 +1,8 @@
+import os
+import shutil
+from pathlib import Path
+
+import pandas as pd
 import pytest
 
 from powersimdata.input.change_table import ChangeTable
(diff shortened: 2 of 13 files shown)


powersimdata/data_access/scenario_list.py
```diff
@@ -75,7 +75,7 @@ def add_entry(self, scenario_info):
         scenario_info.move_to_end("id", last=False)
         table.reset_index(inplace=True)
         entry = pd.DataFrame({k: [v] for k, v in scenario_info.items()})
-        table = table.append(entry)
+        table = pd.concat([table, entry])
```
danielolsen, line 78: What's the purpose of this change?

lanesmith: When I was testing the functionality, there was a deprecation warning indicating that `append` will be deprecated in a future release of `pandas` and `concat` should be used instead

danielolsen: Makes sense, thanks for being proactive on this.

powersimdata/scenario/analyze.py
```diff
@@ -226,6 +228,33 @@ def get_load_shed(self):
 
         return load_shed
 
+        def get_load_shift_up(self):
```
jon-hagg, line 231: FYI based on #592 this method would change to just be `return self._get_data("LOAD_SHIFT_UP")`. Similar for `get_load_shift_dn`.

lanesmith: Done.

powersimdata/input/transform_profile.py
```diff
@@ -117,6 +117,48 @@ def _get_demand_profile(self):
...
+        area_indicator = [1 if "zone." in x else 0 for x in df.columns]
+        if len(area_indicator) == sum(area_indicator):
+            # Demand flexibility profile only contains zones; prune accordingly
+            zone_id = sorted(self.grid.bus.zone_id.unique())
+            zone_id = [f"zone.{x}" for x in zone_id]
+            df = df.loc[:, df.columns.isin(zone_id)]
+        else:
+            bus_id = sorted(self.grid.bus.bus_id.unique())
+            bus_id = [str(x) for x in bus_id]
+            if len(area_indicator) > 0:
```
jon-hagg, line 144: Should this be `if sum(area_indicator) > 0`?

lanesmith: Yes, I'll fix that.

jon-hagg: Related - I think this conditional block could be collapsed to 1 layer deep:

```
cols = []
if sum(area_indicator) > 0:
    # append zone_id to cols
if sum(area_indicator) < len(area_indicator):
    # append bus_id to cols
df = df.loc[:, df.columns.isin(cols)]
```

lanesmith: Yeah, that's cleaner. Thanks! I'll be sure to implement that.

powersimdata/input/tests/test_change_table.py
```diff
@@ -492,3 +510,66 @@ def test_remove_bus(ct):
         ct.remove_bus({845})
     ct.scale_plant_capacity(resource="ng", plant_id={0: 0})
     ct.remove_bus({845})
+
+
+def test_add_demand_flexibility(ct):
+    with pytest.raises(Exception):
```
jon-hagg, line 516: This would raise `ValueError` I think? If so we could swap `Exception` for that - usually good to pick the most specific one.

lanesmith: Oh okay. I can try to find the more specific error that will be raised for each of those.

powersimdata/input/tests/test_change_table.py
```diff
@@ -492,3 +510,66 @@ def test_remove_bus(ct):
...
+    )
+    exp_dict = {
+        "demand_flexibility": {
+            "demand_flexibility_up": "Test",
+            "demand_flexibility_dn": "Test",
+            "demand_flexibility_duration": 6,
+        },
+    }
+    assert ct.ct == exp_dict
+
```
jon-hagg, line 569: I know this is probably safe, but also think we could use a temporary folder instead. Commenting to remind myself to do that later

lanesmith: Do you mean instead of having it in the `ScenarioData` folder? If so, I originally tried to do that, but `get_profile_version_local` looks for the local directory, which is set to a `ScenarioData` folder in the user's home directory.

powersimdata/input/transform_profile.py
lanesmith: Done.

powersimdata/input/tests/test_change_table.py
jon-hagg: Yeah, it wasn't trivial to do but I pushed a branch [here](https://github.com/Breakthrough-Energy/PowerSimData/compare/lane/demand_flex...jon/demand_flex?expand=1) that I think would work. The main aspect is that we mock the `get_data_access` method so that it returns an instance which uses a `TempFS` instead of the local filesystem. If it looks good I can make a PR to your branch.

lanesmith: I'm not as well-versed with the `fs` package or with some of the data access functionality, but it looks good to me. The only other thing I can see is that a similar change would need to be made to the test I added in `test_change_table.py` to get rid of some of the similar file and directory creation.

jon-hagg: Ah thanks, I missed that one, will do.

danielolsen:
I'm able to use this to create and prepare a scenario...
```python
scenario.set_grid("usa_tamu", "Texas")
scenario.set_time("2016-01-01 00:00", "2016-01-01 23:00", "24H")
scenario.set_base_profile("demand", "vJan2021")
scenario.set_base_profile("solar", "vJan2021")
scenario.set_base_profile("wind", "vJan2021")
scenario.set_base_profile("hydro", "vJan2021")
scenario.set_name("foo", "bar2")
scenario.change_table.add_demand_flexibility(
    {
        "demand_flexibility_up": "EFS_Medium_Moderate_Enhanced_2030_Transportation",
        "demand_flexibility_dn": "EFS_Medium_Moderate_Enhanced_2030_Transportation",
    }
)
scenario.create_scenario()
scenario.prepare_simulation_input()
```
...but then when I try to run this Scenario manually using REISE.jl, I end up getting an error:
```julia
julia> import REISE

julia> import Gurobi

julia> REISE.run_scenario_gurobi(; interval=24, n_interval=1, start_index=1, inputfolder=".")
Compute Server job ID: 5266459b-2415-4adb-93a1-37ef15f46091
Capacity available on 'https://ip-10-0-62-149:61000' - connecting...
Established HTTPS encrypted connection
Reading from folder: .
...loading case.mat
...loading demand.csv
...loading hydro.csv
...loading wind.csv
...loading solar.csv
File case_storage.mat not found in .
...loading demand flexibility parameters
The demand flexibility duration parameter is not defined. Will default to being the size of the interval.
The parameter that indicates if the rolling load balance constraint is enabled is not defined. Will default to being enabled.
...loading demand flexibility up profiles
Demand flexibility up-shift cost profiles not found in .. Will default to no cost for up-shifting demand.
...loading demand flexibility dn profiles
Demand flexibility dn-shift cost profiles not found in .. Will default to no cost for dn-shifting demand.

Compute Server communication statistics:
  Sent: 0.001 MBytes in 4 msgs and 0.23s (0.00 MB/s)
  Received: 0.000 MBytes in 1 msgs and 0.16s (0.00 MB/s)

Connection to Gurobi closed successfully!
ERROR: MethodError: Cannot `convert` an object of type InlineStrings.String7 to an object of type Bool
Closest candidates are:
  convert(::Type{T}, ::T) where T<:Number at number.jl:6
  convert(::Type{T}, ::Number) where T<:Number at number.jl:7
  convert(::Type{T}, ::Ptr) where T<:Integer at pointer.jl:23
  ...
Stacktrace:
 [1] REISE.DemandFlexibility(::DataFrames.DataFrame, ::DataFrames.DataFrame, ::Nothing, ::Nothing, ::Int64, ::Bool, ::InlineStrings.String7, ::Bool) at C:\Users\DanielOlsen\repos\bes\REISE.jl\src\types.jl:45
 [2] REISE.DemandFlexibility(; flex_amt_up::DataFrames.DataFrame, flex_amt_dn::DataFrames.DataFrame, cost_dn::Nothing, cost_up::Nothing, duration::Int64, enabled::Bool, interval_balance::InlineStrings.String7, rolling_balance::Bool) at .\util.jl:438
 [3] read_demand_flexibility(::String, ::Int64) at C:\Users\DanielOlsen\repos\bes\REISE.jl\src\read.jl:255
 [4] run_scenario(; num_segments::Int64, interval::Int64, n_interval::Int64, start_index::Int64, inputfolder::String, outputfolder::Nothing, threads::Nothing, optimizer_factory::Gurobi.Env, solver_kwargs::Dict{String,Int64}, model_kwargs::Nothing) at C:\Users\DanielOlsen\repos\bes\REISE.jl\src\REISE.jl:69
 [5] run_scenario_gurobi(; solver_kwargs::Nothing, kwargs::Base.Iterators.Pairs{Symbol,Any,NTuple{4,Symbol},NamedTuple{(:interval, :n_interval, :start_index, :inputfolder),Tuple{Int64,Int64,Int64,String}}}) at C:\Users\DanielOlsen\repos\bes\REISE.jl\src\solver_specific\gurobi.jl:13
 [6] top-level scope at REPL[4]:1
```

Looking a little closer at what's going on under the hood, it seems that `demand_flexibility["interval_balance"]` is `"True"` (a string) rather than `true` (a bool). Is this a bug, or is there something that I'm doing wrong here?

lanesmith:
> Looking a little closer at what's going on under the hood, it seems that `demand_flexibility["interval_balance"]` is `"True"` (a string) rather than `true` (a bool). Is this a bug, or is there something that I'm doing wrong here?

Based on what you describe, I believe this is an issue with the `CSV` package in Julia. I was encountering a similar issue where "True" was being read as a string instead of as a bool. Based on the available options within `CSV.File`, this shouldn't be an issue as most reasonable variations of "true" (e.g., True, true, TRUE) are by default supposed to be parsed as bools. In my case, the problem only seemed to occur when a parameter that had "True" or "False" was listed in the last column of the demand flexibility parameters file. The interesting thing is that it recognizes that other "True" values are supposed to be bools, just not values in the last column. I thought I found a workaround by moving the duration parameter to the last column, which was working for my local tests. I guess not having the duration parameter loses that safeguard though.

There doesn't seem to be a better option for loading .csv files into DataFrames within Julia (at least that I'm aware of). I could make a fix in `REISE.jl` where the value for the boolean demand flexibility parameters are checked and switched to bools if provided as a reasonable string (i.e., any variation of "True" or "False"). I'm also open to other solutions.

lanesmith:
Another potential solution could be changing how I create the demand flexibility parameters file in `execute.py`. I could make it so that all the parameters are always created (e.g., `rolling_balance` and `duration` are also created even when `demand_flexibility_duration` isn't provided in the change table), they just would be set to appropriate values (e.g., `rolling_balance` would be False and `duration` would be a throwaway value such as 0 (the value of `duration` would not matter if `rolling_balance` isn't enabled)). This would probably be the quickest fix and would prevent needing to prop up the `CSV` package.

danielolsen:
`CSV` had been at 0.9.10, but upgrading to 0.10.3 fixed it, so I guess the problem isn't with this PR, but we should probably specify the CSV version within REISE.jl

For reference, the contents of my `demand_flexibility_parameters.csv` file were:
```
,enabled,interval_balance
0,True,True
```

danielolsen (approved):
I haven't looked too deeply at the implementation, but I can confirm that it works and makes files that REISE.jl likes. Thanks!

lanesmith:
> `CSV` had been at 0.9.10, but upgrading to 0.10.3 fixed it, so I guess the problem isn't with this PR, but we should probably specify the CSV version within REISE.jl.

Okay, glad to hear that a later version of `CSV` fixes this. Unless there is any opposition, I think I'll still make the change to `execute.py` that I suggested above. This will make it so the demand flexibility parameters are more consistent across scenarios (i.e., all parameters will always be present, even if they're not enabled)

danielolsen:
@lanesmith I trust you.

PR #592, opened 2022-03-01
Remove unused variable and small refactor to reuse shared code

jon-hagg:

### Purpose
I noticed we have two things called `data_loc` - one is an actual filepath which is used to load a grid, the other one was to allow loading scenarios that were moved to a backup disk. The latter is not needed, but if it was, we could simply change `DATA_ROOT_DIR` for any scenario that was moved. 

### What the code is doing
Remove usages of `data_loc` (the latter ones), and some cleanup in `analyze.py` which was just low hanging fruit.

### Time estimate
5 min


BainanXia (approved):
Good catch!

PR #593, opened 2022-03-02
Use temp filesystem for profile test

jon-hagg:

### Purpose
I wasn't sure how this would work so did some experimenting, made a few tweaks, etc. This makes it so the mocked `DataAccess` instance (or really, any instance) will use the `self.local_fs` attribute to get local profiles, instead of having that be hard coded to use `LOCAL_DIR`. In this case, the local_fs is a `TempFS` so it's completely isolated from existing files. 

### What the code is doing
See above

### Testing
tox

### Time estimate
5-10 min

### DIFF
--- a/powersimdata/data_access/profile_helper.py
+++ b/powersimdata/data_access/profile_helper.py
@@ -1,51 +1,20 @@
-import fs
+def get_profile_version(_fs, grid_model, kind):
+    """Returns available raw profile from the given filesystem
 
-from powersimdata.utility import server_setup
-
-
-def _get_profile_version(_fs, kind):
-    """Returns available raw profiles from the given filesystem
     :param fs.base.FS _fs: filesystem instance
+    :param str grid_model: grid model.
     :param str kind: *'demand'*, *'hydro'*, *'solar'*, *'wind'*,
         *'demand_flexibility_up'*, *'demand_flexibility_dn'*,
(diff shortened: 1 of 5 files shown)


powersimdata/data_access/profile_helper.py
```diff
@@ -30,22 +29,8 @@ def get_profile_version_cloud(grid_model, kind):
...
-    :param str kind: *'demand'*, *'hydro'*, *'solar'*, *'wind'*,
-        *'demand_flexibility_up'*, *'demand_flexibility_dn'*,
-        *'demand_flexibility_cost_up'*, or *'demand_flexibility_cost_dn'*.
-    :return: (*list*) -- available profile version.
-    """
-    profile_dir = fs.path.join(server_setup.LOCAL_DIR, "raw", grid_model)
-    lfs = fs.open_fs(profile_dir, create=True)
-    return _get_profile_version(lfs, kind)
+    bfs = fs.open_fs("azblob://besciences@profiles")
+    return get_profile_version(bfs, grid_model, kind)
```
lanesmith, line 33: Is there a reason this still needs to be its own function? Could the contents of `bfs` just be passed into `get_profile_version` in the `blob_version` located in `data_access.py`?

jon-hagg: Good call, done.

lanesmith (approved):
Looks good to me. Thanks for doing this!

Issue #467, opened 2021-04-27
Refactor ChangeTable scaling methods so that they create ValueErrors on bad input


danielolsen:
Currently, if we try to scale a grid element based on a bad zone name or plant ID, **all** scaling for that grid element is removed, and an error message is printed. I think it would be more intuitive if a ValueError were raised, and any previous scaling were maintained. This would be consistent with the behavior when we add elements, and I believe that we discussed doing the same for scaling, but it fell off the priority list.

PR #594, opened 2022-03-03
refactor: separate add_storage_capacity

jon-hagg:

### Purpose
Move `add_storage_capacity` to its own module. If we like this pattern, I would do the same thing for most of the other change table operations. 

### What the code is doing
Move the logic to a separate function that takes a change table instance, so the actual method on the `ChangeTable` object becomes a one liner.

### Testing
tox

### Time estimate
5 min

### DIFF
--- a/powersimdata/input/change_table.py
+++ b/powersimdata/input/change_table.py
@@ -5,6 +5,8 @@ from powersimdata.design.transmission.upgrade import (
     scale_congested_mesh_branches,
     scale_renewable_stubs,
 )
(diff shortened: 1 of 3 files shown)


danielolsen (approved):
Makes sense. It would be nice to take care of #467 at the same time as necessary (doesn't apply here, but does to other change table methods).

jon-hagg:
> Makes sense. It would be nice to take care of #467 at the same time as necessary (doesn't apply here, but does to other change table methods).

I'll try it out when I come across those methods

PR #596, opened 2022-03-10
Add electrification profiles to change table

jon-hagg:
### Purpose
Define/implement new schema for electrification profiles and scaling by grid/zone. 

### What the code is doing
Define one new function `add_electrification` which can specify profiles and scaling for supported types: currently these are building and transportation. If specifying scaling factors by zone, the zone name is checked against the current grid object. There isn't yet validation for the profile name/version, but I'll add that if the general approach looks good. 

Depending on expected usage patterns, the schema could also be "inverted" by mapping profiles to zones and scaling factors, but I think either way has trade offs in how easy it is to specify the change table and how much code is required to aggregate the profiles during the preparation phase. 

### Testing
New unit tests.

### Usage Example/Visuals
The tests are the best usage example. The full schema will look like:
```
{
    "building": {
        "grid": [scale_factors],
        "zone": { zone_name: [scale_factors], ... }
    },
    "transportation": {...}
}
```
Where the `scale_factors` dict has the structure:
```
{profile_name_and_version_1: s_1, profile_name_and_version_2: s_2, ...} 
```
And `0 <= s_i <= 1`

### Time estimate
20 min

### DIFF
--- a/powersimdata/input/change_table.py
+++ b/powersimdata/input/change_table.py
@@ -523,7 +523,10 @@ class ChangeTable:
                     raise ValueError(f"Available {k} profiles: {', '.join(possible)}")
 
     def add_electrification(self, kind, info):
-        """TODO"""
+        """Add profiles and scaling factors for electrified demand.
+
+        See :func:`powersimdata.input.changes.electrification.add_electrification`
+        """
         add_electrification(self, kind, info)
 
     def add_dcline(self, info):
--- a/powersimdata/input/change_table.py
+++ b/powersimdata/input/change_table.py
@@ -6,6 +6,7 @@ from powersimdata.design.transmission.upgrade import (
     scale_renewable_stubs,
 )
 from powersimdata.input.changes.bus import add_bus, remove_bus
+from powersimdata.input.changes.electrification import add_electrification
 from powersimdata.input.changes.helpers import ordinal
 from powersimdata.input.changes.plant import add_plant, remove_plant, scale_plant_pmin
 from powersimdata.input.changes.storage import add_storage_capacity
@@ -520,6 +521,10 @@ class ChangeTable:
                 else:
                     del self.ct["demand_flexibility"]
                     raise ValueError(f"Available {k} profiles: {', '.join(possible)}")
+
+    def add_electrification(self, kind, info):
+        """TODO"""
+        add_electrification(self, kind, info)
 
     def add_dcline(self, info):
         """Adds HVDC line(s).
--- a/powersimdata/input/changes/electrification.py
+++ b/powersimdata/input/changes/electrification.py
@@ -8,6 +8,8 @@ def _check_scale_factors(scale_factors):
 
     :param dict scale_factors: see :func:`add_electrification`
     """
+    if not isinstance(scale_factors, dict):
+        raise ValueError("scale factors must be a dict")
     if not all(isinstance(k, str) for k in scale_factors):
         raise ValueError("profile name must be str")
     if not all(isinstance(d, (int, float)) for d in scale_factors.values()):
@@ -21,9 +23,9 @@ def _check_scale_factors(scale_factors):
 
 @dataclass
 class ScaleFactors:
-    """Map profile(s) to adoption rate
+    """Map technology to adoption rate
 
-    :param dict sf: a dictionary mapping profiles names to adoption rate
+    :param dict sf: a dictionary mapping tech to adoption rate
     """
 
     sf: Dict[str, float]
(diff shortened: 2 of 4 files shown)


rouille:
Not sure I understand why you constrain the scaling factors to be less than 1

danielolsen:
> Not sure I understand why you constrain the scaling factors to be less than 1

We can't change more than 100% of e.g. fossil fuel heating to electrified heating, and the profiles that we're getting from the collaborators represent 100% conversion of fossil end uses to 100% of one particular technology.

BainanXia:
> Not sure I understand why you constrain the scaling factors to be less than 1

The scaling factor here is essentially adoption rate.

danielolsen:
I like the examples shown in the tests. Being able to scale across the whole grid is a nice addition!

rouille:
> I like the examples shown in the tests. Being able to scale across the whole grid is a nice addition!

Should we implement it for other methods?

powersimdata/input/change_table.py
```diff
@@ -567,6 +568,10 @@ def add_demand_flexibility(self, info):
                     del self.ct["demand_flexibility"]
                     raise ValueError(f"Available {k} profiles: {', '.join(possible)}")
 
+    def add_electrification(self, kind, info):
+        """TODO"""
```
jon-hagg, line 572: How do we want to do docstrings? Currently for example, in `scale_renewable_stubs` we are using (3). 
1. both in `change_table.py` and the imported function
2. only in `change_table.py`, and put a link in the docstring for the imported function 
3. only in the imported function, and put a link to that module here
4. something else?

danielolsen: I like (3) the best, followed by (1).

BainanXia: I think we either do both (1) or only in the function that being called first (I guess this is equivalent to 3?).

rouille: I like 3. This is what you did in #604

powersimdata/input/changes/electrification.py
```diff
@@ -0,0 +1,58 @@
+import copy
+
+
+def _check_scale_factors(scale_factors):
```
rouille, line 4: What about docstrings for private functions?

powersimdata/input/changes/electrification.py
```diff
@@ -0,0 +1,67 @@
...
+        raise ValueError("scaling factors must sum to between 0 and 1")
+
+
+def _check_zone_scaling(obj, info):
+    """Validate schema for zone scaling
+
+    :param powersimdata.input.change_table.ChangeTable obj: change table
+    :param dict info: see :func:`add_electrification`
+    """
+    if not all(isinstance(k, str) for k in info):
```
danielolsen, line 26: Should we check that `info` is a dict?

danielolsen:
Looking at this, I realize that I specified this feature incompletely. I think we need one more layer of nesting in the change table schema, because we have multiple end uses within each of buildings and transportation. So we would go from `[buildings/transportation][zone/grid][technology] = adoption` to `[buildings/transportation][zone/grid][end_use][technology] = adoption`. Currently we have:
```python
info = {
    "New York City": {"advanced_heat_pump_v2": 0.7},
    "Western North Carolina": {
        "standard_heat_pump_v1": 0.5,
        "advanced_heat_pump_v2": 0.5,
    },
}
```
But I think we need something more like:
```python
info = {
    "New York City": {
        "res_heating": {"advanced_heat_pump_v2": 0.7},
        "com_hot_water": {"advanced_heat_pump_v2": 0.7},
        "res_cooking": {"mid_efficiency": 0.5, "high_efficiency": 0.3},
    },
    "Western North Carolina": {
        "res_heating": {
            "standard_heat_pump_v1": 0.5,
            "advanced_heat_pump_v2": 0.5,
        },
    },
}
```
Adoption rates for various technologies still need to sum to less than one for a given `(class, zone, end use)`. Sorry for the bad spec.

powersimdata/input/changes/electrification.py
```diff
@@ -0,0 +1,121 @@
...
+
+    :param dict info: a mapping from end use (*str*) to scale factors (*dict*)
+    :raises ValueError: if info is not a dict, or any keys are not strings
+    """
+
+    end_uses: Dict[str, ScaleFactors]
+
+    def __init__(self, info):
+        if not isinstance(info, dict):
+            raise ValueError("scale factors must be a dict")
```
rouille, line 51: What do you think of `zone/grid scaling must be a dict` since at this point `info` is a `dict` with keys being zone/grid?

jon-hagg: Done

PR #599, opened 2022-03-11
Remove issue and pull request template so default takes effect

kasparm:
### Purpose
Remove issue and pull request template so the default from .github repo takes effect

### Time estimate
5min


rouille (approved):
Thanks

PR #600, opened 2022-03-11
Add import_data module with pypsa importer

FabianHofmann:
### Purpose
Allow to create Grid objects from PyPSA networks. The PR also introduces some changes outside of the newly introduced `import_data` module. This are: 

*  The `Grid` class has a new attribute `SUPPORTED_IMPORTS` used in the `Grid` initialization process. The check in the beginning of the initialization that the `source` argument is in `SUPPORTED_MODELS` was extended to a check that `source` is in `SUPPOERTED_MODELS` or in `SUPPERTED_IMPORTS`. Further, when `data_loc` of the imported data is in `SUPPORTED_IMPORTS`, immutables are not loaded into the Grid. This means, as soon as `data_loc` is in the `SUPPORTED_IMPORTS` set, the model is not required to be in `SUPPORTED_MODELS`. The `SUPPORTED_IMPORTS` set is for now only containing "pypsa". 
* The `Grid` initialization skips loading from cache if the imported object is a pypsa Network.
* An alignment of the implementation introduced in #603 with the import. Since the import from pypsa relies on the same dictionary used for the export, it is helpful to include the `marginal_cost` key into the translation dict again. 

### What the code is doing
For now a new `AbstractGrid` class `FromPyPSA` was implemented. When initializing it takes a pypsa network and converts all data fields according to the `PowerSimData` convention. As discussed, storages are not yet imported from pypsa. This remains a point on the TODO list and will be covered in a separate PR. 

The code does not yet load a `Scenario` object from `PyPSA`. I do not want to bloat this PR, thus it will be introduced via a separate PR. 

### Testing
Tests for initializing a Grid object from a pypsa Network were added to the CI. 

### Where to look

* `powersimdata/input/import_data.py`
* `powersimdata/input/grid.py`

### Time estimate
~ 40 min

### DIFF
--- a/powersimdata/input/export_data.py
+++ b/powersimdata/input/export_data.py
@@ -5,7 +5,6 @@ import numpy as np
 import pandas as pd
 from scipy.io import savemat
 
-from powersimdata import Grid
 from powersimdata.input.transform_profile import TransformProfile
 
 PYPSA_AVAILABLE = True
@@ -317,6 +316,7 @@ def export_to_pypsa(
         generators to the exported pypsa network. This ensures feasibility when
         optimizing the exported pypsa network as is. The default is True.
     """
+    from powersimdata.input.grid import Grid  # avoid circular import
     from powersimdata.scenario.scenario import Scenario  # avoid circular import
 
     if not PYPSA_AVAILABLE:
--- a/powersimdata/input/export_data.py
+++ b/powersimdata/input/export_data.py
@@ -450,7 +450,6 @@ def export_to_pypsa(
     transformers = branches.query(
         "branch_device_type in ['TransformerWinding', 'Transformer']"
     )
-    transformers = transformers.drop(columns="branch_device_type")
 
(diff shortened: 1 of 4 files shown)


danielolsen:
The `Texas` interconnect doesn't have any HVDC lines within it, you might want to use a `Western` grid to test the conversion of that table.

danielolsen:
I don't think there is a meaningful `data_loc` or cache key for Grids converted from a PyPSA Network object. The only thing I can think of would be a pointer to where the object is in memory, but since a Network object is mutable I don't think we can assume that subsequent conversions of the same object will have the same output.

danielolsen:
I think we will eventually want to be able to convert a Network object to a Scenario object, yeah? If so, is this current design compatible with that next step?

FabianHofmann:
@danielolsen I am closing this in the favor of #608 which also includes all of the above commits.

PR #604, opened 2022-03-15
Factor out change table methods

jon-hagg:

### Purpose
Split up change table methods into separate modules/functions. 

### What the code is doing
Add modules for `plant` and `bus` and move (some) associated methods there. Docstrings link to the function implementation as discussed in #596.

### Testing
tox

### Time estimate
10 min

### DIFF
--- a/powersimdata/input/changes/plant.py
+++ b/powersimdata/input/changes/plant.py
@@ -8,6 +8,7 @@ _renewable_resource = {"hydro", "solar", "wind", "wind_offshore"}
 def add_plant(obj, info):
     """Sets parameters of new generator(s) in change table.
 
+    :param powersimdata.input.change_table.ChangeTable obj: change table
     :param list info: each entry is a dictionary. The dictionary gathers
         the information needed to create a new generator.
         Required keys: "bus_id", "Pmax", "type".
@@ -65,6 +66,7 @@ def add_plant(obj, info):
 def remove_plant(obj, info):
     """Remove one or more plants.
 
+    :param powersimdata.input.change_table.ChangeTable obj: change table
     :param int/iterable info: iterable of plant indices, or a single plant index.
     :raises ValueError: if ``info`` contains one or more entries not present in the
         plant table index.
--- a/powersimdata/input/changes/plant.py
+++ b/powersimdata/input/changes/plant.py
@@ -3,7 +3,7 @@ import copy
 from powersimdata.input.transform_grid import TransformGrid
 from powersimdata.utility.distance import find_closest_neighbor
 
-_renewable_resource = {"hydro", "solar", "wind", "wind_offshore"}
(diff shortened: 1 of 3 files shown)


powersimdata/input/changes/plant.py
```diff
@@ -0,0 +1,117 @@
+import copy
+
+from powersimdata.input.transform_grid import TransformGrid
+from powersimdata.utility.distance import find_closest_neighbor
+
+_renewable_resource = {"hydro", "solar", "wind", "wind_offshore"}
```
rouille, line 6: I am wondering if there is not a way to use the `grid.model_immutables.plants.x` where `x` is one of the `set` defined in `powersimdata.network.usa_tamu.constants.plants` (could be network independent in the future). Unfortunately the `renewable_resources` defined there is different than the one defined here. What do you think @BainanXia and @danielolsen?

danielolsen: It seems like for the purposes of the change table, this set is 'the resources which have profiles', such that when we try to add a new one, we find the closest profile to grab as the profile for the new plant. It doesn't seem like 'renewable' in the constants really means anything within this repo, but it's used in PostREISE to represent 'resources which have profile AND are curtailable'.

BainanXia: Yeah, in change table/grid modification, we care more about whether certain entities have profiles associated with them that we need to deal with, which causes the confusion. Probably we should rename it into something like `resources_with_profiles`?

danielolsen: The one in `powersimdata.network.usa_tamu.constants.plants` could then be changed to `curtailable_profile_resources`, or something along those lines.

rouille: To avoid any issue in PostREISE, we could create a new set in `powersimdata.network.usa_tamu.constants.plants` named `curtailable_profile_resources` and set to `{"hydro", "solar", "wind", "wind_offshore"}` that would be accessed in the change table through `obj.model_immutables.plants.curtailable_profile_resources` in `add_plant`

danielolsen: The problem is that `hydro` is a profile resource, but not a curtailable one, so we need two different sets.

danielolsen:
Are we planning to move some of the line methods out as well?

jon-hagg:
> Are we planning to move some of the line methods out as well?

Yeah I just figured I'd break it up into a few smaller PRs.

powersimdata/input/changes/plant.py
rouille: @jon-hagg, it would be more accurate to rename `_renewable_resource` to `profile_resource`. Can you confirm @BainanXia and @danielolsen.

jon-hagg: Done (in plant.py only)

rouille (approved):
Thanks

PR #606, opened 2022-03-24
Fix some of the pandas deprecation warnings

jon-hagg:
### Purpose
Pandas has been yelling at me whenever I run tests because we are still using `.append` instead of `.concat`. This PR should make pandas happier. 

### What the code is doing
Change some of the append to concat. It defaults to `sort=False` (see [docs](https://pandas.pydata.org/docs/reference/api/pandas.concat.html)) so I removed that where applicable. 

### Testing
tox

### Time estimate
10 min

### DIFF
--- a/powersimdata/input/tests/test_grid.py
+++ b/powersimdata/input/tests/test_grid.py
@@ -405,10 +405,8 @@ def test_grid_eq_failure_storage(base_texas):
     test_grid = copy.deepcopy(base_texas)
     gencost = {g: 0 for g in test_grid.storage["gencost"].columns}
     gen = {g: 0 for g in test_grid.storage["gen"].columns}
-    test_grid.storage["gencost"] = test_grid.storage["gencost"].append(
-        gencost, ignore_index=True
-    )
-    test_grid.storage["gen"] = test_grid.storage["gen"].append(gen, ignore_index=True)
+    test_grid.storage["gencost"].loc[0, :] = gencost
+    test_grid.storage["gen"].loc[0, :] = gen
     assert base_texas != test_grid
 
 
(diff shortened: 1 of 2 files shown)


powersimdata/input/tests/test_grid.py
```diff
@@ -405,10 +405,8 @@ def test_grid_eq_failure_storage(base_texas):
     test_grid = copy.deepcopy(base_texas)
     gencost = {g: 0 for g in test_grid.storage["gencost"].columns}
     gen = {g: 0 for g in test_grid.storage["gen"].columns}
-    test_grid.storage["gencost"] = test_grid.storage["gencost"].append(
-        gencost, ignore_index=True
-    )
-    test_grid.storage["gen"] = test_grid.storage["gen"].append(gen, ignore_index=True)
+    test_grid.storage["gencost"].loc[0, :] = gencost
+    test_grid.storage["gen"].loc[0, :] = gen
```
BainanXia, line 409: Will this work the same as `pd.concat`?

jon-hagg: Good point. This assumes that the dataframes are empty, which happens to be true in this case.

BainanXia: For some reason I tried with a simple example and observed following:
image

jon-hagg: Yeah this one works because the data frames are "empty" but initialized with columns, so we can assign a dict to a row. My guess was that we have a convention of creating an empty `grid.storage` and adding to it later, but I could generalize it if that's preferred.

BainanXia: Ah, wasn't paying attention to the `dict`, it's the content of one row rather than a multiple rows. That should work.

BainanXia (approved):
Thanks for taking care of this.

PR #603, opened 2022-03-15
refactor: calculate marginal cost via linearized quadratic

danielolsen:
[Pull Request doc](https://breakthrough-energy.github.io/docs/user/git_guide.html#d-pull-request)

### Purpose
When exporting a `Scenario`/`Grid` to a PyPSA Network, calculate the generators' marginal cost using a linearized quadratic (from Pmin to Pmax), rather than just the linear component of the quadratic curve.

### What the code is doing
For generators where `Pmin != Pmax`, the expression `[(a*Pmax^2 + b*Pmax + c) - (a*Pmin^2 + b*Pmin + c)]/(Pmax - Pmin)` simplifies to `b + a * (Pmax + Pmin)`, so we build the `marginal_cost` column based on this expression. For the edge case where `Pmin = Pmax` (e.g. if we've scaled generators down to 0), the existing linear term will still be used (since `pandas.Series.combine_first` only fills null values or values that aren't present in the first Series' index).

### Testing
Tested manually:
```python
from powersimdata import Scenario
from powersimdata.input.export_data import export_to_pypsa
scenario = Scenario()
scenario.set_grid("usa_tamu", "Texas")
scenario.set_base_profile("demand", "vJan2021")
scenario.set_base_profile("hydro", "vJan2021")
scenario.set_base_profile("solar", "vJan2021")
scenario.set_base_profile("wind", "vJan2021")
scenario.set_time("2016-01-01 00:00", "2016-01-01 23:00", "24H")
network = export_to_pypsa(scenario)
network.lopf(solver_name="gurobi", pyomo=False)
```

### Time estimate
10 minutes.

PR #608, opened 2022-03-25
Pypsa importer for scenario objects

FabianHofmann:
### Purpose
Allow to create Grid or Scenario objects from PyPSA networks. The PR also introduces some changes outside of the newly introduced `import_data` module. This are: 

*  The `Grid` class has a new attribute `SUPPORTED_IMPORTS` used in the `Grid` initialization process. The check in the beginning of the initialization that the `source` argument is in `SUPPORTED_MODELS` was extended to a check that `source` is in `SUPPORTED_MODELS` or in `SUPPORTED_IMPORTS`. Further, when `data_loc` of the imported data is in `SUPPORTED_IMPORTS`, immutables are not loaded into the Grid. This means, as soon as `data_loc` is in the `SUPPORTED_IMPORTS` set, the model is not required to be in `SUPPORTED_MODELS`. The `SUPPORTED_IMPORTS` set is for now only containing "pypsa". 
* The `Grid` initialization skips loading from cache if the imported object is a pypsa Network.
* An alignment of the implementation introduced in #603 with the import. Since the import from pypsa relies on the same dictionary used for the export, it is helpful to include the `marginal_cost` key into the translation dict again. 
* The `scenario` module has a new field `SUPPORTED_IMPORTS` which lists all supported imports 
* The in the `Scenario` initialization the `is_importable` function is used to distinguish a descriptor that can be imported and converted to a Scenario state.  

### What the code is doing
The code changes allow to create a Grid/Scenario object from arbitrary PyPSA networks. With this changes, a PyPSA network can be passed as an argument to `Grid` and `Scenario` initialization. It was tried to make the changes as modular as possible and to create a structure that allows to add further imports/exports conversions from/to other packages in the future.

A new `AbstractGrid` class `FromPyPSA` was implemented. When initializing it takes a pypsa network and converts all data fields according to the `PowerSimData` convention. As discussed, storages are not yet imported from pypsa. This remains a point on the TODO list and will be covered in a separate PR. 

A `AnalyzePypsa` class was implemented which is a subclass of `Analyze`.  It mimics the behavior of the `Analyze` state while keeping all necessary data in memory. 

### Testing
For testing the PyPSA to Grid conversion, a test script was added. 

For testing the PyPSA to Scenario conversion, the following script was used.

```python 
from powersimdata.input.export_data import export_to_pypsa


ref = Scenario()
ref.set_grid(grid_model="usa_tamu", interconnect='Texas')
ref.set_name("test", "dummy")
ref.set_time("2016-08-01 00:00:00", "2016-08-31 23:00:00", "24H")
ref.set_base_profile("demand", "vJan2021")
ref.set_base_profile("hydro", "vJan2021")
ref.set_base_profile("solar", "vJan2021")
ref.set_base_profile("wind", "vJan2021")


kwargs = dict(add_substations=True, add_load_shedding=False, add_all_columns=True)
n = export_to_pypsa(ref, **kwargs)


test = Scenario(n)

assert_frame_equal(test.get_hydro(), ref.get_hydro())
assert_frame_equal(test.get_solar(), ref.get_solar())
assert_frame_equal(test.get_wind(), ref.get_wind())
assert_frame_equal(test.get_bus_demand(), ref.get_bus_demand())
assert_frame_equal(test.get_demand(), ref.get_demand())


# Solved
kwargs = dict(add_substations=False, add_load_shedding=True)
n = export_to_pypsa(ref, **kwargs)
n.set_snapshots(n.snapshots[:24])
n.lopf(pyomo=False, solver_name="gurobi")

solved = Scenario(n)
assert not solved.get_lmp().empty
assert not solved.get_congu().empty
assert not solved.get_pg().empty
assert not solved.get_pf().empty
```
This code should work without assertion errors.

* What files are most important?
`import_data.py`

### Time estimate
~40 min

### DIFF
--- a/powersimdata/input/export_data.py
+++ b/powersimdata/input/export_data.py
@@ -5,7 +5,6 @@ import numpy as np
 import pandas as pd
 from scipy.io import savemat
 
-from powersimdata import Grid
 from powersimdata.input.transform_profile import TransformProfile
 
 PYPSA_AVAILABLE = True
@@ -317,6 +316,7 @@ def export_to_pypsa(
         generators to the exported pypsa network. This ensures feasibility when
         optimizing the exported pypsa network as is. The default is True.
     """
+    from powersimdata.input.grid import Grid  # avoid circular import
     from powersimdata.scenario.scenario import Scenario  # avoid circular import
 
     if not PYPSA_AVAILABLE:
--- a/powersimdata/input/export_data.py
+++ b/powersimdata/input/export_data.py
@@ -450,7 +450,6 @@ def export_to_pypsa(
     transformers = branches.query(
         "branch_device_type in ['TransformerWinding', 'Transformer']"
     )
-    transformers = transformers.drop(columns="branch_device_type")
 
     if scenario:
         lines_t = {}
--- a/powersimdata/input/export_data.py
+++ b/powersimdata/input/export_data.py
@@ -77,6 +77,7 @@ pypsa_const = {
         "rename": {
             "startup": "startup_cost",
             "shutdown": "shutdown_cost",
+            "c1": "marginal_cost",
         }
     },
     "branch": {
@@ -394,14 +395,14 @@ def export_to_pypsa(
     generators["ramp_limit_up"] = generators.ramp_limit.replace(0, np.nan)
     generators.drop(columns=drop_cols + ["ramp_limit"], inplace=True)
 
-    gencost = grid.gencost["before"]
-    gencost = gencost.rename(columns=pypsa_const["cost"]["rename"])
+    gencost = grid.gencost["before"].copy()
     # Linearize quadratic curves as applicable
     fixed = grid.plant["Pmin"] == grid.plant["Pmax"]
     linearized = gencost.loc[~fixed, "c1"] + gencost.loc[~fixed, "c2"] * (
         grid.plant.loc[~fixed, "Pmax"] + grid.plant.loc[~fixed, "Pmin"]
     )
-    gencost["marginal_cost"] = linearized.combine_first(gencost["c1"])
+    gencost["c1"] = linearized.combine_first(gencost["c1"])
+    gencost = gencost.rename(columns=pypsa_const["cost"]["rename"])
     gencost = gencost[pypsa_const["cost"]["rename"].values()]
 
     carriers = pd.DataFrame(index=generators.carrier.unique(), dtype=object)
--- a/powersimdata/input/export_data.py
+++ b/powersimdata/input/export_data.py
@@ -524,5 +524,5 @@ def export_to_pypsa(
         )
         n.add("Carrier", "load", nice_name="Load Shedding", color="red")
 
-    n.name = ", ".join(grid.interconnect)
+    n.name = ", ".join([grid.data_loc] + grid.interconnect)
     return n
--- a/powersimdata/input/export_data.py
+++ b/powersimdata/input/export_data.py
@@ -524,4 +524,5 @@ def export_to_pypsa(
         )
         n.add("Carrier", "load", nice_name="Load Shedding", color="red")
 
+    n.name = ", ".join(grid.interconnect)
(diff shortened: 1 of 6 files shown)


FabianHofmann:
@BainanXia if already tested, now, the test script should also run.

danielolsen:
> The `get_demand_per_bus` raises an errors due to the unsuccessful distribution to buses.

I'm a little confused here. If from PyPSA we have a demand profile for each bus, then when we convert to a Scenario, can we just create a zone for every bus, which contains only that bus? Then calling `Scenario.get_demand()` should return the zone demand, and calling `Scenario.get_bus_demand` should return the bus demand, which should be identical because we translate one to one from zones to buses.

danielolsen:
I'm seeing some interesting behavior when taking a Scenario that's already been run (i.e. it's in Analyze state), converting it to a Network, and then converting it back to a Scenario. First of all, calling `Scenario(network)` prints a whole lot of information that seems to relate to building a connection to Azure. Second, I get an exception when the new code tries to construct the load shedding:
```python
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\scenario\scenario.py", line 85, in __init__
    self.state = analyze_importable(descriptor)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\input\import_data.py", line 495, in analyze_importable
    return AnalyzePypsa(obj)
  File "C:\Users\DanielOlsen\repos\bes\PowerSimData\powersimdata\input\import_data.py", line 357, in __init__
    loadshed = n.generators_t.p[loadshed_gen] * n.generators.sign[loadshed_gen]
  File "C:\Python39\lib\site-packages\pandas\core\frame.py", line 3464, in __getitem__
    indexer = self.loc._get_listlike_indexer(key, axis=1)[1]
  File "C:\Python39\lib\site-packages\pandas\core\indexing.py", line 1314, in _get_listlike_indexer
    self._validate_read_indexer(keyarr, indexer, axis)
  File "C:\Python39\lib\site-packages\pandas\core\indexing.py", line 1374, in _validate_read_indexer
    raise KeyError(f"None of [{key}] are in the [{axis_name}]")
KeyError: "None of [Index(['2010001 load shedding', '2010002 load shedding',\n       '2010003 load shedding', '2010004 load shedding',\n       '2010005 load shedding', '2010006 load shedding',\n       '2010007 load shedding', '2010008 load shedding',\n       '2010009 load shedding', '2010010 load shedding',\n       ...\n       '2090036 load shedding', '2090037 load shedding',\n       '2090038 load shedding', '2090039 load shedding',\n       '2090040 load shedding', '2090041 load shedding',\n       '2090042 load shedding', '2090043 load shedding',\n       '2090044 load shedding', '2090045 load shedding'],\n      dtype='object', name='Generator', length=10046)] are in the [columns]"
```

rouille:
@jon-hagg, can you take a look at this PR. Thanks.

FabianHofmann:
Thanks @danielolsen for the comments, I addressed them and resolved the comments, please reopen if you still think there is something missing.

jon-hagg:
One way we could keep imports at the top level is moving the `AnalyzePypsa` class to the `powersimdata.scenario` subpackage. That's also where the `Analyze` class is, but I think it also avoids any dependency on that package from within `powersimdata.input`. Since we have a bunch of imports in the other direction, that should prevent circular imports. 

I tested that out in this [commit](https://github.com/Breakthrough-Energy/PowerSimData/commit/2629c73dd1c151bb44e8a84c34e742b17a86c2b4).

FabianHofmann:
@jon-hagg thanks for looking at it. You are right about considering the module hierarchy. The benefit of the current implementation is that all code, that is related to other packages, is in the `import_data.py` and there is the structure to include other imports as well (thinking about the switch wrapper and possible others). But from a hierarchical point of view, your change makes more sense. Should we pull it in this PR, so I can test every thing again?

PR #612, opened 2022-04-05
fix: make start_date and end_date applicable to boundaries

BainanXia:

### Purpose
Fix a bug when creating scenarios, `start_date` and `end_date` can be set at the boundaries of the time range, i.e. `start_date` can't be `2016-12-31 23:00:00` and `end_date` can't be `2016-01-01 00:00:00`. This shouldn't be the case given our time range is inclusive, which will prevent users from running 1h scenarios for the first or the last hour of the year.

### What the code is doing
A hot fix on the input checks.

### Testing
An integration test of running a 1h USA scenario is successfully conducted.

### Where to look
`powersimdata/scenario/create.py`

### Usage Example/Visuals
Before it gives value error: 
image
Now it runs successfully

### Time estimate
2 min

### DIFF
--- a/powersimdata/scenario/create.py
+++ b/powersimdata/scenario/create.py
@@ -304,9 +304,9 @@ class _Builder:
         hours = (end_ts - start_ts) / np.timedelta64(1, "h") + 1
         if start_ts > end_ts:
             raise ValueError("start_date > end_date")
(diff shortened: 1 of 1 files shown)


rouille (approved):
good catch

PR #613, opened 2022-04-09
Split input_data into separate modules

jon-hagg:
### Purpose
We need to add support for loading electrification profiles and determining the versions, but they will likely be organized differently than the existing profiles. So rather than add more conditional logic to InputData it seems like a good time to do this. 

### What the code is doing
Move logic to get profiles into the `ProfileInput` class. Most of this PR is renaming accordingly. The shared logic is in `InputBase` which also defines the "abstract" methods that are filled in by subclasses. 

If we like this design (it's not perfect, but hopefully better than before) I will follow the pattern for electrification profiles, which would probably just be another subclass. I'm also on the fence about whether to separate the demand flexibility profiles but that could be another follow up PR. 

### Testing
Other than tox, I ran a simulation in the container environment which completed successfully. 

### Time estimate
20 min


powersimdata/data_access/fs_helper.py
```diff
@@ -0,0 +1,57 @@
+import fs
+from fs.multifs import MultiFS
+
+from powersimdata.data_access.ssh_fs import WrapSSHFS
+from powersimdata.utility import server_setup
+
+
+def get_blob_fs(container):
```
rouille, line 8: Missing docstring

powersimdata/data_access/fs_helper.py
```diff
@@ -0,0 +1,57 @@
...
+    account = "besciences"
+    return fs.open_fs(f"azblob://{account}@{container}")
+
+
+def get_ssh_fs(root=""):
```
rouille, line 13: Missing docstring

powersimdata/data_access/fs_helper.py
```diff
@@ -0,0 +1,57 @@
...
+    port = server_setup.SERVER_SSH_PORT
+    username = server_setup.get_server_user()
+    base_fs = fs.open_fs(f"ssh://{username}@{host}:{port}")
+    return WrapSSHFS(base_fs, root)
+
+
+def get_multi_fs(root):
+    """Create filesystem combining the server (if connected) with profile and scenario
+    containers in blob storage. The priority is in descending order, so the server will
+    be used first if possible
```
rouille, line 24: missing description for `root` parameter

powersimdata/data_access/fs_helper.py
jon-hagg: Added these and squashed into the same commit

powersimdata/input/input_data.py
```diff
@@ -3,139 +3,46 @@
...
 
+    def __init__(self):
+        super().__init__()
+        self._file_extension = {"ct": "pkl", "grid": "mat"}
 
-class InputHelper:
-    @staticmethod
-    def get_file_components(scenario_info, field_name):
-        """Get the file name and relative path for either ct or grid.
+    def _get_file_path(self, scenario_info, field_name):
```
rouille, line 17: I think we can still use the staticmethod decorator here

powersimdata/input/input_data.py
```diff
@@ -3,139 +3,46 @@
...
-            *'grid'*.
-        :return: (*pandas.DataFrame*, *dict*, or *str*) -- demand, hydro, solar, wind,
-            demand_flexibility_up, demand_flexibility_dn, demand_flexibility_cost_up,
-            or demand_flexibility_cost_dn as a data frame; change table as a dictionary;
-            or the path to a matfile enclosing the grid data.
-        :raises FileNotFoundError: if file not found on local machine.
+        :param io.IOBase f: an open file object
+        :param str path: the path corresponding to f
+        :raises ValueError: if extension is unknown.
+        :return: object -- the result
```
rouille, line 34: :return: (*object*). Also, we could be more explicit on what the return object is. Could be either a `dict` (change table) or a `Grid` object

powersimdata/input/input_data.py
jon-hagg: Think that would require moving `self._file_extension` outside the class. Feels simpler to have it be an instance method since I'm not 100% sure about static methods and inheritance (though it does seem to work similarly).

rouille: Oops, I did not see  that the `_file_extension` attribute was used

rouille (approved):
Much cleaner. Thanks!

powersimdata/input/input_data.py
BainanXia: AFAIK, static methods will work with inheritance, as well as class methods, while the latter will be more convenient when dealing with multilayer inheritance.

PR #614, opened 2022-04-11
refactor: move add_demand_flexibility to own module

jon-hagg:
### Purpose
Continue factoring out change table methods. 

### What the code is doing
Move add_demand_flexibility to a separate module. Also combined imports into the `__init__.py`, which is not really a thing we do elsewhere, but seemed ok (can't think of any issues, but if there are we can skip that). 

### Testing
tox

### Time estimate
5 min

### DIFF
--- a/powersimdata/input/change_table.py
+++ b/powersimdata/input/change_table.py
@@ -5,12 +5,17 @@ from powersimdata.design.transmission.upgrade import (
     scale_congested_mesh_branches,
     scale_renewable_stubs,
 )
-from powersimdata.input.changes.bus import add_bus, remove_bus
-from powersimdata.input.changes.electrification import add_electrification
-from powersimdata.input.changes.helpers import ordinal
-from powersimdata.input.changes.plant import add_plant, remove_plant, scale_plant_pmin
-from powersimdata.input.changes.storage import add_storage_capacity
-from powersimdata.input.input_data import InputData
+from powersimdata.input.changes import (
+    add_bus,
+    add_demand_flexibility,
+    add_electrification,
+    add_plant,
+    add_storage_capacity,
+    ordinal,
+    remove_bus,
+    remove_plant,
+    scale_plant_pmin,
+)
 from powersimdata.input.transform_grid import TransformGrid
 
 _resources = (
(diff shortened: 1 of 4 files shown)


powersimdata/input/changes/demand_flex.py
```diff
@@ -0,0 +1,59 @@
...
+def add_demand_flexibility(obj, info):
+    """Adds demand flexibility to the system.
+
+    :param powersimdata.input.change_table.ChangeTable obj: change table
+    :param dict info: Each key refers to a different component required to
+        parameterize the demand flexibility model. Each value associated with the
+        keys corresponds to the profile version of the profile in question.
+        Required keys: "demand_flexibility_up", "demand_flexibility_dn".
+        Optional keys: "demand_flexibility_duration", "demand_flexibility_cost_up",
+            "demand_flexibility_cost_dn".
```
rouille, line 15: extra tab here

powersimdata/input/changes/demand_flex.py
```diff
@@ -0,0 +1,59 @@
...
+    :raises TypeError:
+    :raises ValueError:
```
rouille, line 17: We will need to add description for the exceptions

rouille:
What are thr pros and cons to have the import statements in the `__init__.py` file? I need a refresher.

danielolsen:
> What are the pros and cons to have the import statements in the `__init__.py` file? I need a refresher.

- Pros: we can import some things from the module, e.g. `powersimdata.input.changes` instead of from `powersimdata.input.changes.x`.
- Cons: if we change things in the submodules, we need to make sure to change them in the init files as well.

There are probably others but those are the two biggest ones that come to mind.

powersimdata/input/change_table.py
```diff
@@ -471,56 +476,9 @@ def add_storage_capacity(self, info):
...
-                # Add the profile to the change table
-                if len(possible) == 0:
-                    del self.ct["demand_flexibility"]
-                    raise ValueError(f"No {k} profile available.")
-                elif info[k] in possible:
-                    self.ct["demand_flexibility"][k] = info[k]
-                else:
-                    del self.ct["demand_flexibility"]
-                    raise ValueError(f"Available {k} profiles: {', '.join(possible)}")
+        add_demand_flexibility(self, info)
```
lanesmith, line 481: I'm a little confused by this structure. Is the purpose to allow `add_demand_flexibility` to still be called as a method for `ChangeTable` objects with the code living in a different location so that `change_table.py` doesn't get to long?

jon-hagg: Exactly, that's pretty much it. It also helps break up imports and other module level variables, but that's kind of a nice side effect.

BainanXia (approved):
LGTM

powersimdata/input/change_table.py
lanesmith: Oh okay, that makes sense. Thanks for clarifying!

lanesmith (approved):
Looks good to me too. Thanks, Jon!

rouille:
> > What are the pros and cons to have the import statements in the `__init__.py` file? I need a refresher.
> 
>     * Pros: we can import some things from the module, e.g. `powersimdata.input.changes` instead of from `powersimdata.input.changes.x`.
> 
>     * Cons: if we change things in the submodules, we need to make sure to change them in the init files as well.
> 
> 
> There are probably others but those are the two biggest ones that come to mind.

Does anybody have a strong preference regarding the `import` statements?

rouille (approved):
LGTM

PR #616, opened 2022-04-16
Add subclass to get electrification profiles and versions

jon-hagg:
### Purpose
Enable loading electrification profiles from blob storage or local disk, and listing the available versions. This is kind of a draft since we haven't actually uploaded the profiles or determined if this is the best way to organize them, but it is easy to change that later if we want.

### What the code is doing
New `InputBase` subclass to reuse some of the functionality (e.g. caching) but rather than overloading the `scenario_info` parameter in `get_data` I just created a separate method with the relevant parameters. Also, to reuse the chunk of code in `DataAccess` that combines blob and local profiles, the logic is factored out into a function specific to each module and passed as a callback that takes just the filesystem object as an argument. 

### Testing
New unit test.

### Time estimate
15 min

### DIFF
--- a/powersimdata/input/electrified_demand_input.py
+++ b/powersimdata/input/electrified_demand_input.py
@@ -24,7 +24,7 @@ class ElectrifiedDemand(ProfileInput):
         :param str profile: the filename
         :return: (*pandas.DataFrame*) -- profile data frame
         """
-        path = f"raw/{grid_model}/electrification/{kind}/{profile}.csv"
+        path = f"raw/{grid_model}/{kind}/{profile}.csv"
         return self._get_data_internal(path)
 
     def get_profile_version(self, grid_model, kind, end_use, tech):
--- a/powersimdata/input/electrified_demand_input.py
+++ b/powersimdata/input/electrified_demand_input.py
@@ -1,0 +1,43 @@
+from powersimdata.input.profile_input import ProfileInput
+
+
+def get_profile_version(_fs, grid_model, kind, end_use, tech):
+    _fs = _fs.makedirs(f"raw/{grid_model}/{kind}", recreate=True)
+    base_name = f"{end_use}_{tech}_"
+    matching = [f for f in _fs.listdir(".") if base_name in f]
+
+    return [f.replace(base_name, "").replace(".csv", "") for f in matching]
+
+
+class ElectrifiedDemand(ProfileInput):
+    """Loads electrification profile data"""
+
+    def __init__(self):
+        super().__init__()
+        self._file_extension = {}
+
+    def get_profile(self, grid_model, kind, profile):
+        """Get the specified profile
+
+        :param str grid_model: the grid model
+        :param str kind: the kind of electrification
+        :param str profile: the filename
+        :return: (*pandas.DataFrame*) -- profile data frame
+        """
+        path = f"raw/{grid_model}/electrification/{kind}/{profile}.csv"
+        return self._get_data_internal(path)
+
+    def get_profile_version(self, grid_model, kind, end_use, tech):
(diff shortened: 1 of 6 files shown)


powersimdata/input/electrified_demand_input.py
```diff
@@ -0,0 +1,43 @@
+from powersimdata.input.profile_input import ProfileInput
+
+
+def get_profile_version(_fs, grid_model, kind, end_use, tech):
+    _fs = _fs.makedirs(f"raw/{grid_model}/{kind}", recreate=True)
+    base_name = f"{end_use}_{tech}_"
+    matching = [f for f in _fs.listdir(".") if base_name in f]
+
+    return [f.replace(base_name, "").replace(".csv", "") for f in matching]
```
jon-hagg, line 9: Note: I changed `lstrip` and `rstrip` to `replace` here because it turns out those strip any of the given characters starting from one side until one doesn't match, so it could lead to unpredictable results.

danielolsen: What if we have `'demand'` or `'solar'` or whatever elsewhere within the string, it seems that will get replaced as well. Should we do this via regex instead? E.g.
```python
>>> import re
>>> s = "ab_cd_ef"
>>> match = re.search("^(?P<prefix>ab_)?(?P<main>[^_]+)(?P<suffix>_ef)?$", s)
>>> match.group('prefix')
'ab_'
>>> match.group('main')
'cd'
>>> match.group('suffix')
'_ef'
```

danielolsen: String parsing methods are definitely clearer, but have the potential for weird side effects if the strings are not constructed in the way that we want.

jon-hagg: With this implementation, `kind` is either "building" or "transportation". So the profiles in each of those folders will have the structure `[end_use]_[tech]_[version].csv`.

BainanXia: Agree. I will defer to @jon-hagg's decision at this point.

jon-hagg: I think since we are responsible for profile names and version, and in this case replacing the entire prefix up to the version, it's unlikely we run into issues. E.g. the version should hopefully not contain ".csv" or "res_cooking_standard_heat_pump". I do like regex, but since the naming is so consistent we probably don't need it at this point?

jon-hagg: It does operate on the local folder. I guess we should document the expected structure for custom profiles if we don't have that yet. My assumption is that it's not _supposed_ to work for arbitrary names, just custom versions.

danielolsen: I trust your judgement. It's not like we have a ton of users out there adding a bunch of weirdly named files yet.

powersimdata/input/electrified_demand_input.py
```diff
@@ -0,0 +1,43 @@
...
+        :param str kind: *'building'*, *'transportation'*
+        :param str end_use: electrification use case
+        :param str tech: the technology used for the given use case
+        :return: (*list*) -- available profile version.
+        """
+
+        def _callback(fs):
+            return get_profile_version(fs, grid_model, kind, end_use, tech)
+
+        return self.data_access.get_profile_version(_callback)
```
danielolsen, line 43: Is there a benefit to defining `_callback` here rather than passing a lambda function to `get_profile_version`?

jon-hagg: No benefit, or any real difference I can think of.

danielolsen (approved):
LGTM

Issue #567, opened 2021-11-30
Allow Grids to be loaded from CSVs with an arbitrary number of non-essential data columns


danielolsen:
# :rocket:

- [ ] Is your feature request essential for your project?

### Describe the workflow you want to enable
Allow `Grid` objects to be loaded from CSVs with an arbitrary number of non-essential data columns, for example data on asset ownership, names, transmission line routing, etc.

### Describe your proposed implementation
This will require refactoring how the `case.mat` files are generated, since currently the data tables are generated by _dropping_ specified columns, rather than only _writing_ certain columns.

The more complicated design issue is how to work with extra Grid information in the context of the current `Scenario` object framework. Currently, the information that's dropped from the tables that get interpreted by the simulation engine (those required for the power flow) get added to the case.mat file as additional tables which get re-loaded later by a Scenario in Analyze state which is reading the grid.mat file that gets produced by REISE.jl as a record of the exact power system parameters that were used in optimization (since REISE.jl makes some modifications to things like ramp rates, Pmin values, and linearization of cost curves). If we don't pass the extra information to the case.mat file to be read later, then we can only infer information by looking at the corresponding information in the base grid, which may have have changed since the Scenario was run, and won't cover any additions to the Grid that were made as part of the scenario setup. If we do pass extra information into the case.mat file for some grid models, then we end up with grid.mat objects with different sets of extra information that need to be interpreted differently.

### Additional context
The corresponding PreREISE issue to write extra information to the HIFLD grid model CSVs is https://github.com/Breakthrough-Energy/PreREISE/issues/242.

danielolsen:
This could also be relevant for Grids/Scenarios that are converted from PyPSA: we may want to be able to accept additional columns during translation, without breaking the ability to run PCMs.

rouille:
Addressed in #701

PR #623, opened 2022-04-26
feat: EIA ID of buses in the synthetic grid

dongqi-wu:

### Purpose
Add the EIA ID LSE of each bus in the system as an additional column, to be used later to match with LSE-based demand flexibility and use for simulation in REISE.jl [(link).](https://github.com/Breakthrough-Energy/REISE.jl/pull/178) The EIA IDs are identified using an offline process that will be a separate PR in PreREISE.

### What the code is doing
added a new column in network/usa_tamu/bus.csv

### Where to look
only change is in network/usa_tamu/bus.csv

### Time estimate
5 min

### DIFF
--- a/powersimdata/input/const/grid_const.py
+++ b/powersimdata/input/const/grid_const.py
@@ -96,6 +96,7 @@ col_name_bus = [
     "mu_Vmax",
     "mu_Vmin",
     "interconnect",
+    "eia_id",
     "lat",
     "lon",
 ]
--- a/powersimdata/input/const/pypsa_const.py
+++ b/powersimdata/input/const/pypsa_const.py
@@ -22,6 +22,7 @@ pypsa_const = {
             "mu_Vmax",
             "mu_Vmin",
             "GenFuelCost",
+            "eia_id",
         ],
     },
     "generator": {


rouille:
I think we want to fixup the second commit.

dongqi-wu:
> I think we want to fixup the second commit.

I just did that

danielolsen:
I think the order that makes the most sense for merging all these PRs is:
1. https://github.com/Breakthrough-Energy/PreREISE/pull/282 to make sure we're generating this data as we expect
2. https://github.com/Breakthrough-Energy/REISE.jl/pull/178 to be able to use this data, if it exists (testing both with Scenarios prepared with `develop` and with this branch)
3. This PR (https://github.com/Breakthrough-Energy/PowerSimData/pull/623). This probably also relates to https://github.com/Breakthrough-Energy/PowerSimData/issues/567, since in addition to adding the data to the CSVs, we'll want to make sure that they get prepared properly for downstream use with REISE.jl.

lanesmith (approved):
Looks good to me, thanks!

BainanXia:
Not sure whether adding additional column in those const files will potentially break anything outside the tests. Could you confirm @rouille before I merge it?

rouille:
> Not sure whether adding additional column in those const files will potentially break anything outside the tests. Could you confirm @rouille before I merge it?

I will look into it. Keep you updated.

BainanXia:
> > Not sure whether adding additional column in those const files will potentially break anything outside the tests. Could you confirm @rouille before I merge it?
> 
> I will look into it. Keep you updated.

🙏

powersimdata/input/const/grid_const.py
```diff
@@ -96,6 +96,7 @@
     "mu_Vmax",
     "mu_Vmin",
     "interconnect",
+    "eia_id",
```
rouille, line 99: You also need to add `"int"`to the `col_type_bus` list right before the second to last element. Not sure, we actually use the type of the field now that that we don't translate the content **grid.mat** file to  a `Grid` object but still it would be cleaner

rouille:
If you look at **pyreisejl/utility/converters.py** module in ***REISE.jl***, it seems that only `"bus_id"`, `"Pd"` and `"zone_id"` will be exported in the **bus.csv** file. Is that correct @jenhagg?

I am not sure what would be the good workflow here. Should we include it in the file if it is in the `bus` data frame of the **grid.pkl**. This would mean that we would need to drop off the column when we create the column if the user does not want to enable demand flexibility. Is that what you have in mind @BainanXia and @lanesmith , i.e., having a user interface function `enable_demand_flexibility` in the **create** state that would keep the `"eia_id"` column if called?

lanesmith:
> If you look at **pyreisejl/utility/converters.py** module in _**REISE.jl**_, it seems that only `"bus_id"`, `"Pd"` and `"zone_id"` will be exported in the **bus.csv** file. Is that correct @jenhagg?
> 
> I am not sure what would be the good workflow here. Should we include it in the file if it is in the `bus` data frame of the **grid.pkl**. This would mean that we would need to drop off the column when we create the column if the user does not want to enable demand flexibility. Is that what you have in mind @BainanXia and @lanesmith , i.e., having a user interface function `enable_demand_flexibility` in the **create** state that would keep the `"eia_id"` column if called?

Good catch on `pyreisejl/utility/converters.py`. Here's an idea of a workflow we could follow: 

It appears that we still have the `enable_doe_flexibility` parameter in `REISE.jl`. If demand flexibility is enabled but the flexibility profiles aren't provided in `PowerSimData`, we could set the `enable_doe_flexibility` parameter to True. Then when files are being read in through `pyreisejl/utility/call.py`, we could check for the existence of a `demand_flexibility_parameters.csv` file. If it's there (indicating that demand flexibility is enabled), we could check that file for the `enable_doe_flexibility` parameter. If it's True, we could include the `eia_id` column in the `bus.csv` file via `pyreisejl/utility/converters.py`; if it's False, we wouldn't include the `eia_id` column.

rouille:
> > If you look at **pyreisejl/utility/converters.py** module in _**REISE.jl**_, it seems that only `"bus_id"`, `"Pd"` and `"zone_id"` will be exported in the **bus.csv** file. Is that correct @jenhagg?
> 
> > 
> 
> > I am not sure what would be the good workflow here. Should we include it in the file if it is in the `bus` data frame of the **grid.pkl**. This would mean that we would need to drop off the column when we create the column if the user does not want to enable demand flexibility. Is that what you have in mind @BainanXia and @lanesmith , i.e., having a user interface function `enable_demand_flexibility` in the **create** state that would keep the `"eia_id"` column if called?
> 
> 
> 
> Good catch on `pyreisejl/utility/converters.py`. Here's an idea of a workflow we could follow: 
> 
> 
> 
> It appears that we still have the `enable_doe_flexibility` parameter in `REISE.jl`. If demand flexibility is enabled but the flexibility profiles aren't provided in `PowerSimData`, we could set the `enable_doe_flexibility` parameter to True. Then when files are being read in through `pyreisejl/utility/call.py`, we could check for the existence of a `demand_flexibility_parameters.csv` file. If it's there (indicating that demand flexibility is enabled), we could check that file for the `enable_doe_flexibility` parameter. If it's True, we could include the `eia_id` column in the `bus.csv` file via `pyreisejl/utility/converters.py`; if it's False, we wouldn't include the `eia_id` column.

Not sure I understand what you propose but since I never worked on this project I don't know what would be the desired workflow and how to enable it. I let you take care of the implementation and I will be happy to review the PRs

PR #626, opened 2022-04-28
Aggregate demand profiles during scenario preparation

jon-hagg:

### Purpose
Apply the change table scaling for electrification profiles, and combine that with scaled demand profile when preparing a scenario.

### What the code is doing
Add new class `TransformDemand`, where we use the `.value()` method to get the demand profile for a given grid, and given class of electrification. In `TransformProfile`, we add that to the base demand, for each kind of electrification specified in the change table.

Most of the logic in `TransformDemand` is to kind of invert the change table mapping, which makes the transformation logic simpler, as we just fetch one profile at a time, scale it, and sum the results.  

### Testing
Haven't done integration tests yet, but we have a new unit test that combines grid and zone scaling, which checks that the zone scaling is applied correctly, and the rest of the grid is scaled as defined in the `ct["building"]["grid"]` entry (in the case of building electrification). 

### Time estimate
20 mins

### DIFF
--- a/powersimdata/input/tests/test_transform_demand.py
+++ b/powersimdata/input/tests/test_transform_demand.py
@@ -6,10 +6,11 @@ from powersimdata.input.transform_demand import TransformDemand
 from powersimdata.tests.mock_profile_input import MockProfileInput
 
 
-def test_profile_to_zone():
+def test_profile_mappings():
     grid = Grid("Texas")
     ct = ChangeTable(grid)
-    info = {
+    grid_info = {"res_heating": {"standard_heat_pump_v2": 0.3}}
+    zone_info = {
         "East": {"res_cooking": {"advanced_heat_pump_v2": 0.7}},
         "Coast": {
             "com_hot_water": {
@@ -23,10 +24,14 @@ def test_profile_to_zone():
     }
 
     kind = "building"
-    ct.add_electrification(kind, {"zone": info})
+    ct.add_electrification(kind, {"zone": zone_info, "grid": grid_info})
     td = TransformDemand(grid, ct, kind)
-    actual = td._get_profile_to_zone()
 
+    actual = td.p2g
+    expected = {"res_heating_standard_heat_pump_v2.csv": 0.3}
+    assert expected == actual
+
+    actual = td.p2z
     expected = {
         "res_cooking_advanced_heat_pump_v2.csv": [(308, 0.7), (301, 0.3)],
         "com_hot_water_standard_heat_pump_v1.csv": [(307, 0.6)],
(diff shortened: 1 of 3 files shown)


powersimdata/input/transform_demand.py
```diff
@@ -0,0 +1,101 @@
...
+        self._profile_data = ElectrifiedDemand()
+        self._set_scale_factors()
+
+    def _get_base_profile(self, profile):
+        """Return the base profile from local or blob storage
+
+        :return: (*pandas.DataFrame*) -- profile data frame, filtered to zones within
+            the current grid
+        """
+        zone_id = sorted(self.grid.bus.zone_id.unique())
```
rouille, line 26: The `Grid` has a `id2zone` attribute

powersimdata/input/transform_demand.py
```diff
@@ -0,0 +1,101 @@
...
+        demand = self._profile_data.get_profile(model, self.kind, profile).loc[
+            :, zone_id
+        ]
+        return demand
+
+    def _get_profile_to_zone(self):
+        """Maps profile name to scale factors for each zone
+
+        :return: (*dict*) -- a dictionary mapping str to list of tuples of (zone_id,
+                scale_factor)
```
rouille, line 37: Extra indent

powersimdata/input/transform_demand.py
```diff
@@ -0,0 +1,101 @@
...
+        return df
+
+    def value(self):
+        """Return the combined electrified demand
+
+        :return: (*pandas.DataFrame*) -- data frame with hourly index and zone columns,
+            where the values are demand (in MWh)
+        """
+        profiles = set(self.p2z.keys()) | set(self.p2g.keys())
+        return sum(self.get_profile(p) for p in profiles)
```
rouille, line 101: Nice. I did not know you could sum data frames this way. I believe it works because all the data frames in the list have the same columns:
```
>>> a = pd.DataFrame({"A": [1,2,3], "B":[4,5,6]})
>>> b = pd.DataFrame({"A": [1,2,3]})
>>> sum([a,b])
   A   B
0  2 NaN
1  4 NaN
2  6 NaN
>>> 
```

powersimdata/input/tests/test_transform_demand.py
```diff
@@ -0,0 +1,62 @@
...
+    mock_input = MockProfileInput(grid)
+    demand = mock_input.get_data(None, "demand")
+    mock_input.get_profile = lambda *args: demand
+
+    td = TransformDemand(grid, ct, kind)
+    td._profile_data = mock_input
+    result = td.value()
+
+    pd.testing.assert_series_equal(0.7 * demand.loc[:, 308], result.loc[:, 308])
+    pd.testing.assert_frame_equal(0.3 * demand.loc[:, :307], result.loc[:, :307])
```
rouille, line 62: Thanks for adding these tests

rouille (approved):
Thanks.

PR #627, opened 2022-05-03
Create abstract grid class for model enclosed in CSV files

rouille:

### Purpose
Create an abstract grid class for grid models that enclosed in CSVs files. In the future, some grid models will be stored in other forms so it is important to make the distinction

### What the code is doing
Create a new class dedicated to the reading of CSV grid models (`AbstractGridCSV`) that inherits from a higher level `AbstractGrid` class

### Testing
Run `tox` successfully

### Where to look
The `powersimdata.input.abstract_grid` module encloses the classes.

### Usage Example/Visuals
N/A

### Time estimate
5min

### DIFF
--- a/powersimdata/input/abstract_grid.py
+++ b/powersimdata/input/abstract_grid.py
@@ -31,6 +31,10 @@ class AbstractGrid:
         self.branch = pd.DataFrame()
         self.storage = storage_template()
 
+
+class AbstractGridCSV(AbstractGrid):
+    """Grid Builder."""
+
     def _set_data_loc(self, top_dirname):
         """Sets data location.
 
--- a/powersimdata/network/hifld/model.py
+++ b/powersimdata/network/hifld/model.py
@@ -1,10 +1,10 @@
 import os
 
-from powersimdata.input.abstract_grid import AbstractGrid
+from powersimdata.input.abstract_grid import AbstractGridCSV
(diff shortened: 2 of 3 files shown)


powersimdata/input/abstract_grid.py
```diff
@@ -31,6 +31,14 @@ def __init__(self):
         self.branch = pd.DataFrame()
         self.storage = storage_template()
 
+
+class AbstractGridCSV(AbstractGrid):
+    """Grid Builder."""
+
+    def __init__(self):
+        """Constructor"""
+        super().__init__()
```
danielolsen, line 40: Do we need this if it only calls `super`?

danielolsen (approved):

powersimdata/input/abstract_grid.py
rouille: No we don't

jon-hagg (approved):

powersimdata/input/abstract_grid.py
BainanXia: Indeed. If `__init__` behavior are the same with parent, it can be inherited without explicit calls: https://stackoverflow.com/questions/54911999/does-a-subclass-need-to-initialize-an-empty-super-class

BainanXia (approved):

PR #628, opened 2022-05-05
Refactor Grid object

rouille:

### Purpose
Set the `grid_model` and `model_immutables` attributes of the `Grid` class within their respective model. The model immutables  will no longer be a static object for Europe. The loadzone are the buses and it is necessary to instantiate a PyPSA `Network` object to access the bus table. Moreover the number of loadzone will depend on the spatial reduction. 

### What the code is doing
The `model_immtables` attributes is now set in the  `TAMU`, `HIFLD`, `FromREISE` and `FromREISEjl`. This will set the ground for Europe. To avoid to build the network if we just want to access the model immutables, a `build` method is added to the `TAMU` and `HIFLD` class. One change too is that the `model_immutables` of a `Grid` object now only encloses the mapping for the specified interconnect(s) and not the full USA as previously.

### Testing
Run `tox` successfully

### Where to look
* I would start with `AbstractGrid` -> `AbstractGridCSV` and `ScenarioGrid` -> `TAMU`, `HIFLD`, `FromREISE` and `FromREISEjl` --> `Grid`
* I have removed the dependency on the model immutables in the `powersimdata.scenario.create` module
* I have fixed a bug in the `powersimdata.network.constants.region.usa` module that impacted state/nterconnect(s) with Texas and cast some single element `Int64Index` from `pandas` to `int`.

### Usage Example/Visuals
N/A

### Time estimate
20min

### DIFF
--- a/powersimdata/network/constants/region/usa.py
+++ b/powersimdata/network/constants/region/usa.py
@@ -173,6 +173,7 @@ def get_state_mapping(zone):
 
     mapping["state"] = set(zone["state"])
     mapping["abv"] = set(zone["abv"])
+    mapping["state_abbr"] = set(zone["abv"])
     mapping["state2loadzone"] = {
         k: set(v)
         for k, v in zone.groupby("state")["zone_name"].unique().to_dict().items()
--- a/powersimdata/network/constants/region/usa.py
+++ b/powersimdata/network/constants/region/usa.py
@@ -204,7 +204,9 @@ def get_loadzone_mapping(zone):
     mapping["timezone2id"] = {
         t: set(i) for t, i in zone.groupby("time_zone").groups.items()
     }
-    mapping["loadzone2id"] = {l: i for l, i in zone.groupby("zone_name").groups.items()}
+    mapping["loadzone2id"] = {
+        l: i[0] for l, i in zone.groupby("zone_name").groups.items()
+    }
     mapping["loadzone2state"] = dict(zip(zone["zone_name"], zone["state"]))
     mapping["loadzone2abv"] = dict(zip(zone["zone_name"], zone["abv"]))
     mapping["loadzone2interconnect"] = dict(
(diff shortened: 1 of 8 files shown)


powersimdata/network/constants/region/usa.py
```diff
@@ -175,6 +173,7 @@ def get_state_mapping(zone):
 
     mapping["state"] = set(zone["state"])
     mapping["abv"] = set(zone["abv"])
+    mapping["state_abbr"] = set(zone["abv"])
```
jon-hagg, line 176: Do we need the same entry twice?

rouille: It kind of help in my refactoring of the `area_to_loadzone` function and functions that check that areas exist in a grid model that I do in `ben/eu` since one of the area type a user can pass to the function is `state_abbr` and `country_abbr`. See https://github.com/Breakthrough-Energy/PowerSimData/blob/d232cc0f5cb0965c00045127e3bddc7e4c4417b4/powersimdata/network/model.py#L43 for instance

powersimdata/network/constants/region/usa.py
```diff
@@ -206,7 +205,9 @@ def get_loadzone_mapping(zone):
     mapping["timezone2id"] = {
         t: set(i) for t, i in zone.groupby("time_zone").groups.items()
     }
-    mapping["loadzone2id"] = {l: i for l, i in zone.groupby("zone_name").groups.items()}
+    mapping["loadzone2id"] = {
+        l: i[0] for l, i in zone.groupby("zone_name").groups.items()
+    }
```
danielolsen, line 210: This is the fix you mentioned where we previously had a single-element index but now we have just an int?

danielolsen (approved):
Makes sense.

powersimdata/network/constants/region/usa.py
rouille: Yes. Since there is a one-to-one relation between `zone_name` and `zone_id`, we don't want a `list`. This is what we are doing in `develop`.

BainanXia (approved):
Thanks. Avoiding parsing csvs if only the mappings from `model_immutables` are needed is nice.

danielolsen:
I can confirm that this works end-to-end from preparing a Scenario, running it, and then loading it again and getting the Grid. Tested with Texas, USA, and Texas_Western.

rouille:
> I can confirm that this works end-to-end from preparing a Scenario, running it, and then loading it again and getting the Grid. Tested with Texas, USA, and Texas_Western.

Thanks!

PR #629, opened 2022-05-07
Combine scenario list from different sources

jon-hagg:
### Purpose
Fix the issue where a user won't get a scenario list with existing local scenarios if not connected to the server. 

### What the code is doing
Best effort to download and use any possible copies of scenario or execute list csv. The local copies are updated with the latest version if possible (always for blob storage, and server copy if connected). Then the results are combined, since we might as well handle a case where the local copy hasn't been updated since the last time we add scenarios to blob storage. 

Unrelated: noticed that the `.copy()` data access method is only used in one place, so removed it from the interface. 

### Testing
Manual testing: load `Scenario(824)` which is not in blob storage, but is in my local ScenarioList.csv. If not connected to the server, it will load as much as possible but fails since I don't have the grid.mat, as expected.

### Time estimate
10 min

### DIFF
--- a/powersimdata/data_access/csv_store.py
+++ b/powersimdata/data_access/csv_store.py
@@ -44,21 +44,27 @@ class CsvStore:
         self.data_access = data_access
 
     def get_table(self):
-        """Read the given file from the server, falling back to local copy if
-        unable to connect.
+        """Attempt to download the file from server and blob storage, falling back to
+        local copy if one exists, and return the combined result.
 
         :return: (*pandas.DataFrame*) -- the specified table as a data frame.
         """
         filename = self._FILE_NAME
-        try:
-            return self._get_table(filename)
-        except:  # noqa
-            return self._get_table(filename + ".2")
+        orig = self._get_table(filename)
+        blob = self._get_table(filename + ".2")
+        df = pd.concat([orig, blob])
+        return df[~df.index.duplicated()]
 
     def _get_table(self, filename):
-        self.data_access.copy_from(filename)
-        with self.data_access.get(filename) as (f, _):
-            return _parse_csv(f)
+        try:
+            self.data_access.copy_from(filename)
+        except:  # noqa
+            pass
+        try:
+            with self.data_access.get(filename) as (f, _):
+                return _parse_csv(f)
+        except:  # noqa
+            return pd.DataFrame()
 
     def commit(self, table, checksum):
         """Save to local directory and upload if needed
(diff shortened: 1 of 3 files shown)


powersimdata/data_access/csv_store.py
```diff
@@ -50,15 +50,21 @@ def get_table(self):
         :return: (*pandas.DataFrame*) -- the specified table as a data frame.
```
BainanXia, line 50: Should we update the doc of this function as well?

powersimdata/data_access/csv_store.py
```diff
@@ -50,15 +50,21 @@ def get_table(self):
...
         """
         filename = self._FILE_NAME
-        try:
-            return self._get_table(filename)
-        except:  # noqa
-            return self._get_table(filename + ".2")
+        orig = self._get_table(filename)
+        blob = self._get_table(filename + ".2")
```
BainanXia, line 54: Here, we either get a server copy or a local copy in `orig` and always get a blob copy in `blob`, right? Given a blob copy will always be a subset of a server copy/local copy, should we check whether we fail in `orig` (i.e. not connect to server and no local copy) first? And get a blob copy only if `orig` fails?

powersimdata/data_access/csv_store.py
```diff
@@ -50,15 +50,21 @@ def get_table(self):
...
-            return _parse_csv(f)
+        try:
+            self.data_access.copy_from(filename)
+        except:  # noqa
+            pass
+        try:
+            with self.data_access.get(filename) as (f, _):
+                return _parse_csv(f)
+        except:  # noqa
+            return pd.DataFrame()
```
BainanXia, line 67: Reading through the logic here, it seems it will also work if one doesn't have a local copy and not connects to the internet, but have a local copy of `csv.2`, i.e. the one on the blob storage? It will be picked up by the second `try` block in the `blob` query in l.54. It should be fine though, just wanna make sure that's the intended behavior.

powersimdata/data_access/csv_store.py
rouille: I guess you can a very old copy locally because you lost access to the sever for some reason and meanwhile the blob storage list has been updated with scenarios that are not in the local copy.

powersimdata/data_access/csv_store.py
jon-hagg: yep, that's more or less the intent. I wasn't specifically considering lack of internet, since e.g. if there are no local copies you'd get an empty data frame, and I'm not what downstream failures would look like. But say you had a local `.csv.2` as well as local scenario files (pf, pg, etc), then I think everything would still work in that case.

BainanXia (approved):
LGTM

PR #631, opened 2022-05-11
Add PyPSA-Eur

rouille:

### Purpose
Load a PyPSA-eur `Network` object into memory, generate model immutables for the grid model and refactor all functions mapping one area to another for a grid model or functions checking that areas or resources are part of a grid model.

### What the code is doing
* Update **Pipfile**, **Pipfile.lock** and **requirements.txt** files with new `zenodo_get` package and pinned version of `fs` (there is an issue with the most recent version as dicussed [here](https://github.com/PyFilesystem/pyfilesystem2/issues/535))
* Update **.gitignore**
* Refactor the `area_to_loadzone` function and functions checking that areas/rsources are part of a grid model
* Create model immutables for Europe
* Create a new class, `TUB`, where a PyPSA `network` object is instantiated from a netCDF file distributed on zenodo and the model immutables are loaded.
* Write new tests and update existing ones

### Testing
* Existing unit tests run successfully, 
* New unit tests were written
* Feature was tested manually:
```
>>> from powersimdata.network.europe_tub.model import TUB
>>> tub = TUB("ContinentalEurope", reduction=128)
Title: PyPSA-Eur: An Open Optimisation Model of the European Transmission System (Dataset)
Keywords: power system model, capacity expansion model, energy system model, electricity system model, python
Publication date: 2021-09-22
DOI: 10.5281/zenodo.5521712
Total size: 1767.1 MB

Link: https://zenodo.org/api/files/6e58c6dc-083b-46c8-8255-8516e3eb0ed8/networks.zip   size: 1767.1 MB
networks.zip is already downloaded correctly.
All files have been downloaded.
>>> tub.build()
WARNING:pypsa.io:
Importing PyPSA from older version of PyPSA than current version.
Please read the release notes at https://pypsa.readthedocs.io/en/latest/release_notes.html
carefully to prepare your network for import.
Currently used PyPSA version [0, 19, 3], imported network file PyPSA version [0, 18, 0].

INFO:pypsa.io:Imported network elec_s_128.nc has buses, carriers, generators, lines, links, loads, storage_units
>>> tub.model_immutables.zones["interconnect2loadzone"]
{'ContinentalEurope': {'DE0 14', 'DE0 15', 'ES0 8', 'FR0 9', 'IT0 7', 'BE0 2', 'PL0 2', 'IT0 5', 'FR0 5', 'CZ0 0', 'BE0 0', 'IT0 6', 'FR0 12', 'DK0 0', 'FR0 0', 'IT0 0', 'FR0 14', 'FR0 16', 'DE0 7', 'RS0 0', 'IT0 2', 'RO0 1', 'IT0 1', 'DE0 17', 'IT0 3', 'FR0 1', 'ES0 4', 'FR0 6', 'PL0 4', 'ES0 7', 'FR0 8', 'ES2 0', 'PT0 0', 'FR0 7', 'BE0 1', 'MK0 0', 'ES0 0', 'FR0 13', 'DE0 8', 'DE0 6', 'BG0 0', 'ES0 5', 'FR0 4', 'DE0 11', 'DE0 12', 'IT0 10', 'DE0 18', 'DE0 0', 'ES0 2', 'IT0 11', 'DE0 13', 'DE0 16', 'PT0 1', 'PL0 1', 'FR0 10', 'FR0 18', 'SI0 0', 'IT0 9', 'DE0 2', 'DE0 3', 'IT0 8', 'PL0 0', 'DE0 19', 'AL0 0', 'BA0 0', 'CZ0 1', 'SK0 0', 'HU0 0', 'DE0 1', 'ES0 6', 'FR0 11', 'IT1 0', 'DK3 0', 'RO0 0', 'ME0 0', 'DE0 5', 'DE0 10', 'ES0 3', 'AT0 0', 'FR0 17', 'DE0 9', 'ES0 1', 'DE0 4', 'AT0 1', 'HR0 0', 'PL0 3', 'CH0 0', 'FR0 2', 'GR0 0', 'FR0 3', 'FR0 15', 'LU0 0', 'IT0 4'}}
>>> 
```
or
```
>>> from powersimdata.network.europe_tub.model import TUB
>>> tub = TUB("Baltic", reduction=128)
Title: PyPSA-Eur: An Open Optimisation Model of the European Transmission System (Dataset)
Keywords: power system model, capacity expansion model, energy system model, electricity system model, python
Publication date: 2021-09-22
DOI: 10.5281/zenodo.5521712
Total size: 1767.1 MB

Link: https://zenodo.org/api/files/6e58c6dc-083b-46c8-8255-8516e3eb0ed8/networks.zip   size: 1767.1 MB
networks.zip is already downloaded correctly.
All files have been downloaded.
>>> tub.build()
WARNING:pypsa.io:
Importing PyPSA from older version of PyPSA than current version.
Please read the release notes at https://pypsa.readthedocs.io/en/latest/release_notes.html
carefully to prepare your network for import.
Currently used PyPSA version [0, 19, 3], imported network file PyPSA version [0, 18, 0].

INFO:pypsa.io:Imported network elec_s_128.nc has buses, carriers, generators, lines, links, loads, storage_units
>>> tub.network.buses
tub.network.buses    tub.network.buses_t  
>>> tub.network.buses
       v_nom          x          y country control   generator type carrier  unit  v_mag_pu_set  v_mag_pu_min  v_mag_pu_max sub_network
Bus                                                                                                                                    
EE6 0  380.0  25.879152  58.979906      EE   Slack  EE6 0 CCGT           AC  None           1.0           0.0           inf            
LT6 0  380.0  23.114445  55.196651      LT      PQ                       AC  None           1.0           0.0           inf            
LV6 0  380.0  23.724680  56.838802      LV      PQ                       AC  None           1.0           0.0           inf            
>>> tub.network.generators
                     p_nom_max        weight   p_nom   capital_cost  ... ramp_limit_up ramp_limit_down  ramp_limit_start_up  ramp_limit_shut_down
Generator                                                            ...                                                                         
EE6 0 CCGT                 inf      0.000000   173.0       0.000000  ...           NaN             NaN                  1.0                   1.0
EE6 0 OCGT                 inf      0.000000   250.0       0.000000  ...           NaN             NaN                  1.0                   1.0
EE6 0 offwind-ac  3.353159e+04  13485.355948     0.0  217338.995520  ...           NaN             NaN                  1.0                   1.0
EE6 0 offwind-dc  7.074024e+03   3202.679493     0.0  230061.246431  ...           NaN             NaN                  1.0                   1.0
EE6 0 oil                  inf      0.000000  2111.0       0.000000  ...           NaN             NaN                  1.0                   1.0
EE6 0 onwind      9.366104e+04  26562.271969     0.0  109295.569252  ...           NaN             NaN                  1.0                   1.0
EE6 0 solar       2.067910e+04   1929.797484     0.0   55064.071419  ...           NaN             NaN                  1.0                   1.0
LT6 0 OCGT                 inf      0.000000  1575.0       0.000000  ...           NaN             NaN                  1.0                   1.0
LT6 0 offwind-ac  3.789093e+03   1487.387960     0.0  199927.923232  ...           NaN             NaN                  1.0                   1.0
LT6 0 offwind-dc  2.331307e+03   1069.423004     0.0  215223.519446  ...           NaN             NaN                  1.0                   1.0
LT6 0 onwind      1.277605e+05  29677.320189     0.0  109295.569252  ...           NaN             NaN                  1.0                   1.0
LT6 0 solar       5.896199e+04   5451.525285     0.0   55064.071419  ...           NaN             NaN                  1.0                   1.0
LV6 0 CCGT                 inf      0.000000  1025.0       0.000000  ...           NaN             NaN                  1.0                   1.0
LV6 0 offwind-ac  2.432949e+04   9715.381125     0.0  203629.370626  ...           NaN             NaN                  1.0                   1.0
LV6 0 offwind-dc  9.714325e+03   4442.971859     0.0  216893.775841  ...           NaN             NaN                  1.0                   1.0
LV6 0 onwind      1.486485e+05  36529.730498     0.0  109295.569252  ...           NaN             NaN                  1.0                   1.0
LV6 0 ror                  inf    642.100000   642.1  270940.715283  ...           NaN             NaN                  1.0                   1.0
LV6 0 solar       3.999053e+04   3750.348474     0.0   55064.071419  ...           NaN             NaN                  1.0                   1.0

[18 rows x 31 columns]
>>> 
```

 ### Where to look
Most important modules are:
* `powersimdata.network.model` where `area_to_loadzone` is refactored
* `powersimdata.network.constants.region.europe` where the geographical immutables for Europe are defined 
* `powersimdata.network.europe_tub.model` where the `TUB` class is defined

### Usage Example/Visuals
N/A

### Time estimate
30min

### DIFF
--- a/powersimdata/input/check.py
+++ b/powersimdata/input/check.py
@@ -299,19 +299,20 @@ def _check_grid_type(grid):
         raise TypeError(f"grid must be a {_grid.Grid} object")
 
 
-def _check_areas_and_format(areas, grid_model="usa_tamu"):
+def _check_areas_and_format(areas, mi=ModelImmutables("usa_tamu")):
     """Ensure that areas are valid. Duplicates are removed and state abbreviations are
     converted to their actual name.
 
     :param str/list/tuple/set areas: areas(s) to check. Could be load zone name(s),
         state name(s)/abbreviation(s) or interconnect(s).
-    :param str grid_model: grid model.
-    :raises TypeError: if areas is not a list/tuple/set of str.
-    :raises ValueError: if areas is empty or not valid.
+    :param powersimdata.network.model.ModelImmutables mi: immutables of a grid model.
+    :raises TypeError: if ``areas`` is not a list/tuple/set of str.
+    :raises ValueError: if ``areas`` is empty or not valid.
     :return: (*set*) -- areas as a set. State abbreviations are converted to state
         names.
     """
-    mi = ModelImmutables(grid_model)
+    division = [a for a in mi.zones["mappings"] if "abbr" in a][0].split("_")[0]
+
     if isinstance(areas, str):
         areas = {areas}
     elif isinstance(areas, (list, set, tuple)):
@@ -322,12 +323,7 @@ def _check_areas_and_format(areas, grid_model="usa_tamu"):
         raise TypeError("areas must be a str or a list/tuple/set of str")
     if len(areas) == 0:
         raise ValueError("areas must be non-empty")
-    all_areas = (
-        mi.zones["loadzone"]
-        | mi.zones["abv"]
-        | mi.zones["state"]
-        | mi.zones["interconnect"]
-    )
+    all_areas = set().union(*(mi.zones[z] for z in mi.zones["mappings"]))
     if not areas <= all_areas:
         diff = areas - all_areas
         raise ValueError("invalid area(s): %s" % " | ".join(diff))
@@ -335,21 +331,20 @@ def _check_areas_and_format(areas, grid_model="usa_tamu"):
     abv_in_areas = [z for z in areas if z in mi.zones["abv"]]
     for a in abv_in_areas:
         areas.remove(a)
-        areas.add(mi.zones["abv2state"][a])
+        areas.add(mi.zones[f"abv2{division}"][a])
 
     return areas
 
 
-def _check_resources_and_format(resources, grid_model="usa_tamu"):
+def _check_resources_and_format(resources, mi=ModelImmutables("usa_tamu")):
     """Ensure that resources are valid and convert variable to a set.
 
     :param str/list/tuple/set resources: resource(s) to check.
-    :param str grid_model: grid model.
+    :param powersimdata.network.model.ModelImmutables mi: immutables of a grid model.
     :raises TypeError: if resources is not a list/tuple/set of str.
     :raises ValueError: if resources is empty or not valid.
     :return: (*set*) -- resources as a set.
     """
-    mi = ModelImmutables(grid_model)
     if isinstance(resources, str):
         resources = {resources}
     elif isinstance(resources, (list, set, tuple)):
@@ -366,17 +361,17 @@ def _check_resources_and_format(resources, grid_model="usa_tamu"):
     return resources
 
 
-def _check_resources_are_renewable_and_format(resources, grid_model="usa_tamu"):
+def _check_resources_are_renewable_and_format(
+    resources, mi=ModelImmutables("usa_tamu")
+):
     """Ensure that resources are valid renewable resources and convert variable to
     a set.
-
+    :param powersimdata.network.model.ModelImmutables mi: immutables of a grid model.
     :param str/list/tuple/set resources: resource(s) to analyze.
-    :param str grid_model: grid model.
     :raises ValueError: if resources are not renewables.
     return: (*set*) -- resources as a set
     """
-    mi = ModelImmutables(grid_model)
-    resources = _check_resources_and_format(resources, grid_model=grid_model)
+    resources = _check_resources_and_format(resources, mi=mi)
     if not resources <= mi.plants["renewable_resources"]:
         diff = resources - mi.plants["all_resources"]
         raise ValueError("invalid renewable resource(s): %s" % " | ".join(diff))
--- a/powersimdata/input/check.py
+++ b/powersimdata/input/check.py
@@ -308,11 +308,9 @@ def _check_areas_and_format(areas, mi=ModelImmutables("usa_tamu")):
     :param powersimdata.network.model.ModelImmutables mi: immutables of a grid model.
     :raises TypeError: if ``areas`` is not a list/tuple/set of str.
     :raises ValueError: if ``areas`` is empty or not valid.
-    :return: (*set*) -- areas as a set. State abbreviations are converted to state
-        names.
+    :return: (*set*) -- areas as a set. State/Country abbreviations are converted to
+        state/country names.
     """
-    division = [a for a in mi.zones["mappings"] if "abbr" in a][0].split("_")[0]
-
     if isinstance(areas, str):
         areas = {areas}
     elif isinstance(areas, (list, set, tuple)):
@@ -331,7 +329,7 @@ def _check_areas_and_format(areas, mi=ModelImmutables("usa_tamu")):
     abv_in_areas = [z for z in areas if z in mi.zones["abv"]]
     for a in abv_in_areas:
         areas.remove(a)
-        areas.add(mi.zones[f"abv2{division}"][a])
+        areas.add(mi.zones[f"abv2{mi.zones['division']}"][a])
 
     return areas
 
--- a/powersimdata/input/tests/test_check.py
+++ b/powersimdata/input/tests/test_check.py
@@ -24,7 +24,16 @@ from powersimdata.input.check import (
     _check_time_series,
     check_grid,
 )
+from powersimdata.network.europe_tub.model import TUB
+from powersimdata.network.model import ModelImmutables
 from powersimdata.tests.mock_scenario import MockScenario
+
+
+@pytest.fixture
+def europe():
+    tub = TUB("Europe", reduction=128)
+    tub.build()
+    return tub
 
 
 @pytest.fixture
@@ -218,16 +227,25 @@ def test_check_areas_and_format_argument_value():
             _check_areas_and_format(a)
 
 
-def test_check_areas_and_format():
+def test_check_areas_and_format(europe):
     _check_areas_and_format(["Western", "NY", "El Paso", "Arizona"])
-    areas = _check_areas_and_format(["California", "CA", "NY", "TX", "MT", "WA"])
-    assert areas == {"Washington", "Texas", "Montana", "California", "New York"}
+    assert _check_areas_and_format(["California", "CA", "NY", "TX", "MT", "WA"]) == {
+        "Washington",
+        "Texas",
+        "Montana",
+        "California",
+        "New York",
+    }
+    assert _check_areas_and_format(
+        ["FR", "Germany", "Slovakia", "IT", "Italy", "NO", "Norway", "Nordic", "GB5 9"],
+        mi=europe.model_immutables,
+    ) == {"France", "Germany", "Slovakia", "Italy", "Norway", "Nordic", "GB5 9"}
 
 
 def test_check_resources_and_format_argument_type():
     arg = (
         1,
-        {"coal": [1, 2, 3], "htdro": [4, 5, 6]},
+        {"coal": [1, 2, 3], "hydro": [4, 5, 6]},
         [1, 2, 3, 4],
         (1, 2, 3, 4),
         {1, 2, 3, 4},
@@ -247,8 +265,9 @@ def test_check_resources_and_format_argument_value():
 
 def test_check_resources_and_format():
     _check_resources_and_format(["dfo", "wind", "solar", "ng"])
-    _check_resources_and_format("wind_offshore")
+    _check_resources_and_format("wind_offshore", mi=ModelImmutables("europe_tub"))
     _check_resources_and_format({"nuclear"})
+    _check_resources_and_format("geothermal", mi=ModelImmutables("europe_tub"))
 
 
 def test_check_resources_are_renewable_and_format_argument_value():
@@ -260,6 +279,9 @@ def test_check_resources_are_renewable_and_format():
     _check_resources_are_renewable_and_format(["wind_offshore", "wind"])
     _check_resources_are_renewable_and_format("solar")
     _check_resources_are_renewable_and_format({"wind"})
+    _check_resources_are_renewable_and_format(
+        {"solar"}, mi=ModelImmutables("europe_tub")
+    )
 
 
 def test_check_areas_are_in_grid_and_format_argument_type(mock_grid):
--- a/powersimdata/network/constants/region/europe.py
+++ b/powersimdata/network/constants/region/europe.py
@@ -1,9 +1,9 @@
-from itertools import combinations
+from itertools import chain
 
 import pandas as pd
 
 from powersimdata.network.constants.model import model2interconnect
-from powersimdata.network.helpers import interconnect_to_name
+from powersimdata.network.helpers import interconnect_to_name, powerset
 
 abv2country = {
     "AL": "Albania",
@@ -77,7 +77,6 @@ abv2timezone = {
     "SK": "ETC/GMT-1",
 }
 
-
 interconnect2abv = {
     "ContinentalEurope": {
         "AL",
@@ -111,27 +110,22 @@ interconnect2abv = {
     "Ireland": {"IE"},
     "Baltic": {"EE", "LT", "LV"},
 }
-
-cb = [i for j in range(2, 6) for i in combinations(model2interconnect["europe_tub"], j)]
-for c in cb:
-    interconnect2abv[interconnect_to_name(c, model="europe_tub")] = {
-        a for i in c for a in interconnect2abv[i]
-    }
+for c in powerset(model2interconnect["europe_tub"], 2):
+    interconnect2abv[interconnect_to_name(c, model="europe_tub")] = set(
+        chain(*[interconnect2abv[i] for i in c])
+    )
 
 name2interconnect = {
-    interconnect_to_name(i, model="europe_tub"): set(i)
-    for c in range(1, 6)
-    for i in combinations(model2interconnect["europe_tub"], c)
+    interconnect_to_name(c, model="europe_tub"): set(c)
+    for c in powerset(model2interconnect["europe_tub"], 1)
 }
 
 name2component = name2interconnect.copy()
 name2component.update({"Europe": set(name2interconnect) - {"Europe"}})
 
-
 interconnect2timezone = {
-    interconnect_to_name(i, model="europe_tub"): "ETC/GMT-1"
-    for c in range(1, 6)
-    for i in combinations(model2interconnect["europe_tub"], c)
+    interconnect_to_name(c, model="europe_tub"): "ETC/GMT-1"
+    for c in powerset(model2interconnect["europe_tub"], 1)
 }
 interconnect2timezone.update(
     {
--- a/powersimdata/network/constants/region/europe.py
+++ b/powersimdata/network/constants/region/europe.py
@@ -1,0 +1,257 @@
+from itertools import combinations
+
+import pandas as pd
+
+from powersimdata.network.constants.model import model2interconnect
+from powersimdata.network.helpers import interconnect_to_name
+
+abv2country = {
+    "AL": "Albania",
+    "AT": "Austria",
+    "BA": "Bosnia And Herzegovina",
+    "BE": "Belgium",
+    "BG": "Bulgaria",
+    "CH": "Switzerland",
+    "CZ": "Czech Republic",
+    "DE": "Germany",
+    "DK": "Danemark",
+    "EE": "Estonia",
+    "ES": "Spain",
+    "FI": "Finland",
+    "FR": "France",
+    "GB": "Great Britain",
+    "GR": "Greece",
+    "HR": "Croatia",
+    "HU": "Hungary",
+    "IE": "Ireland",
+    "IT": "Italy",
+    "LT": "Lithuania",
+    "LU": "Luxembourg",
+    "LV": "Latvia",
+    "ME": "Montenegro",
+    "MK": "Macedonia",
+    "NL": "Netherlands",
+    "NO": "Norway",
+    "PL": "Poland",
+    "PT": "Portugal",
+    "RO": "Romania",
+    "RS": "Serbia",
+    "SE": "Sweden",
+    "SI": "Slovenia",
+    "SK": "Slovakia",
+}
+
+abv2timezone = {
+    "AL": "ETC/GMT-1",
+    "AT": "ETC/GMT-1",
+    "BA": "ETC/GMT-1",
+    "BE": "ETC/GMT-1",
+    "BG": "ETC/GMT-2",
+    "CH": "ETC/GMT-1",
+    "CZ": "ETC/GMT-1",
+    "DE": "ETC/GMT-1",
+    "DK": "ETC/GMT-1",
+    "EE": "ETC/GMT-2",
+    "ES": "ETC/GMT-1",
+    "FI": "ETC/GMT-2",
+    "FR": "ETC/GMT-1",
+    "GB": "ETC/GMT",
+    "GR": "ETC/GMT-2",
+    "HR": "ETC/GMT-1",
+    "HU": "ETC/GMT-1",
+    "IE": "ETC/GMT",
+    "IT": "ETC/GMT-1",
+    "LT": "ETC/GMT-2",
+    "LU": "ETC/GMT-1",
+    "LV": "ETC/GMT-2",
+    "ME": "ETC/GMT-1",
+    "MK": "ETC/GMT-1",
+    "NL": "ETC/GMT-1",
+    "NO": "ETC/GMT-1",
+    "PL": "ETC/GMT-1",
+    "PT": "ETC/GMT",
+    "RO": "ETC/GMT-2",
+    "RS": "ETC/GMT-1",
+    "SE": "ETC/GMT-1",
+    "SI": "ETC/GMT-1",
+    "SK": "ETC/GMT-1",
+}
+
+
+interconnect2abv = {
+    "ContinentalEurope": {
+        "AL",
+        "AT",
+        "BA",
+        "BE",
+        "BG",
+        "CH",
+        "CZ",
+        "DE",
+        "DK",
+        "ES",
+        "FR",
+        "GR",
+        "HR",
+        "HU",
+        "IT",
+        "LU",
+        "ME",
+        "MK",
+        "NL",
+        "PL",
+        "PT",
+        "RO",
+        "RS",
+        "SI",
+        "SK",
+    },
+    "Nordic": {"FI", "NO", "SE"},
+    "GreatBritain": {"GB"},
+    "Ireland": {"IE"},
+    "Baltic": {"EE", "LT", "LV"},
+}
+
+cb = [i for j in range(2, 6) for i in combinations(model2interconnect["europe_tub"], j)]
+for c in cb:
+    interconnect2abv[interconnect_to_name(c, model="europe_tub")] = {
+        a for i in c for a in interconnect2abv[i]
+    }
+
+name2interconnect = {
+    interconnect_to_name(i, model="europe_tub"): set(i)
+    for c in range(1, 6)
+    for i in combinations(model2interconnect["europe_tub"], c)
+}
+
+name2component = name2interconnect.copy()
+name2component.update({"Europe": set(name2interconnect) - {"Europe"}})
+
+
+interconnect2timezone = {
+    interconnect_to_name(i, model="europe_tub"): "ETC/GMT-1"
+    for c in range(1, 6)
+    for i in combinations(model2interconnect["europe_tub"], c)
+}
+interconnect2timezone.update(
+    {
+        interconnect_to_name("GreatBritain", model="europe_tub"): "ETC/GMT",
+        interconnect_to_name("Ireland", model="europe_tub"): "ETC/GMT",
+        interconnect_to_name(
+            ["GreatBritain", "Ireland"], model="europe_tub"
+        ): "ETC/GMT",
+        interconnect_to_name("Baltic", model="europe_tub"): "ETC/GMT-2",
+        interconnect_to_name(["Nordic", "Baltic"], model="europe_tub"): "ETC/GMT-2",
+    }
+)
+
+
+def get_interconnect_mapping(zone, model):
+    """Return interconnect mapping.
+
+    :param pandas.DataFrame zone: information on zones of a grid model.
+    :param str model: the grid model.
+    :return: (*dict*) -- mappings of interconnect to other areas.
+    """
+    mapping = dict()
+
+    name = interconnect_to_name(zone["interconnect"], model=model)
+
+    mapping["interconnect"] = name2component[name] | {name}
+    mapping["name2interconnect"] = {
+        i: name2interconnect[i] for i in mapping["interconnect"]
+    }
+    mapping["name2component"] = {i: name2component[i] for i in mapping["interconnect"]}
+    mapping["interconnect2timezone"] = {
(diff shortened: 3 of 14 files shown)


powersimdata/input/check.py
```diff
@@ -7,6 +7,7 @@
 
 # Importing the module, not anything in it, to avid a circular import
 import powersimdata.input.grid as _grid
+from powersimdata.network.europe_tub.model import TUB
```
jon-hagg, line 10: I'm not sure (haven't tested) if this causes the circular import issue, but regardless I wonder if this module should "know" about specific grid models. It's more of a design thing than a future concern, but suppose we had another model like the europe one which takes a `reduction` as a constructor argument. In that case we wouldn't want to hard code this module to use `TUB`. One way around this (just the first thing that popped into my head) is taking the `ModelImmutables` instance instead of the `reduction`, so we don't have to decide _here_ which grid model it would be, shifting that responsibility to the caller.

rouille: It makes sense.

rouille: Done. Let me know what you think.

powersimdata/network/europe_tub/model.py
```diff
@@ -0,0 +1,140 @@
...
+            network = pypsa.Network(os.path.join(path, f"elec_s_{self.reduction}.nc"))
+        else:
+            raise ValueError(
+                "Invalid Resolution. Choose among: None | 1024 | 512 | 256 | 128 | 37"
+            )
+        id2zone = {i: l for i, l in enumerate(network.buses.index)}
+        zone2id = {l: i for i, l in id2zone.items()}
+
+        if "Europe" in self.interconnect:
+            self.network = network
```
BainanXia, line 72: So we will have `self.network` which contains all the tables instead of `self.plant`, `self.branch`, `self.bus`, etc in other grid models. How this would impact the downstream functions in `PostREISE` and will this be compatible with [export_to_pypsa() ](https://github.com/Breakthrough-Energy/PowerSimData/blob/5682629dec9f8b7470bf32335a7f74b2310df076/powersimdata/input/export_data.py#L289)?

powersimdata/network/model.py
```diff
@@ -34,75 +37,59 @@ def __init__(self, model, interconnect=None):
...
-        elif "interconnect" in mappings and area in zones["interconnect2loadzone"]:
-            loadzone_set = zones["interconnect2loadzone"][area]
-        elif area == "all":
-            loadzone_set = zones["loadzone"]
-        else:
-            print("%s is incorrect." % area)
+        zones["all"] = "all"
+        loadzone = set().union(
+            *(area2loadzone[a](area) for a in area2loadzone if area in zones[a])
+        )
```
BainanXia, line 91: Brilliant! I didn't realize `state` is always the superset of load zones among the mappings of a unique keyword, i.e. if `area='North Carolina'`, we first evaluate it as a state, which gives two load zones `North Carolina and Western North Carolina`, then we evaluate it as a loadzone, which gives `North Carolina` only, finally the union won't change the result, same applies for `Texas` since $ERCOT\ \subseteq \ Texas$ in terms of load zones. Hence the if/else logic in the original implementation is equivalent to taking the union among the results of all searches. The only thing is in this way we wouldn't need `area2loadzone` to be an `OrderedDict` anymore, right?

powersimdata/network/tests/test_model.py
```diff
@@ -0,0 +1,76 @@
...
+        area_to_loadzone("usa_tamu", "California", area_type="loadzone")
+
+    with pytest.raises(ValueError, match="Invalid area / area_type combination"):
+        area_to_loadzone("usa_tamu", "WA", area_type="interconnect")
+
+    with pytest.raises(ValueError, match="Invalid area / area_type combination"):
+        area_to_loadzone("europe_tub", "France", area_type="country_abbr")
+
+
+def test_area_to_loadzone():
```
BainanXia, line 28: I think we should be able to compare two sets directly by `assert set(A) == set(B)` instead of looking into the length of difference being 0?

BainanXia (commented):
This looks pretty good in general and thanks for adding the tests.

powersimdata/input/check.py
```diff
@@ -299,19 +299,23 @@ def _check_grid_type(grid):
...
     converted to their actual name.
 
     :param str/list/tuple/set areas: areas(s) to check. Could be load zone name(s),
         state name(s)/abbreviation(s) or interconnect(s).
-    :param str grid_model: grid model.
-    :raises TypeError: if areas is not a list/tuple/set of str.
-    :raises ValueError: if areas is empty or not valid.
+    :param powersimdata.network.model.ModelImmutables mi: immutables of a grid model.
+    :param int reduction: reduction parameter (number of nodes in network). If None,
+        the full network is loaded. Applies to the *'europe_tub'* grid model.
```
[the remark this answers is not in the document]
rouille: I forgot to remove the entry in the docstring after modifying the design following @jon-hagg suggestion (see [comment](https://github.com/Breakthrough-Energy/PowerSimData/pull/631#discussion_r876562730)).

The model immutables is all you need for the check function. It is an attribute of the model class (`TAMU`, `HIFLD`, `TUB` and future others) and of a `Grid` object. If a network is reduced or interconnects/states/countries/etc are dropped from the network, the model immutables will reflect that

powersimdata/input/check.py
```diff
@@ -299,19 +299,23 @@ def _check_grid_type(grid):
...
+    :raises TypeError: if ``areas`` is not a list/tuple/set of str.
+    :raises ValueError:
+        if ``areas`` is empty or not valid.
```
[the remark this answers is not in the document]
rouille: No we don't.

powersimdata/input/check.py
```diff
@@ -299,19 +299,23 @@ def _check_grid_type(grid):
...
     :return: (*set*) -- areas as a set. State abbreviations are converted to state
         names.
     """
-    mi = ModelImmutables(grid_model)
+    division = [a for a in mi.zones["mappings"] if "abbr" in a][0].split("_")[0]
```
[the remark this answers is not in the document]
rouille: I agree, it is a quick and dirty to distinguish US models from European model (state vs country). I will work on it.

powersimdata/input/tests/test_check.py
```diff
@@ -218,16 +227,25 @@ def test_check_areas_and_format_argument_value():
...
+        ["FR", "Germany", "Slovakia", "IT", "Italy", "NO", "Norway", "Nordic", "GB5 9"],
+        mi=europe.model_immutables,
+    ) == {"France", "Germany", "Slovakia", "Italy", "Norway", "Nordic", "GB5 9"}
 
 
 def test_check_resources_and_format_argument_type():
     arg = (
         1,
-        {"coal": [1, 2, 3], "htdro": [4, 5, 6]},
+        {"coal": [1, 2, 3], "hydro": [4, 5, 6]},
```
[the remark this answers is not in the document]
rouille: The error was raise anyway

powersimdata/network/constants/region/europe.py
```diff
@@ -0,0 +1,258 @@
...
+    "Nordic": {"FI", "NO", "SE"},
+    "GreatBritain": {"GB"},
+    "Ireland": {"IE"},
+    "Baltic": {"EE", "LT", "LV"},
+}
+
+cb = [i for j in range(2, 6) for i in combinations(model2interconnect["europe_tub"], j)]
+for c in cb:
+    interconnect2abv[interconnect_to_name(c, model="europe_tub")] = {
+        a for i in c for a in interconnect2abv[i]
```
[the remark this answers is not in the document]
rouille: Thanks

powersimdata/network/constants/region/europe.py
```diff
@@ -0,0 +1,258 @@
...
+
+name2component = name2interconnect.copy()
+name2component.update({"Europe": set(name2interconnect) - {"Europe"}})
+
+
+interconnect2timezone = {
+    interconnect_to_name(i, model="europe_tub"): "ETC/GMT-1"
+    for c in range(1, 6)
+    for i in combinations(model2interconnect["europe_tub"], c)
+}
```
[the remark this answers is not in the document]
rouille: Sounds good. It could be used for US models too

powersimdata/network/constants/region/europe.py
```diff
@@ -0,0 +1,258 @@
...
+    """Return zone constants.
+
+    :para list interconnect: interconnect(s).
+    :param str model: the grid model.
+    :return: (*dict*) -- zones information.
+    """
+    zones = dict()
+    zones["mappings"] = {"loadzone", "country", "country_abbr", "interconnect"}
+
+    query = (  # noqa: F841
```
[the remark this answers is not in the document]
rouille: Yes we should

powersimdata/network/constants/region/europe.py
```diff
@@ -0,0 +1,258 @@
...
+        model2interconnect[model] if "Europe" in interconnect else interconnect
+    )
+    {"abv": {a for i in query for a in interconnect2abv[i]}}
```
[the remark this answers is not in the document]
rouille: Nothing. I believe I copied/pasted from the US modules and forgot to remove it. Thanks.

powersimdata/network/europe_tub/model.py
```diff
@@ -0,0 +1,140 @@
...
+    interconnect_to_name,
+)
+from powersimdata.network.model import ModelImmutables
+from powersimdata.utility.helpers import _check_import
+
+pypsa = _check_import("pypsa")
+zenodo_get = _check_import("zenodo_get")
+
+
+class TUB:
```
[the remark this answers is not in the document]
rouille: I assumed that eventual storage was included in the `Network` object. If not, we should at least get the empty attributes (a place holder that we will fill via a change table) when we use the function in #608 to translate a `Network` object to a `Grid` object

powersimdata/network/europe_tub/model.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+
+    def __init__(self, interconnect, reduction=None, overwrite=False):
+        """Constructor."""
+        self.grid_model = "europe_tub"
+        self.interconnect = check_and_format_interconnect(
+            interconnect, model=self.grid_model
+        )
+        self.data_loc = os.path.join(os.path.dirname(__file__), "data")
+        self.zenodo_record_id = "3601881"
```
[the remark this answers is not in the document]
rouille: It should

powersimdata/network/europe_tub/model.py
```diff
@@ -0,0 +1,140 @@
...
+        self.reduction = reduction
+
+        if overwrite:
+            self.remove_data()
+
+        self.retrieve_data()
```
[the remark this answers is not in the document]
rouille: If the data are already on disk, it will look at the checksum and only download if there is a newer version:
```
>>> from powersimdata.network.europe_tub.model import TUB
>>> tub = TUB("Nordic", reduction=128)
Title: PyPSA-Eur: An Open Optimisation Model of the European Transmission System (Dataset)
Keywords: power system model, capacity expansion model, energy system model, electricity system model, python
Publication date: 2021-09-22
DOI: 10.5281/zenodo.5521712
Total size: 1767.1 MB

Link: https://zenodo.org/api/files/6e58c6dc-083b-46c8-8255-8516e3eb0ed8/networks.zip   size: 1767.1 MB
networks.zip is already downloaded correctly.
All files have been downloaded.
```
We can improve the logic. For example, we don't have to unzip every time we create an object (I tested it and we would gain few ms if we don't do it).

powersimdata/network/europe_tub/model.py
```diff
@@ -0,0 +1,140 @@
...
+        path = os.path.join(self.data_loc, "networks")
+        if self.reduction is None:
+            network = pypsa.Network(os.path.join(path, "elec_s.nc"))
+        elif os.path.exists(os.path.join(path, f"elec_s_{self.reduction}.nc")):
+            network = pypsa.Network(os.path.join(path, f"elec_s_{self.reduction}.nc"))
+        else:
+            raise ValueError(
+                "Invalid Resolution. Choose among: None | 1024 | 512 | 256 | 128 | 37"
+            )
+        id2zone = {i: l for i, l in enumerate(network.buses.index)}
```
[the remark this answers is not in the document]
rouille: No preference

powersimdata/network/europe_tub/model.py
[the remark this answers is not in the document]
BainanXia: Ah, I didn't look into the doc of `zenodo_get`. That's nice.

powersimdata/network/europe_tub/model.py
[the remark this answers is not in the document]
BainanXia: Me neither;)

powersimdata/network/europe_tub/model.py
rouille: It is up to discussion. Where do we want to convert the `Network` object to a `Grid` object? In PR #632 where I refactor the `_Builder` to have a base class (`_Builder`) and two child classes (`FromCSV` and from `PyPSA`), see: https://github.com/Breakthrough-Energy/PowerSimData/blob/67902e65440daf20cc18adf92af1b3e24544bb83/powersimdata/scenario/create.py#L419 we get the `Grid` and the profiles from the `Network`, and then we can populate the change table and transform the grid and profiles accordingly. The workflow is:
```

scenario = Scenario()

scenario.set_grid(grid_model="europe_tub", interconnect="Nordic")

scenario.set_name("test", "dummy")
scenario.set_time("2016-08-01 00:00:00", "2016-08-31 23:00:00", "24H")
scenario.change_table.scale_plant_capacity(...)
```

powersimdata/network/helpers.py
```diff
@@ -57,7 +57,12 @@ def interconnect_to_name(interconnect, model="hifld"):
     :param str model: the grid model.
     :return: (*str*): name of grid model.
     """
-    return "_".join(sorted(check_and_format_interconnect(interconnect, model=model)))
+    return "_".join(
+        sorted(
+            i.replace(" ", "")
```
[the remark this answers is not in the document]
rouille: It allows the user to do instantiate the model with `Great Britain` or `Continental Europe` for an interconnect. We will be able to do:
```

scenario = Scenario()

scenario.set_grid(grid_model="europe_tub", interconnect="Continental Europe")
```

powersimdata/network/tests/test_model.py
rouille: Indeed

powersimdata/network/europe_tub/model.py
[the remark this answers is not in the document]
danielolsen: Storage units for a Network object live in the `storage_units` attribute. If there are no storage units, this attribute will be an empty dataframe with defined columns.

NB: wthin PyPSA, "storage units" are distinct from "stores". I believe that the distinction is that a Store is purely an 'energy-holder', while a StorageUnit combines an energy-holder within details on how energy can flow into and out of the energy-holder. I believe that the functionality of a StorageUnit can be emulated via Stores and Links.

I think that within PowerSimData, it should be straightforward to translate a StorageUnit to our storage schema, but finding a way to represent a Store in the general case may be tough.

powersimdata/network/model.py
danielolsen: > $ERCOT\ \subseteq \ Texas$

Wow, fancy!

rouille: I think we need an `OrderedDict` here to guarantee we loop in the right order as described in the docstring. Not sure I completely understand your question.

powersimdata/network/tests/test_model.py
danielolsen: Are we testing equality here, or whether one is a subset of the other? Maybe we want to be checking `assert set(A) <= set(B)`

rouille: We are testing equality. I was not sure `==` would be able to compare sets properly, that is why I took the difference and look at the length

powersimdata/network/model.py
BainanXia: We are going to union the results no matter in what order it loops through, right? Previously I thought this might end up with different behavior from the original implementation, but thinking it twice I found the behavior is the same since we don't have a loadzone with the same name as a state and defines a larger area.

powersimdata/network/helpers.py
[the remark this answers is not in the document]
BainanXia: I see. Then the user can also do "C o n t          in e  n t a l    Euro p  e", which is okay I think.

powersimdata/network/europe_tub/model.py
BainanXia: I see. I didn't look into #635 yet. So the higher level workflow when a user wants to run a `europe_tub` model would be: a) instantiate a scenario and set the grid model to be `europe_tub` b) add modifications to change table c) run with PyPSA engine (CEM) (in this case, we will need to convert the scenario object back to a network object upon feeding into PyPSA) or REISE.jl engine (PCM). Then in step a) we will instantiate a grid object from a network object. The grid is built via `FromCSV` when the model is `TAMU` or `HIFLD` and via `FromPyPSA` when the model is `TUB`. Given the implementation of `TUB` (the build function) here, when instantiating the grid object based on it, we will need `tub_network = TUB(interconnect) self.bus = tub_network.network.buses` instead of `hifld_network = HIFLD(interconnect) self.bus = hifld_network.bus`, right?

powersimdata/network/helpers.py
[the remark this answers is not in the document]
rouille: It should work! I will add some tests.

powersimdata/network/model.py
rouille: I understand now. I did not think too much about it, I followed the same search sequence than in the previous implementation. What you say makes sense and I will use a standard dict.

powersimdata/network/europe_tub/model.py
rouille: We will do as usual for usa_tamu (and hifld) and for europe_tub, we will create a ˋTUB` object, build the network from zenodo's NetCDF files (for now) via ˋtub = TUB(interconnect)` and ˋtub.build()` and pass ˋtub.network` to the constructor of the ˋFromPyPSAˋ class in #608 (I realize this class has the same name than the child class of the builder) to get a grid object. Not sure how we will get the profile yet.

BainanXia: > We will do as usual for usa_tamu (and hifld) and for europe_tub, we will create a ˋTUB` object, build the network from zenodo's NetCDF files (for now) via ˋtub = TUB(interconnect)` and ˋtub.build()` and pass ˋtub.network` to the constructor of the ˋFromPyPSAˋ class in #608 (I realize this class has the same name than the child class of the builder) to get a grid object. Not sure how we will get the profile yet.

That makes perfect sense. Thanks!

powersimdata/input/check.py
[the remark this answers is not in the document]
rouille: Done

powersimdata/input/check.py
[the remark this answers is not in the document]
rouille: Done

powersimdata/network/constants/region/europe.py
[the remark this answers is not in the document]
rouille: Done

powersimdata/network/constants/region/europe.py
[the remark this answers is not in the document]
rouille: Done

powersimdata/network/tests/test_model.py
rouille: Done

powersimdata/network/model.py
rouille: Done

powersimdata/network/helpers.py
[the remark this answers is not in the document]
rouille: In fact *Continental Europe* (or any other interconnect with a space in it) could not be passed because the `check_and_format_interconnect` function in the same module would raise an error if the interconnect is not defined in the `model2interconnect` dictionary in the `powersimdata.network.constants.model` module.

I have updated the `check_and_format_interconnect` function to be able to pass both *Continental Europe* and *Continental Europe* and added some tests.

powersimdata/network/constants/region/europe.py
```diff
@@ -0,0 +1,257 @@
...
+        "RO",
+        "RS",
+        "SI",
+        "SK",
+    },
+    "Nordic": {"FI", "NO", "SE"},
+    "GreatBritain": {"GB"},
+    "Ireland": {"IE"},
+    "Baltic": {"EE", "LT", "LV"},
+}
```
danielolsen, line 113: Could we combine all of this information into a CSV, indexed by the country abbreviation, with columns for the long name, the time zone, and the interconnect? We would need to do a bit of reverse-mapping to get `interconnect2abv` back but I don't think it would be too bad.

powersimdata/network/europe_tub/model.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+class TUB:
+    """PyPSA Europe network.
+
+    :param str/iterable interconnect: interconnect name(s).
+    :param int reduction: reduction parameter (number of nodes in network). If None,
+        the full network is loaded.
+    :param bool overwrite: the existing dataset is deleted a new dataset is downloaded
+        from zenodo.
```
danielolsen, line 28: `the existing dataset is deleted a new dataset is downloaded` -> `the existing dataset is deleted and a new dataset is downloaded`?

powersimdata/network/europe_tub/model.py
```diff
@@ -0,0 +1,140 @@
...
+        elif os.path.exists(os.path.join(path, f"elec_s_{self.reduction}.nc")):
+            network = pypsa.Network(os.path.join(path, f"elec_s_{self.reduction}.nc"))
+        else:
+            raise ValueError(
+                "Invalid Resolution. Choose among: None | 1024 | 512 | 256 | 128 | 37"
+            )
+        id2zone = {i: l for i, l in enumerate(network.buses.index)}
+        zone2id = {l: i for i, l in id2zone.items()}
+
+        if "Europe" in self.interconnect:
```
danielolsen, line 71: Here and elsewhere we do this check `if "Europe" in self.interconnect:` to see if we have the whole region. It seems from `check_and_format_interconnect` that if this is the case, `self.interconnect` will always be `["Europe"]`; should we check more clearly `if self.interconnect == ["Europe"]:`?

powersimdata/network/europe_tub/model.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # loadzone
+        mapping.zones["loadzone"] = set(self.zone2id)
+        mapping.zones["id2loadzone"] = self.id2zone
+        mapping.zones["loadzone2id"] = self.zone2id
+        mapping.zones["loadzone2abv"] = self.network.buses["country"].to_dict()
+        mapping.zones["loadzone2country"] = {
+            l: abv2country[a]
+            for l, a in self.network.buses["country"].to_dict().items()
+        }
```
danielolsen, line 99: Is this equivalent to `mapping.zones["loadzone2country"] = self.network.buses["country"].map(abv2country).to_dict()`?

powersimdata/network/europe_tub/model.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # country
+        mapping.zones["country2loadzone"] = {
+            abv2country[a]: set(l)
+            for a, l in self.network.buses.groupby("country").groups.items()
+        }
+        mapping.zones["abv2loadzone"] = {
+            a: set(l) for a, l in self.network.buses.groupby("country").groups.items()
+        }
```
danielolsen, line 119: Is this equivalent to
```python
mapping.zones["abv2loadzone"] = (
    self.network.buses.groupby("country").groups.map(set).to_dict()
)
```
?

powersimdata/network/model.py
```diff
@@ -34,75 +35,57 @@ def __init__(self, model, interconnect=None):
...
+    if not isinstance(area, str):
+        raise TypeError("area must be a str")
 
     if area_type is not None and not isinstance(area_type, str):
-        raise TypeError("'area_type' should be either None or str.")
+        raise TypeError("area_type must be either None or str")
+
+    division = [a for a in mappings if "abbr" in a][0].split("_")[0]
+    area2loadzone = {
+        f"{division}": lambda x: zones[f"{division}2loadzone"][x],
```
danielolsen, line 70: Is `lambda x: zones[f"{division}2loadzone"][x]` equivalent to `zones[f"{division}2loadzone"].get`? If so, I think we could also simplify the values for `f"{division}_abbr"` and `"interconnect"`.

powersimdata/input/check.py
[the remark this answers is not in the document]
rouille: @BainanXia and @danielolsen, can you look at 56d1a464489d3e17d59717e5085d9aeebbe4c4c7 and tell me if you like the implementation?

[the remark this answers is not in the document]
danielolsen: It should work for the current use case, where we have exactly three items in the list `["interconnect", SOMETHING, "loadzone"]`. I'm not sure how it would extend if we wanted to be able to add sub-country divisions, but maybe that's far beyond what we need right now. Either way it's definitely cleaner than the list comprehension solution.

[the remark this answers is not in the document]
rouille: I thought about that. We will have to augment the hierarchy with new areas (counties, province) and decide which area we want to map to/from loadzone.

powersimdata/network/europe_tub/model.py
rouille: Good catch

powersimdata/network/constants/region/europe.py
rouille: We can do that. We would have a *zone.csv* in the *powersimdata/network/europe_tub/data* folder?

powersimdata/network/europe_tub/model.py
rouille: This gives me an `AttributeError`:
```
AttributeError: 'PrettyDict' object has no attribute 'map'
```

danielolsen: Ah, right, `.groups` is a `PrettyDict`, not a pandas `Series`.

powersimdata/network/europe_tub/model.py
rouille: Done

powersimdata/network/europe_tub/model.py
rouille: Done

powersimdata/network/europe_tub/model.py
rouille: Done

powersimdata/network/europe_tub/model.py
rouille: I have kept the current implementation

powersimdata/input/check.py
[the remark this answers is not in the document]
rouille: I kept things simple and added a `division` key in the `zones` attribute of a `ModelImmutables` object with value set to `state` or `country` depending on the grid model. This will do the job for now.

powersimdata/network/constants/region/europe.py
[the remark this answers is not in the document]
rouille: Done

powersimdata/network/constants/region/europe.py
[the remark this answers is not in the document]
rouille: Done

powersimdata/network/model.py
rouille: Done

BainanXia (approved):
This is nice and good to go. I agree with @danielolsen [suggestion](https://github.com/Breakthrough-Energy/PowerSimData/pull/631#discussion_r879909958) on combining those big dictionaries into a csv which will make it look cleaner. Thanks!

PR #635, opened 2022-05-13
Refactor builder

### LINKED ISSUE #632: Expand building phase of a scenario to grid models in form of PyPSA Network object
- [x] Base class is named  `_Builder` 
- [x] Child classes are `FromCSV` and `FromPyPSA`
- [x] Factory method creates the desired object based in the `set_grid` method of the `Create` class using the `grid_model` parameter.

rouille:

### Purpose
Refactor the building phase of a scenario since we will in the near future use grid model in the form of a PyPSA `Network` object. Closes #632.

### What the code is doing
Allow to build a scenario from a PyPSA `Network` object

### Testing
Existing unit tests. Especially, the `powersimdata.scenario.tests.test_create`

### Where to look
The `powersimdata.scenario.create` module

### Usage Example/Visuals
N/A

### Time estimate
15 min

### DIFF
--- a/powersimdata/scenario/create.py
+++ b/powersimdata/scenario/create.py
@@ -1,5 +1,4 @@
 import copy
-import warnings
 
 import numpy as np
 import pandas as pd
@@ -29,7 +28,6 @@ class Create(State):
     default_exported_methods = {
         "create_scenario",
         "get_bus_demand",
-        "set_builder",
         "set_grid",
     } | State.exported_methods
 
@@ -130,37 +128,23 @@ class Create(State):
             )
             self.switch(Execute)
 
-    def set_builder(self, *args, **kwargs):
-        """Alias to :func:`~powersimdata.scenario.create.Create.set_grid`"""
-        warnings.warn(
-            "set_builder is deprecated, use set_grid instead", DeprecationWarning
-        )
-        self.set_grid(*args, **kwargs)
-
-    def set_grid(self, grid_model="usa_tamu", interconnect="USA"):
+    def set_grid(self, grid_model="usa_tamu", interconnect="USA", **kwargs):
         """Sets grid builder.
 
         :param str grid_model: name of grid model. Default is *'usa_tamu'*.
         :param str/list interconnect: name of interconnect(s). Default is *'USA'*.
+        :param \\*\\*kwargs: optional parameters used to instantiate a Grid object.
         """
-        self.builder = _Builder(
-            grid_model, interconnect, self._scenario_list_manager.get_scenario_table()
+        scenario_table = self._scenario_list_manager.get_scenario_table()
+
+        self.builder = get_builder(
+            grid_model,
+            interconnect,
+            scenario_table,
+            **kwargs,
         )
-        self.exported_methods |= _Builder.exported_methods
 
-        print("--> Summary")
-        print("# Existing study")
-        if self.builder.existing.empty:
-            print("Nothing yet")
-        else:
-            plan = [p for p in self.builder.existing.plan.unique()]
-            print("%s" % " | ".join(plan))
-
-        print("# Available profiles")
-        for p in ["demand", "hydro", "solar", "wind"]:
-            possible = self.builder.get_base_profile(p)
-            if len(possible) != 0:
-                print("%s: %s" % (p, " | ".join(possible)))
+        self.exported_methods |= self.builder.exported_methods
 
         self._scenario_info["grid_model"] = self.builder.grid_model
         self._scenario_info["interconnect"] = self.builder.interconnect
@@ -189,7 +173,6 @@ class _Builder:
     wind = ""
     engine = "REISE.jl"
     exported_methods = {
-        "set_base_profile",
         "set_engine",
         "set_name",
         "set_time",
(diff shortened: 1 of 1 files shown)


powersimdata/scenario/create.py
```diff
@@ -372,3 +371,118 @@ def get_base_grid(self):
...
+        self.set_base_grid()
+        self.set_change_table()
+
+    def set_base_grid(self):
+        """Set base grid"""
+        raise NotImplementedError()
+
+    def set_change_table(self):
+        """Set change table"""
+        raise NotImplementedError()
```
rouille, line 469: We will have to use functions frm #608 to convert the PyPSA `NetWork` to a `Grid` object and to get the base profiles

merrielle:
This is a test comment. Can someone reply so I can see if I get email notifications now?

powersimdata/scenario/create.py
```diff
@@ -1,5 +1,4 @@
 import copy
-import warnings
 
 import numpy as np
 import pandas as pd
```
merrielle, line 4: This is a test comment. Can someone reply so I can see if I get email notifications now?

danielolsen: Test

rouille: Thanks for posting

merrielle: Notifications are now working! For some reason they were being sent to the wrong email address.

jon-hagg (approved):

powersimdata/scenario/create.py
```diff
@@ -372,3 +371,118 @@ def get_base_grid(self):
...
+    :param str grid model: the grid model
+    :param list interconnect: list of interconnect(s) to build.
+    :param pandas.DataFrame table: scenario list table.
+    :param \\*\\*kwargs: optional parameters used to instantiate a Grid object.
+    """
+
+    def __init__(self, grid_model, interconnect, table, **kwargs):
+        super().__init__(grid_model, interconnect, table)
+
+        self.reduction = None if "reduction" not in kwargs else kwargs["reduction"]
```
merrielle, line 456: Can we document what can be included in kwargs?

merrielle (approved):
I like the class design you went with. I'm following your example with the change table.

powersimdata/scenario/create.py
```diff
@@ -314,35 +318,30 @@ def set_time(self, start_date, end_date, interval):
...
         """Returns available base profiles.
 
         :param str kind: one of *'demand'*, *'hydro'*, *'solar'*, *'wind'*.
         :return: (*list*) -- available version for selected profile kind.
+        :raises NotImplementedError: always - implemented in child classes.
         """
-        return ProfileInput().get_profile_version(self.grid_model, kind)
+        raise NotImplementedError("Implemented in the child classes")
 
     def set_base_profile(self, kind, version):
```
merrielle, line 337: I'm a bit confused. Are these functions supposed to be available in all child classes? Does the _Builder class have to include every function from all its children?

rouille: We have used this pattern for all parent/child classes, define all methods in the parent class, implement the ones common to all child classes and overwrite the ones that are specific to a child class. I don't have a preference and I can remove from the parent class the methods only available in the `FromCSV` child class. What do you think @jon-hagg ?

jon-hagg: Makes sense to me - the parent class can just have all the methods _required_ from child classes but if I'm trying to be object oriented I would say it _shouldn't_ know about methods specific to child classes. E.g. suppose another library extends the pattern with new subclasses, we certainly wouldn't update our code based on those packages.

powersimdata/scenario/create.py
rouille: I have added documentation in the docstring. Let me know what you think.

powersimdata/scenario/create.py
rouille: Done

powersimdata/network/helpers.py
```diff
@@ -60,7 +60,12 @@ def interconnect_to_name(interconnect, model="hifld"):
     :param str model: the grid model.
     :return: (*str*): name of grid model.
     """
-    return "_".join(sorted(check_and_format_interconnect(interconnect, model=model)))
+    return "_".join(
+        sorted(
+            i.replace(" ", "")
+            for i in check_and_format_interconnect(interconnect, model=model)
+        )
+    )
```
BainanXia, line 68: I think [l.40](https://github.com/Breakthrough-Energy/PowerSimData/blob/f652b5b67743003af20e597ee6109c5688570f8f/powersimdata/network/helpers.py#L40) in `check_and_format_interconnect` will take care of the spaces, no?

rouille: Yeah. I forgot to remove it in #631 after adding it in the `check_and_format_interconnect` function, see [comment](https://github.com/Breakthrough-Energy/PowerSimData/pull/631#discussion_r879899038)

rouille: Yeah. I forgot to remove it in #631 after adding it in the `check_and_format_interconnect` function, see [comment](https://github.com/Breakthrough-Energy/PowerSimData/pull/631#discussion_r879899038)

powersimdata/scenario/create.py
```diff
@@ -344,31 +415,52 @@ def set_base_profile(self, kind, version):
...
+
+    def set_change_table(self):
+        """Set change table"""
+        raise NotImplementedError()
+
+
+class BuilderFactory:
+    """Map a grid model to a builder"""
+
+    MAP = {"usa_tamu": FromCSV, "hifld": FromCSV, "europe_tub": FromPyPSA}
```
BainanXia, line 453: I'm trying to understand why we will need a class to handle this static mapping. Do we have plans to deal with more complicated situations here in the future?

rouille: This is usually how this is done, e.g., see the [`config`](https://github.com/Breakthrough-Energy/PowerSimData/blob/develop/powersimdata/utility/config.py) module. I agree that the mapping could be moved in the `get_builder` function. @jenhagg, what is your take on that?

jenhagg: I agree, we don't need a class here, just having the mapping in `get_builder` is fine. I'm not sure we even need the one in the config module. I think it was done that way to have a single place for the string constants but we have e.g. `"usa_tamu"` in lots of places so that wouldn't really accomplish much in this case.

BainanXia (approved):
LGTM. Thanks!

powersimdata/scenario/create.py
```diff
@@ -205,14 +188,35 @@ class _Builder:
...
 
+    def print_existing_study(self):
+        """Print existing study"""
+
+        print("--> Begin: Existing Study")
+        if self.existing.empty:
+            print("Nothing yet")
+        else:
+            plan = [p for p in self.existing.plan.unique()]
+            print("%s" % " | ".join(plan))
```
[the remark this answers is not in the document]
rouille: Done

powersimdata/network/helpers.py
rouille: Done

powersimdata/scenario/create.py
rouille: Done

PR #637, opened 2022-05-20
fix: save correct data to case_storage.mat

jon-hagg:

### Purpose
Fix obvious typo affecting scenarios that include storage in the change table.

### What the code is doing
Save the right content to the mat file.

### Testing
Ran it in docker.

### Time estimate
1 min

### DIFF
--- a/powersimdata/scenario/execute.py
+++ b/powersimdata/scenario/execute.py
@@ -217,7 +217,7 @@ class SimulationInput:
 
         if mpc_storage is not None:
             with self._data_access.write(storage_file_path, save_local=False) as f:
-                savemat(f, mpc, appendmat=False)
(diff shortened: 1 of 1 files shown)


BainanXia (approved):
Ah, good catch.

danielolsen:
Aha! I think I see what's going on here. In REISE.jl, we have a `try`/`catch` around the whole process of loading the file as well as loading some of the data within the file (at certain locations), so we were successfully reading the file but then generating and catching an error when the right information wasn't in the right place, and telling the user that the file itself isn't there: https://github.com/Breakthrough-Energy/REISE.jl/blob/0ab4ad2430b26bf0e2288f4d882beb5b9c2f523b/src/read.jl#L80

PR #639, opened 2022-05-25
fix: pin fs version to avoid regression in latest

jenhagg:
### Purpose
We should probably create a release to fix the bug in scenarios with storage. Since there is an issue with the latest `fs`, we were going to pin the version (it's done in one of the feature branches) but I'm not sure when that will be merged. If it's soon, let me know and we can close this.

### What the code is doing
Pin `fs==2.4.14`. 

### Testing
Ran a scenario using this version of the package, also it's what we've been using, so no actual change here.

### Time estimate
2 min


rouille (approved):
Thanks

rouille:
The feature branch (`ben/eu`) will be merged in another feature branch (`hifld`). I am not sure when `hifld` will be merged into `develop` so it is great to have the fix.

PR #642, opened 2022-05-26
Bump version to v0.5.3 

jenhagg:
### Purpose
Bump the version so we can do a new release. Also forgot to pin the fs version in setup.cfg.

### What the code is doing
See above

### Testing
Ran `tox` in PostREISE after installing this branch.

### Time estimate
3 min


BainanXia (approved):
Thanks

PR #645, opened 2022-06-03
fix: pass grid instead of grid.dcline

jenhagg:
### Purpose
I got an error using `calculate_dcline_difference` since it checks that the parameters are `Grid` instances. We use other grid attributes within that function, so we simply pass the grid object (instead of removing the type check). 

### What the code is doing
Pass full grid object.

### Testing
Manually in a container. I ended up with a different error due to having an empty `grid.dcline` but that is not surprising. 

### Time estimate
2 min


BainanXia (approved):
Make sense.

PR #650, opened 2022-06-08
Reorganize grid related modules and objects

rouille:

### Purpose
Move and rename modules/classes/functions/tests related to grid creation from different sources and engines. The purpose is to make it clear where the data come from. To illustrate, we had a `powersimdata.input.scenario_grid` module for reading the **grid.mat** file created by REISE/REISE.jl. In the future we will likely have CEM scenarios ran with PyPSA and the grid will be converted differently. For this reason the current `scenario_grid` module is renamed `reise_to_grid`.

Please propose other names/location ideas if you have some in mind.

### What the code is doing
No new code. Modules /classes/functions/tests have been renamed and relocated.

### Testing
Run existing tests

### Where to look
In the new `powersimdata.input.converter` and `powersimdata.input.exporter` modules

### Usage Example/Visuals
N/A

### Time estimate
It should be short.


jenhagg (approved):

powersimdata/input/converter/helpers.py
```diff
@@ -0,0 +1,108 @@
+import pandas as pd
+
+
+def add_column_to_data_frame(data_frame, column_dict):
```
FabianHofmann, line 4: Just looking through (and this is really just a stupid side-comment), there is the nice pandas version `assign` for that purpose
```python 
df = df.assign(**column_dict)
``` 
might be helpful for future implementations

rouille: Should I get rid of the `add_column_to_data_frame` and used `assign` when needed?

rouille: Done

merrielle (approved):

PR #653, opened 2022-06-14
fix: add conversion factor for co2 emissions when exporting to pypsa

FabianHofmann:
### Purpose
Fix the unit conversion when exporting co2 emission rates to pypsa networks. Pypsa specifies co2 emissions in tonnes per MWh_thermal, powersimdata in kg per MWh_electric. A new dictionary was added to `plants.py` specifying the type dependent efficiencies. These are necessary for the conversion and are taken from the Danish Energy Agency.   

### Testing
No tests added

### Time estimate
2 min

### DIFF
--- a/powersimdata/network/constants/plants.py
+++ b/powersimdata/network/constants/plants.py
@@ -60,6 +60,16 @@ carbon_per_mwh = {
     "ng": 469,
 }
 
+# MWh_electric to MWh_thermal
+# Source: Danish Energy Agency, "Technology Data - Generation of Energy and District Heating",
+# https://ens.dk/sites/ens.dk/files/Analyser/technology_data_catalogue_for_el_and_dh.pdf
+efficiency = {
+    "coal": 0.33,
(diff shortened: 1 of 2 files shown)


BainanXia (approved):
Thanks!

powersimdata/network/constants/plants.py
```diff
@@ -60,6 +60,16 @@
     "ng": 469,
 }
 
+# MWh_electric to MWh_thermal
```
rouille, line 63: Should it be MWh_termal to MWh_electric? The supply of x MW of heat from burning coal/dfo/ng (input) for every y MW of electricity (output) it produces

FabianHofmann: The ratio `efficiency` given below is `MWh_electric`/`MWh_thermal`. So the electric power output a generator produces for burning 1 MWh of thermal energy. Does that make sense?

rouille (approved):
Looks good

PR #661, opened 2022-07-27
Fix bugs in the Analyze class affecting demand flexibility

rouille:

### Purpose
Fix bugs in `Analyze` class and update docstring of the `OutputData` class

### What the code is doing
Unindent the `get_load_shift_up` and `get_load_shift_dn`methods that were declared within the `get_load_shed` methods

### Testing
Existing unit tests

### Where to look
Changes touch the `Analyze` class. A docstring is updated in the `OutputData` class

### Usage Example/Visuals
N/A

### Time estimate
2 min

### DIFF
--- a/powersimdata/scenario/analyze.py
+++ b/powersimdata/scenario/analyze.py
@@ -202,22 +202,22 @@ class Analyze(Ready):
 
         return load_shed
 
-        def get_load_shift_up(self):
-            """Returns LOAD_SHIFT_UP data frame. This is the amount that flexible demand
-            deviates above (e.g., recovers) the base demand.
+    def get_load_shift_up(self):
+        """Returns LOAD_SHIFT_UP data frame. This is the amount that flexible demand
+        deviates above (e.g., recovers) the base demand.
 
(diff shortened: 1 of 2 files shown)


BainanXia (approved):
Thanks

powersimdata/scenario/analyze.py
```diff
@@ -187,7 +187,7 @@ def get_load_shed(self):
         try:
             # It's either on the server or in our local ScenarioData folder
             load_shed = self._get_data("LOAD_SHED")
-        except OSError:
+        except ValueError:
```
jenhagg, line 190: I think we would still get an `OSError`from [here](https://github.com/Breakthrough-Energy/PowerSimData/blob/fc4da59642f6c175a4faf061b29b3e821ae48ab9/powersimdata/data_access/data_access.py#L119) if the file doesn't exist on the server (or blob storage, or locally), but based on the slack discussion it sounds like that shouldn't be the case.

rouille: Done. I dropped the commit where the change happened

jenhagg (approved):

PR #666, opened 2022-08-05
Enable loading any scenario from blob storage

jenhagg:
### Purpose
Switch to the new esmi blob storage account which is populated with existing scenario data. This is represented as one of a few filesystems which are combined into a single `MultiFS` which acts as a "remote" data store. 

### What the code is doing
* Point to the new storage account, keeping priority the same. We still use the server as the source of truth, but if a user isn't connected to the vpn when the multi filesystem is instantiated, we do not repeatedly attempt to connect (no different than before, just adding context)
* Create smaller multi-filesystems depending on context. E.g. when we are using `InputData`, there is no need to get profiles, so there is no need to add the profiles container as a possible location to check. This is done by passing the `make_fs` parameter in `context.py.`
* Add a read only [shared access signature](https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview) to `server_setup.py`. This is defined (as you can kind of tell by the query params) to allow read and list access on blobs and containers only (not queues, tables, etc). The reason for this is that it turns out anonymous directory listing is not supported for gen2 HNS accounts (as opposed to the original blob storage accounts). See the [issue](https://github.com/Azure/azure-sdk-for-python/issues/25476) I submitted with azure for more context. The workaround suggested there does work, but it's just as safe to use a SAS instead and is more likely to be stable over time, as it's an officially supported use case.

### Testing
Ran a scenario in docker container, after deleting existing volumes and containers. Also loaded a scenario that did not exist locally yet and checked that the files are cached locally.

### Usage Example/Visuals
Loading a scenario from blob storage.
```
In [1]: from powersimdata import Scenario

In [2]: scl = Scenario().get_scenario_table()
/Users/jenhagg/.local/share/virtualenvs/PowerSimData-9jRUr7-X/lib/python3.9/site-packages/paramiko/transport.py:236: CryptographyDeprecationWarning: Blowfish has been deprecated
  "class": algorithms.Blowfish,
Could not connect to ssh server
Initialized remote filesystem with scenario_fs
Transferring ScenarioList.csv.2 from scenario_fs

In [3]: scl.head().index
Out[3]: Int64Index([398, 400, 403, 404, 405], dtype='int64', name='id')

In [4]: s404 = Scenario(404)
Could not connect to ssh server
Initialized remote filesystem with scenario_fs
Transferring ScenarioList.csv.2 from scenario_fs
Transferring ExecuteList.csv.2 from scenario_fs
SCENARIO: test | EasternBase_2020_two_day_test

--> State
analyze
Could not connect to ssh server
Initialized remote filesystem with scenario_fs
--> Loading grid
Could not connect to ssh server
Initialized remote filesystem with profile_fs,scenario_fs
Transferring ScenarioList.csv.2 from scenario_fs
Loading bus
Loading plant
Loading heat_rate_curve
Loading gencost_before
Loading gencost_after
Loading branch
Loading dcline
Loading sub
Loading bus2sub
--> Loading ct
```
At this point, if I `cd ScenarioData` and `ls -la data/output | grep 404` there are no results.

So we can load some outputs

```
In [5]: avg_cong = s404.get_averaged_cong()
Could not connect to ssh server
Initialized remote filesystem with scenario_fs
--> Loading AVERAGED_CONG
data/output/404_AVERAGED_CONG.pkl not found on local machine
Transferring 404_AVERAGED_CONG.pkl from scenario_fs

In [6]: dcline_pf = s404.get_dcline_pf()
Could not connect to ssh server
Initialized remote filesystem with scenario_fs
--> Loading PF_DCLINE
data/output/404_PF_DCLINE.pkl not found on local machine
Transferring 404_PF_DCLINE.pkl from scenario_fs
```
And confirm that the files have been downloaded

```
❯ ls data/output | grep 404
404_AVERAGED_CONG.pkl
404_PF_DCLINE.pkl
~/ScenarioData ❯                                                                                      

```

Note that this example would have worked essentially the same way before, but only for a handful of scenarios, instead of the full set of existing ones.

### Time estimate
20 min - I don't think this breaks anything or is super risky

### DIFF
--- a/powersimdata/data_access/data_access.py
+++ b/powersimdata/data_access/data_access.py
@@ -16,9 +16,8 @@ from powersimdata.utility import server_setup
 class DataAccess:
     """Interface to a local or remote data store."""
 
-    def __init__(self, root):
+    def __init__(self):
         """Constructor"""
-        self.root = root
         self.join = fs.path.join
         self.local_fs = None
 
@@ -163,10 +162,10 @@ class DataAccess:
 class LocalDataAccess(DataAccess):
     """Interface to shared data volume"""
 
-    def __init__(self, root=server_setup.LOCAL_DIR):
-        super().__init__(root)
-        self.local_fs = fs.open_fs(root)
-        self.fs = self._get_fs()
+    def __init__(self, _fs=None):
+        super().__init__()
+        self.local_fs = fs.open_fs(server_setup.LOCAL_DIR)
+        self.fs = _fs if _fs is not None else self._get_fs()
 
     def _get_fs(self):
         mfs = MultiFS()
@@ -193,22 +192,12 @@ class LocalDataAccess(DataAccess):
 class SSHDataAccess(DataAccess):
     """Interface to a remote data store, accessed via SSH."""
 
-    def __init__(self, root=server_setup.DATA_ROOT_DIR):
+    def __init__(self, _fs=None):
         """Constructor"""
-        super().__init__(root)
-        self._fs = None
+        super().__init__()
+        self.root = server_setup.DATA_ROOT_DIR
+        self.fs = _fs if _fs is not None else get_multi_fs(self.root)
         self.local_fs = fs.open_fs(server_setup.LOCAL_DIR)
-
-    @property
-    def fs(self):
-        """Get or create the filesystem object
-
-        :raises IOError: if connection failed or still within retry window
-        :return: (*fs.multifs.MultiFS*) -- filesystem instance
-        """
-        if self._fs is None:
-            self._fs = get_multi_fs(self.root)
-        return self._fs
 
     def exec_command(self, command):
         ssh_fs = self.fs.get_fs("ssh_fs")
@@ -285,7 +274,7 @@ class _DataAccessTemplate(SSHDataAccess):
 
     def __init__(self, fs_url):
         self.local_fs = fs.open_fs(fs_url)
-        self._fs = self._get_fs(fs_url)
+        self.fs = self._get_fs(fs_url)
         self.root = "foo"
         self.join = fs.path.join
 
--- a/powersimdata/data_access/data_access.py
+++ b/powersimdata/data_access/data_access.py
@@ -196,8 +196,14 @@ class SSHDataAccess(DataAccess):
         """Constructor"""
         super().__init__()
         self.root = server_setup.DATA_ROOT_DIR
-        self.fs = _fs if _fs is not None else get_multi_fs(self.root)
+        self._fs = _fs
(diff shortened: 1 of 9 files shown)


powersimdata/data_access/data_access.py
```diff
@@ -193,19 +192,15 @@ def push(self, file_name, checksum):
...
+        self._fs = _fs
         self.local_fs = fs.open_fs(server_setup.LOCAL_DIR)
 
     @property
     def fs(self):
-        """Get or create the filesystem object
-
-        :raises IOError: if connection failed or still within retry window
-        :return: (*fs.multifs.MultiFS*) -- filesystem instance
-        """
```
BainanXia, line 208: Wonder why we remove these doc strings?

powersimdata/utility/server_setup.py
```diff
@@ -14,6 +14,7 @@
 MODEL_DIR = config.MODEL_DIR
 ENGINE_DIR = config.ENGINE_DIR
 DEPLOYMENT_MODE = get_deployment_mode()
+BLOB_TOKEN_RO = "?sv=2021-06-08&ss=b&srt=co&sp=rl&se=2050-08-06T01:31:08Z&st=2022-08-05T17:31:08Z&spr=https&sig=ORHiRQQCocyaHXV2phhSN92GFhRnaHuGOecskxsmG3U%3D"
```
BainanXia, line 17: This is a permanent token, right?

powersimdata/data_access/fs_helper.py
```diff
@@ -49,3 +50,22 @@ def get_multi_fs(root):
     remotes = ",".join([f[0] for f in mfs.iterate_fs()])
     print(f"Initialized remote filesystem with {remotes}")
     return mfs
+
+
+def get_scenario_fs():
+    """Create filesystem combining the server (if connected) with blob storage,
+        prioritizing the server if connected.
```
BainanXia, line 57: Redundant indent here.

BainanXia (approved):
This is nice. Thanks!

rouille (approved):
Tested it and it works well. I have, as you do, this annoying `CryptographyDeprecationWarning`. I will investigate

powersimdata/utility/server_setup.py
jenhagg: It expires in 2050, so essentially permanent.

powersimdata/data_access/data_access.py
jenhagg: I initially removed the property but put it back later. Looking at it again, the old docstring is not accurate, so will add a new one.

PR #672, opened 2022-09-01
ci: update action versions

jenhagg:
### Purpose
I noticed there were new versions of the checkout and setup-python actions, so might as well upgrade.

### What the code is doing
Nothing, as far as I can tell, other than using a newer version of node.js on the action runners. There might be new features we could use, but I haven't looked closely.

### Testing
The checks still pass.

### Time estimate
1 min

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -11,10 +11,10 @@ jobs:
 
     name: Generate coverage report
     steps:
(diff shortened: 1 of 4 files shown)


rouille (approved):
🙏

BainanXia (approved):

dmuldrew (approved):
🚢

PR #673, opened 2022-09-01
Perf improvements for tests and minor fixes

jenhagg:
### Purpose
Reduce test duration by reusing objects that don't really need to be recreated. 

### What the code is doing
* Create fewer instances of `OutputData`, `InputData`, and `ProfileInput`.
* Use texas grid for demand flexibility test instead of usa
* use `None` as default argument
* skip test that downloads from zenodo 

### Testing
Running the tests locally shows a small but not totally insignificant reduction in test duration. 

### Time estimate
2 min

### DIFF
--- a/powersimdata/input/change_table.py
+++ b/powersimdata/input/change_table.py
@@ -16,6 +16,7 @@ from powersimdata.input.changes import (
     remove_plant,
     scale_plant_pmin,
 )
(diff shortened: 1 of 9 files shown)


rouille (approved):
It can only help

BainanXia (approved):

rouille (approved):

PR #676, opened 2022-09-06
fix: update import statement

jenhagg:

### Purpose
Import from the correct module. This might succeed at runtime due to the function already having been imported, but that's not reliable. 

### Time estimate
10 sec


rouille (approved):
Thanks

BainanXia (approved):

PR #677, opened 2022-09-07
docs: add packages to requirements.txt, to fix import errors in sphinx

jenhagg:
### Purpose
Generating docs for a given module requires importing it, so we need to install any packages it uses at the top level. E.g. the [docs for europe_tub](https://breakthrough-energy.github.io/docs/powersimdata.network.europe_tub.html#module-powersimdata.network.europe_tub) is missing for this reason.

### What the code is doing
Add packages to requirements.txt and fix one place where the newline causes docs to not render properly ([this one](https://breakthrough-energy.github.io/docs/powersimdata.input.exporter.html#powersimdata.input.exporter.export_to_pypsa.export_to_pypsa))

### Testing
Built the docs locally.

### Time estimate
1 min

### DIFF
--- a/powersimdata/input/exporter/export_to_pypsa.py
+++ b/powersimdata/input/exporter/export_to_pypsa.py
@@ -144,7 +144,7 @@ def export_to_pypsa(
     .. note::
         This function does not export storages yet.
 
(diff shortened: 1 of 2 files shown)


BainanXia (approved):
Thanks

PR #678, opened 2022-09-07
Re-enable roundtrip conversion

FabianHofmann:
As discussed on slack

### DIFF
--- a/powersimdata/input/const/casemat_const.py
+++ b/powersimdata/input/const/casemat_const.py
@@ -46,7 +46,6 @@ col_type_branch = [
 ]
 
 col_name_bus = [
-    "bus_id",
     "type",
     "Pd",
     "Qd",
--- a/powersimdata/input/converter/pypsa_to_grid.py
+++ b/powersimdata/input/converter/pypsa_to_grid.py
@@ -165,27 +165,28 @@ class FromPyPSA(AbstractGrid):
             data_loc = "pypsa"
 
         # Read in data from PyPSA
-        bus_in_pypsa = n.buses
-        sub_in_pypsa = pd.DataFrame()
-        bus2sub_in_pypsa = pd.DataFrame()
+        bus_pypsa = n.buses
+        sub_pypsa = pd.DataFrame()
+        bus2sub_pypsa = pd.DataFrame()
         gencost_cols = ["start_up_cost", "shut_down_cost", "marginal_cost"]
-        gencost_in_pypsa = n.generators[gencost_cols]
-        plant_in_pypsa = n.generators.drop(gencost_cols, axis=1)
-        lines_in_pypsa = n.lines
-        transformers_in_pypsa = n.transformers
-        branch_in_pypsa = pd.concat(
-            [lines_in_pypsa, transformers_in_pypsa],
+        gencost_pypsa = n.generators[gencost_cols]
+        plant_pypsa = n.generators.drop(gencost_cols, axis=1)
+        lines_pypsa = n.lines
+        transformers_pypsa = n.transformers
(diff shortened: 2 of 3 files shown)


powersimdata/input/const/casemat_const.py
```diff
@@ -46,7 +46,6 @@
 ]
 
 col_name_bus = [
-    "bus_id",
```
rouille, line 49: I expect droping `bus_id` here will cause some problems since MATPOWER and REISE.jl expects a `bus_id` columns in the `bus` table (see [Table B-1](https://matpower.org/docs/MATPOWER-manual.pdf)). The problem is not writing the input data (`case.mat` file) since a `Grid` object does not have the `bus_id` column and it is added in the `export_case_mat` function [here](https://github.com/Breakthrough-Energy/PowerSimData/blob/1955c6f6a048797472a91e99f57c74f42bcc380d/powersimdata/input/exporter/export_to_reise.py#L36) but it is a problem when we read the output grid from the engine in the `reise_to_grid` module [here](https://github.com/Breakthrough-Energy/PowerSimData/blob/1955c6f6a048797472a91e99f57c74f42bcc380d/powersimdata/input/converter/reise_to_grid.py#L62) and, more specifically [there](https://github.com/Breakthrough-Energy/PowerSimData/blob/1955c6f6a048797472a91e99f57c74f42bcc380d/powersimdata/input/converter/reise_to_grid.py#L183)

@BainanXia and @jenhagg, what is your opinion? Note that we are currently moving away from this serializer/deserializer (case.mat/grid.mat).

rouille: We probably need to remove `bus_id` from the `grid_const` module and use these sets of constants instead. I think this makes sense since as mentioned above the `bus` data frame of a `Grid` object does not have a `bus_id` column.

BainanXia: > We probably need to remove `bus_id` from the `grid_const` module and keep it in the `casemat_const`. I think this makes sense since as mentioned above the `bus` data frame of a `Grid` object does not have a `bus_id` column.

I agree.

rouille: The idea would be to add [here](https://github.com/Breakthrough-Energy/PowerSimData/blob/f0e7641947e6e50389536524bd22954fc78eaf76/powersimdata/input/const/grid_const.py#L47):
```
col_name_bus.pop(0)
col_name_type.pop(0)
```

jenhagg: Makes sense, and did some local testing which worked as expected. I was curious about the reason we need to handle bus index/cols "different" than similar dataframes, but I _think_ this is a result of having to match the matpower format, so that's a separate issue.

rouille: @FabianHofmann, do you want to take care of that or should I do it after it is merged into `ben/grideq`?

FabianHofmann: I'd prefer you take over after merging since it's not super clear to me which engines expect which keys. But just a side note: I'd rather use a 
```
if 'bus_id' in col_name_bus:
   col_name_bus.remove("bus_id") 
```
pattern, which makes sure one does not delete the wrong keys.

rouille: Sounds good. Let me know when you are done with your changes and I will go over before approving it.

FabianHofmann: Thanks, I think that the PR can be merged then. The tests all pass and everything should be fixed as far as I see.

rouille (approved):
Thanks

PR #641, opened 2022-05-26
DRAFT: add data structure for expansion candidates

danielolsen:
[Pull Request doc](https://breakthrough-energy.github.io/docs/user/git_guide.html#d-pull-request)

### Purpose
Add a thin data structure enclosing expansion candidate information for generation, transmission, and storage.

### What the code is doing
Basic column checking on `branch`, `plant`, `storage` parameters if they're passed, otherwise instantiating an empty dataframe with the appropriate columns.

### Usage Example/Visuals
```python
>>> from powersimdata.input.expansion_candidates import ExpansionCandidates
>>> import pandas as pd
>>> ec = ExpansionCandidates()
>>> ec.branch
Empty DataFrame
Columns: [from_bus, to_bus, inv_cost, max_capacity]
Index: []
>>> ec.plant
Empty DataFrame
Columns: [bus_id, type, marginal_cost, inv_cost, max_capacity]
Index: []
>>> ec.storage
Empty DataFrame
Columns: [bus_id, inv_cost_power, inv_cost_energy, max_capacity_energy, max_capacity_power]
Index: []
>>> plant = pd.DataFrame({"bus_id": [1, 10, 10, 100], "type": ["ng", "ng", "wind", "solar"], "marginal_cost": [40, 40, 0, 0]})
>>> ec = ExpansionCandidates(plant=plant)
>>> ec.branch
Empty DataFrame
Columns: [from_bus, to_bus, inv_cost, max_capacity]
Index: []
>>> ec.plant
   bus_id   type  marginal_cost  inv_cost  max_capacity
0       1     ng             40       NaN           NaN
1      10     ng             40       NaN           NaN
2      10   wind              0       NaN           NaN
3     100  solar              0       NaN           NaN
>>> ec.storage
Empty DataFrame
Columns: [bus_id, inv_cost_power, inv_cost_energy, max_capacity_energy, max_capacity_power]
Index: []
```

### Time estimate
?

PR #679, opened 2022-09-09
add data structure for expansion candidates

jenhagg:
### Purpose
Define a new data structure for expansion candidates and some preliminary validation based on the current grid. The schema is the same as in https://github.com/Breakthrough-Energy/PowerSimData/pull/641 but I refactored it a bit.

### What the code is doing
Add the new class, which can have candidates for branches, plants, and storage. Each one has a set of allowed columns and required columns. When assigning a given type of candidate, there are some basic checks - e.g. buses are in the grid. I imagine more checks could be done in subsequent PRs. 

We've discussed expanding the costs over multiple years to account for inflation - this isn't accounted for yet, and I suspect it might be done during the preparation phase, after a user has defined the investment years/time span. 

In general there is probably a lot from the original epic that isn't implemented yet. E.g. defining branch candidates by substation instead of buses. This is just meant to be a usable subset of the possible/desired schema.

### Testing
New unit tests

### Usage Example/Visuals
Create an instance:
```
from powersimdata import Grid
from powersimdata.input.expansion_candidates import ExpansionCandidates
grid = Grid("Texas")
ec = ExpansionCandidates(grid)
```
After assigning some values (using the example from the unit test), we can inspect it using the custom `repr`
```
In [10]: ec
Out[10]:
ExpansionCandidates(branch: cols=['from_bus', 'to_bus', 'inv_cost', 'max_capacity'], rows=5
plant: cols=['bus_id', 'type', 'marginal_cost', 'inv_cost', 'max_capacity'], rows=5
storage: cols=['bus_id', 'inv_cost_power', 'inv_cost_energy', 'max_capacity_energy', 'max_capacity_power'], rows=0)
```
Or look at the data frames directly (similar for storage). Note the indexes are not required, that's an artifact of using the existing grid as an example:
```
In [11]: ec.plant
Out[11]:
           bus_id   type  marginal_cost  inv_cost  max_capacity
plant_id
12875     3001004   wind              4       NaN           NaN
12876     3001006   wind              4       NaN           NaN
12877     3001009   wind              4       NaN           NaN
12878     3001011  solar              4       NaN           NaN
12879     3001021   wind              4       NaN           NaN

In [12]: ec.branch
Out[12]:
           from_bus   to_bus  inv_cost  max_capacity
branch_id
100915      3001001  3001064       NaN           NaN
100916      3001001  3001064       NaN           NaN
100917      3001001  3001071       NaN           NaN
100918      3001001  3001071       NaN           NaN
100919      3001002  3001007       NaN           NaN
```

### Time estimate
30 min

### DIFF
--- a/powersimdata/input/expansion_candidates.py
+++ b/powersimdata/input/expansion_candidates.py
@@ -63,6 +63,15 @@ def check_storage(storage, grid):
     check_bus_id(storage.bus_id, grid)
 
 
+def _validate_df(name, df):
+    assert name in ("branch", "plant", "storage")
+    if not isinstance(df, pd.DataFrame):
+        raise TypeError(f"{name} must be a data frame")
+    if not set(_required_columns[name]) <= set(df.columns):
+        raise TypeError(f"{name} must have columns {_required_columns[name]}")
+    return df.reindex(_columns[name], axis=1)
+
+
 _columns = {
     "branch": ("from_bus", "to_bus", "inv_cost", "max_capacity"),
     "plant": ("bus_id", "type", "marginal_cost", "inv_cost", "max_capacity"),
@@ -110,20 +119,12 @@ class ExpansionCandidates:
         storage = _short_description(self.storage, "storage")
         return f"{name}({branch}\n{plant}\n{storage})"
 
-    def _validate_df(self, name, df):
-        assert name in ("branch", "plant", "storage")
-        if not isinstance(df, pd.DataFrame):
-            raise TypeError(f"{name} must be a data frame")
-        if not set(_required_columns[name]) <= set(df.columns):
-            raise TypeError(f"{name} must have columns {_required_columns[name]}")
-        return df.reindex(_columns[name], axis=1)
-
     def set_branch(self, branch):
         """Validate and assign branch candidates
 
         :param pd.DataFrame branch: branch dataframe
         """
-        branch = self._validate_df("branch", branch)
+        branch = _validate_df("branch", branch)
         check_branch(branch, self.grid)
         self.branch = branch
 
@@ -132,7 +133,7 @@ class ExpansionCandidates:
 
         :param pd.DataFrame plant: plant dataframe
         """
-        plant = self._validate_df("plant", plant)
+        plant = _validate_df("plant", plant)
         check_plant(plant, self.grid)
         self.plant = plant
 
@@ -141,6 +142,6 @@ class ExpansionCandidates:
 
         :param pd.DataFrame storage: storage dataframe
         """
-        storage = self._validate_df("storage", storage)
+        storage = _validate_df("storage", storage)
         check_storage(storage, self.grid)
         self.storage = storage
--- a/powersimdata/input/expansion_candidates.py
+++ b/powersimdata/input/expansion_candidates.py
@@ -99,6 +99,17 @@ class ExpansionCandidates:
         self.storage = pd.DataFrame(columns=_columns["storage"])
         self.grid = grid
 
+    def __repr__(self):
+        name = self.__class__.__name__
+
+        def _short_description(df, name):
+            return f"{name}: cols={list(df.columns)}, rows={df.shape[0]}"
+
+        branch = _short_description(self.branch, "branch")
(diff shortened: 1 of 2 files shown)


powersimdata/input/expansion_candidates.py
```diff
@@ -0,0 +1,146 @@
...
+    check_bus_id(plant.bus_id, grid)
+
+
+def check_storage(storage, grid):
+    """Check storage expansion candidates relative to a specific grid
+
+    :param pd.DataFrame storage: dataframe of storage candidates
+    :param powersimdata.input.grid.Grid grid: reference grid
+    """
+    check_bus_id(storage.bus_id, grid)
```
rouille, line 63: It seems that all these functions could be used in a PCM workflow and hence could live in the `powersimdata.input.check` module. To illustrate, we could use these functions when adding new storage, branches, plants and storage in the change table.

jenhagg: I had the same thought but didn't want to expand the scope. The bus id check could be easy to refactor. I just noticed we have a function `_check_line_voltages` already but that is used to check the consistency of an existing grid, so we would have to refactor it (or add a separate function) to compare new branches with a given bus dataframe. I'm not seeing a check for that in the change table - is that something we should have? If it's missing, that might be worth adding, but I was planning to hold off on refactoring until there is a bit more cem progress.

rouille: Sounds good.

I believe we check that the bus(es) exist when adding a new line, a plant or a storage in the change table. You are right, we don't check the voltages. It could be added.

powersimdata/input/expansion_candidates.py
```diff
@@ -0,0 +1,146 @@
...
+        storage = _short_description(self.storage, "storage")
+        return f"{name}({branch}\n{plant}\n{storage})"
+
+    def _validate_df(self, name, df):
+        assert name in ("branch", "plant", "storage")
+        if not isinstance(df, pd.DataFrame):
+            raise TypeError(f"{name} must be a data frame")
+        if not set(_required_columns[name]) <= set(df.columns):
+            raise TypeError(f"{name} must have columns {_required_columns[name]}")
+        return df.reindex(_columns[name], axis=1)
```
rouille, line 119: It seems that it could be defined outside the class

powersimdata/input/expansion_candidates.py
```diff
@@ -0,0 +1,146 @@
...
+    def __repr__(self):
+        name = self.__class__.__name__
+
+        def _short_description(df, name):
+            return f"{name}: cols={list(df.columns)}, rows={df.shape[0]}"
+
+        branch = _short_description(self.branch, "branch")
+        plant = _short_description(self.plant, "plant")
+        storage = _short_description(self.storage, "storage")
+        return f"{name}({branch}\n{plant}\n{storage})"
```
rouille, line 111: This is neat

powersimdata/input/expansion_candidates.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+    check_bus_id(storage.bus_id, grid)
+
+
+def _validate_df(name, df):
+    assert name in ("branch", "plant", "storage")
+    if not isinstance(df, pd.DataFrame):
+        raise TypeError(f"{name} must be a data frame")
+    if not set(_required_columns[name]) <= set(df.columns):
+        raise TypeError(f"{name} must have columns {_required_columns[name]}")
```
rouille, line 71: Should the error raised be a `ValueError`?

rouille (approved):
Looks good

BainanXia (approved):
Recall our discussion of data structures to incorporate multiple investment years in CEM. We will either do `ec_all[year] = ec` or having columns for each year in candidate data frames. With this implementation, I presume we lean towards the former.

jenhagg:
> Recall our discussion of data structures to incorporate multiple investment years in CEM. We will either do `ec_all[year] = ec` or having columns for each year in candidate data frames. With this implementation, I presume we lean towards the former.

Yeah I hadn't figured that part out yet, but I think adding the columns makes sense, and usually prefer not to duplicate data.

PR #689, opened 2022-09-29
Fix roundtrip PyPSA -> PSD -> PyPSA

FabianHofmann:
This fixes the conversion for the roundtrip PyPSA -> PSD -> PyPSA. 

Test it with 

```python
from powersimdata.input.converter.pypsa_to_grid import FromPyPSA
from powersimdata.network.europe_tub.model import TUB
from powersimdata.input.exporter.export_to_pypsa import export_to_pypsa

tub = TUB("Europe", reduction=1024)
tub.build()

g = FromPyPSA(tub.network)

n = export_to_pypsa(g)

for c in n.iterate_components():
    if c.name in ['Carrier']:
        continue
    orig = tub.df(c.name)
    conv = c.df

    assert_frame_equal(conv[orig.columns].head(), orig.head())

```

Missing points:

* The distinction between Stores and StorageUnits is not yet completed
* The carrier roundtrip conversion is not complete as it depends on internal static data on PSD side.

### DIFF
--- a/powersimdata/input/exporter/export_to_pypsa.py
+++ b/powersimdata/input/exporter/export_to_pypsa.py
@@ -7,6 +7,16 @@ from powersimdata.scenario.scenario import Scenario
 from powersimdata.utility.helpers import _check_import
 
 
+def restore_original_columns(df, overwrite=[]):
+    prefix = "pypsa_"
+    for col in df.columns[df.columns.str.startswith(prefix)]:
+        target = col[len(prefix) :]
+        fallback = df.pop(col)
+        if target not in df or target in overwrite:
+            df[target] = fallback
+    return df
+
+
 def export_to_pypsa(
     scenario_or_grid,
     add_all_columns=False,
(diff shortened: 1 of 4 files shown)


powersimdata/input/exporter/export_to_pypsa.py
```diff
@@ -7,6 +7,16 @@
 from powersimdata.utility.helpers import _check_import
 
 
+def restore_original_columns(df, overwrite=[]):
```
BainanXia, line 10: This is minor: we might what to avoid immutable defaults for kwargs, i.e. `overwrite=None` with the logic in the function `if not overwrite: overwrite = []`.

BainanXia (approved):
Looks good.

powersimdata/input/exporter/export_to_pypsa.py
rouille: It will be taken care of after it is merged

rouille (approved):
Thanks

Issue #668, opened 2022-08-15
Delegate responsibility to engine to create grid data for simulation


rouille:
# :rocket:

- [ ] Is your feature request essential for your project?

### Describe the workflow you want to enable
Move the writing of the transformed grid into the binary file read by the engine into the wrapper. Different engine need different data to run. To illustrate, REISE.jl can now handle demand flexibility while MATPOWER can't. Keeping the current MAT-file structure to run CEM scenarios with PyPSA in the future will also  be extremely challenging.

### Describe your proposed implementation, if applicable
* Pickle the transformed grid in the `SimulationInput` class.
* Send the binary file to the **tmp** folder on the server/container along with the profiles.
* Give to each wrapper the responsibility to load the transformed grid enclosed in the pickle file and operate on the object to create the adequate input for the engine. The current `export_case_mat` function located in the `powersimdata.input.exporter.export_to_reise` module can be moved in the **REISE** and **REISE.jl** repositories. The extra-column giving the demand flexibility located in the `bus` data frame of a `Grid` object can be added to the `bus` table of the **case.mat** file for REISE.jl.
* Convert the **grid.mat** into a pickle file in each wrapper after a run (applicable to REISE and REISE.jl engine). Classes and functions enclosed in the `powersimdata.input.converter.reise_to_grid` module can be moved in each wrapper.
* Fetch the `id_grid.pkl` in the`Ready` class via the `get_grid` method.

The wrappers will likely depend on **PowerSimData**. For example the class responsible for converting the **grid.mat** file created after the simulation (`REISEConverter`) inherits from the `AbstractGrid`. We will have to add **PowerSimData** as a dependency in the wrappers.

### Describe alternatives you've considered, if relevant
We spent a considerable amount of time trying to figure out a way to use the same input files for different engines and there is no easy solution. Moving away from MAT-file is also very appealing.

### Additional context
This is a multi-repository feature. We will also need to convert all existing scenario's **grid.mat** file on the server/blob. Note that this can easily be done with the current version of the code by: 
* Looping over the scenario list, using the current `get_grid` method;
* Dumping the `Grid` object in a pickle file;
* Uploading the files on the server/blob

rouille:
Addressed in #701

PR #690, opened 2022-09-29
Replace case.mat with grid.pkl

jenhagg:

### Purpose
Pass a grid.pkl file as input to REISE.jl instead of case.mat. Partially addresses #668 along with https://github.com/Breakthrough-Energy/REISE.jl/pull/185

### What the code is doing
Remove `export_case_mat` which is now in REISE.jl temporarily. Preserve index names in transform_grid.py so operations on it after unpickling work as expected.

### Testing
Ran a scenario in a container.

### Time estimate
10 min

### DIFF
--- a/powersimdata/scenario/execute.py
+++ b/powersimdata/scenario/execute.py
@@ -1,8 +1,8 @@
+import pickle
+
 import pandas as pd
-from scipy.io import savemat
 
 from powersimdata.data_access.context import Context
-from powersimdata.input.exporter.export_to_reise import export_case_mat
 from powersimdata.input.grid import Grid
 from powersimdata.input.input_data import InputData
 from powersimdata.input.transform_grid import TransformGrid
@@ -107,7 +107,7 @@ class Execute(Ready):
             for p in ["demand", "hydro", "solar", "wind"]:
                 si.prepare_profile(p, profiles_as)
 
-            si.prepare_mpc_file()
+            si.prepare_grid()
 
             if "demand_flexibility" in self.ct:
                 # Prepare all specified demand flexibility profiles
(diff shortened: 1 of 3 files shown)


rouille (approved):
Looks good

powersimdata/scenario/execute.py
```diff
@@ -204,20 +203,12 @@ def __init__(self, data_access, scenario_info, grid, ct):
...
-        with self._data_access.write(file_path, save_local=False) as f:
-            savemat(f, mpc, appendmat=False)
-
-        if mpc_storage is not None:
-            with self._data_access.write(storage_file_path, save_local=False) as f:
-                savemat(f, mpc_storage, appendmat=False)
+    def prepare_grid(self):
+        """Prepare grid for simulation."""
+        print("--> Preparing grid data")
+        dest_path = "/".join([self.REL_TMP_DIR, "grid.pkl"])
```
BainanXia, line 209: Not sure whether we should use `os.join` here or it is taken care of in `_data_access` already.

powersimdata/scenario/execute.py
```diff
@@ -258,11 +246,8 @@ def prepare_demand_flexibility_parameters_file(self):
...
             else 0
         ]
-
-        # Convert the dictionary to a DataFrame to more easily save as a .csv file
         params = pd.DataFrame.from_dict(params)
 
-        # Save the parameters as a .csv file and move the file to the correct location
-        file_path = "/".join([self.REL_TMP_DIR, "demand_flexibility_parameters.csv"])
-        with self._data_access.write(file_path, save_local=False) as f:
+        dest_path = "/".join([self.REL_TMP_DIR, "demand_flexibility_parameters.csv"])
```
lanesmith, line 251: Similar comment to @BainanXia's above.

lanesmith (approved):
Looks good to me. Thanks!

powersimdata/scenario/execute.py
rouille: The same syntax is used in `prepare_profile`. I believe it is taken care of in `DataAccess`. @jenhagg will know for sure.

jenhagg: Right - we use `"/".join` here since we're dealing with a pyfilesystem path that is os independent.

BainanXia (approved):

Issue #687, opened 2022-09-27
Fetch usa_tamu data from zenodo 


rouille:
# :rocket:

- [ ] Is your feature request essential for your project?

### Describe the workflow you want to enable
Similarly to what is done for PyPSA-Eur (see the `TUB` class enclosed in the `powersimdata.network.europe_tub.model` module), the usa TAMU csv files should be downloaded from Zenodo to efficiently track the grid model version. 

### Describe your proposed implementation, if applicable
* A new column in the ScenarioList file would specify the grid model version
* A new parameter in the constructor of the  `Grid` class would allow to access different version. A new attribute would enclose the grid model version.

### Additional context
Every time, we update the model (`ramp_30`, `Pmax`, etc), a new release should be created on Zenodo.

PR #697, opened 2022-11-01
Create a Zenodo download manager

rouille:

### Purpose
Build the `Zenodo` class around the [Zenodo Rest API](https://developers.zenodo.org/) to efficiently handle data coming from different record. Partially addresses #687.

### What the code is doing
* downloads data of a user specified record
* calculate checksum of previously downloaded data and erase then download new copy if necessary
* print information about record (title, version, etc)

### Testing
Manual testing.

### Where to look
* write `Zenodo` class in `powersimdata.network.zenodo` module
* update the `TUB` class located in the `powersimdata.network.europe_tub.model` module
* remove obsolete `zenodo_get` package from **Pipfile** and **requirements.txt**

### Usage Example/Visuals
```
>>> from powersimdata.network.zenodo import Zenodo
>>> z = Zenodo("3601881")
Title: PyPSA-Eur: An Open Optimisation Model of the European Transmission System (Dataset)
Publication date: 2022-09-20
Version: v0.6.1
DOI: 10.5281/zenodo.7251657
>>> z.load_data("powersimdata/network/europe_tub")
100% [....................................................................] 1784815481 / 1784815481
networks.zip (1702.1 MB)
>>> z.load_data("powersimdata/network/europe_tub")
networks.zip has been downloaded previously
>>> z = Zenodo("7251657")
Title: PyPSA-Eur: An Open Optimisation Model of the European Transmission System (Dataset)
Publication date: 2022-09-20
Version: v0.6.1
DOI: 10.5281/zenodo.7251657
>>> z.load_data("powersimdata/network/europe_tub")
networks.zip has been downloaded previously
```

We can now do that:
```
>>> tub = TUB("Europe", zenodo_record_id="latest", reduction=128)
Title: PyPSA-Eur: An Open Optimisation Model of the European Transmission System (Dataset)
Publication date: 2022-09-20
Version: v0.6.1
DOI: 10.5281/zenodo.7251657
networks.zip has been downloaded previously
>>> tub.build()
INFO:pypsa.io:Imported network elec_s_128_ec.nc has buses, carriers, generators, lines, links, loads, storage_units, stores
>>> tub = TUB("Europe", reduction=128)
Title: PyPSA-Eur: An Open Optimisation Model of the European Transmission System (Dataset)
Publication date: 2022-09-20
Version: v0.6.1
DOI: 10.5281/zenodo.7251657
networks.zip has been downloaded previously
>>> tub.build()
INFO:pypsa.io:Imported network elec_s_128_ec.nc has buses, carriers, generators, lines, links, loads, storage_units, stores
```

### Time estimate
30min

### DIFF
--- a/powersimdata/network/europe_tub/model.py
+++ b/powersimdata/network/europe_tub/model.py
@@ -1,6 +1,4 @@
 import os
-import shutil
-from zipfile import ZipFile
 
 from powersimdata.network.constants.region.geography import get_geography
 from powersimdata.network.constants.region.zones import from_pypsa
@@ -9,59 +7,67 @@ from powersimdata.network.helpers import (
     interconnect_to_name,
 )
 from powersimdata.network.model import ModelImmutables
+from powersimdata.network.zenodo import Zenodo
 from powersimdata.utility.helpers import _check_import
 
 pypsa = _check_import("pypsa")
-zenodo_get = _check_import("zenodo_get")
 
 
 class TUB:
     """PyPSA Europe network.
 
     :param str/iterable interconnect: interconnect name(s).
+    :param str zenodo_record_id: the zenodo record id. If set to None, v0.6.1 will
+        be used. If set to latest, the latest version will be used.
     :param int reduction: reduction parameter (number of nodes in network). If None,
         the full network is loaded.
-    :param bool overwrite: the existing dataset is deleted and a new dataset is
-        downloaded from zenodo.
     """
 
-    def __init__(self, interconnect, reduction=None, overwrite=False):
+    def __init__(self, interconnect, zenodo_record_id=None, reduction=None):
         """Constructor."""
         self.grid_model = "europe_tub"
         self.interconnect = check_and_format_interconnect(
             interconnect, model=self.grid_model
         )
-        self.data_loc = os.path.join(os.path.dirname(__file__), "data")
-        self.zenodo_record_id = "3601881"
-        self.reduction = reduction
 
-        if overwrite:
-            self.remove_data()
+        if zenodo_record_id is None:
+            z = Zenodo("7251657")
+        elif zenodo_record_id == "latest":
+            z = Zenodo("3601881")
+        else:
+            z = Zenodo(zenodo_record_id)
 
-        self.retrieve_data()
+        z.load_data(os.path.dirname(__file__))
+        self.data_loc = os.path.join(z.dir, "networks")
+        self._set_reduction(reduction)
 
-    def remove_data(self):
-        """Remove data stored on disk"""
-        print("Removing PyPSA-Eur dataset")
-        shutil.rmtree(self.data_loc)
+    def _set_reduction(self, reduction):
+        """Validate and set reduction parameter
 
-    def retrieve_data(self):
-        """Fetch data"""
-        zenodo_get.zenodo_get([self.zenodo_record_id, "-o", f"{self.data_loc}"])
-        with ZipFile(os.path.join(self.data_loc, "networks.zip"), "r") as zip_network:
-            zip_network.extractall(self.data_loc)
+        :raises ValueError: if ``reduction`` is not available.
+        """
+        if reduction is None:
+            self.reduction = None
+        else:
+            available = [
+                s
+                for f in os.listdir(self.data_loc)
+                for s in f.split("_")
+                if s.isdigit()
+            ]
+            if str(reduction) in available:
+                self.reduction = reduction
+            else:
+                raise ValueError(f"Available reduced network: {' | '.join(available)}")
 
     def build(self):
         """Build network"""
-        path = os.path.join(self.data_loc, "networks", "elec_s")
+        path = os.path.join(self.data_loc, "elec_s")
         if self.reduction is None:
             network = pypsa.Network(path + ".nc")
-        elif os.path.exists(path + f"_{self.reduction}_ec.nc"):
-            network = pypsa.Network(path + f"_{self.reduction}_ec.nc")
         else:
-            raise ValueError(
-                "Invalid Resolution. Choose among: None | 1024 | 512 | 256 | 128 | 37"
-            )
+            network = pypsa.Network(path + f"_{self.reduction}_ec.nc")
+
         id2zone = {i: l for i, l in enumerate(network.buses.index)}
         zone2id = {l: i for i, l in id2zone.items()}
 
--- a/powersimdata/network/zenodo.py
+++ b/powersimdata/network/zenodo.py
@@ -1,0 +1,144 @@
+import hashlib
+import json
+import os
+import shutil
+from contextlib import contextmanager
+from zipfile import ZipFile
+
+import requests
+from tqdm import tqdm
+
+url = "https://zenodo.org/api/records/"
+
+
+class Zenodo:
+    """Get data from a Zenodo archive.
+
+    :param str record_id: zenodo record id
+    """
+
+    def __init__(self, record_id):
+        """Constructor"""
+        self.record_id = record_id
+        self.content = self._get_record_content()
+
+    def _get_record_content(self):
+        """Make HTTP request to zenodo API and retrieve content.
+
+        :return: (*dict*) -- content of the response in unicode.
+        :raises Exception: if connection times out.
+        :raises ValueError: if record is invalid.
+        """
+        try:
+            r = requests.get(url + self.record_id, timeout=10)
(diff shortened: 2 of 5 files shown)


powersimdata/network/zenodo.py
```diff
@@ -0,0 +1,134 @@
...
+        :raises ValueError: if record is invalid.
+        """
+        try:
+            r = requests.get(url + self.record_id, timeout=10)
+        except requests.exceptions.ConnectTimeout:
+            raise Exception("Connection to zenodo.org timed out")
+
+        if not r.ok:
+            raise ValueError("Record could not be accessed")
+        else:
```
jenhagg, line 39: No need for the `else` (raising error would exit the function).

powersimdata/network/zenodo.py
```diff
@@ -0,0 +1,134 @@
...
+        :raises Exception: if connection times out.
+        :raises ValueError: if record is invalid.
+        """
+        try:
+            r = requests.get(url + self.record_id, timeout=10)
+        except requests.exceptions.ConnectTimeout:
+            raise Exception("Connection to zenodo.org timed out")
+
+        if not r.ok:
+            raise ValueError("Record could not be accessed")
```
jenhagg, line 38: Not sure what the error cases are, but we could include `r.status_code` in the message.

jenhagg:
We might want to update `Pipfile.lock`

powersimdata/network/zenodo.py
```diff
@@ -0,0 +1,134 @@
...
+            h = hashlib.new(f["checksum"].split(":")[0])
+            with open(filename, "rb") as file:
+                bytes = file.read()
+            h.update(bytes)
+            return h.hexdigest()
+
+    @contextmanager
+    def _change_dir(self):
+        work_dir = os.getcwd()
+        os.chdir(os.path.expanduser(self.dir))
```
BainanXia, line 76: TIL!

rouille:
> We might want to update `Pipfile.lock`. Also, looks like `wget` is only installed as part of `zenodo_get`, so we should add that explicitly (otherwise it will be removed when the lock file is regenerated). I also noticed the `wget` package hasn't been updated since 2015, but.. if it works, that's probably fine.

Would you recommend using `urllib` instead?

powersimdata/network/zenodo.py
```diff
@@ -0,0 +1,134 @@
...
+            yield
+        finally:
+            os.chdir(work_dir)
+
+    def _download_data(self, f):
+        """Fetch data.
+
+        :param dict f: information on the file to download.
+        """
+        with self._change_dir():
```
BainanXia, line 87: Given `delete` ->`download` -> `unzip` are called sequentially, maybe we could put the decorator `with self._change_dir()` outside to call it once in `load_data`.

BainanXia:
> > We might want to update `Pipfile.lock`. Also, looks like `wget` is only installed as part of `zenodo_get`, so we should add that explicitly (otherwise it will be removed when the lock file is regenerated). I also noticed the `wget` package hasn't been updated since 2015, but.. if it works, that's probably fine.
> 
> Would you recommend using `urllib` instead?

I'm about to say the same thing since it's in STL.

jenhagg:
> > We might want to update `Pipfile.lock`. Also, looks like `wget` is only installed as part of `zenodo_get`, so we should add that explicitly (otherwise it will be removed when the lock file is regenerated). I also noticed the `wget` package hasn't been updated since 2015, but.. if it works, that's probably fine.
> 
> Would you recommend using `urllib` instead?

We can do this with requests too:
```
def _wget(url, filename, size=None):
    with requests.get(url, stream=True) as r:
        r.raise_for_status()
        if size is None:
            size = r.headers.get("Content-Length")
        with open(filename, "wb") as f:
            with tqdm(
                unit="B",
                unit_scale=True,
                unit_divisor=1024,
                miniters=1,
                total=size,
            ) as pbar:
                for chunk in r.iter_content(chunk_size=8192):
                    f.write(chunk)
                    pbar.update(len(chunk))
```

powersimdata/network/europe_tub/model.py
```diff
@@ -9,59 +7,67 @@
...
-        self.data_loc = os.path.join(os.path.dirname(__file__), "data")
-        self.zenodo_record_id = "3601881"
-        self.reduction = reduction
 
-        if overwrite:
-            self.remove_data()
+        if zenodo_record_id is None:
+            z = Zenodo("7251657")
+        elif zenodo_record_id == "latest":
+            z = Zenodo("3601881")
```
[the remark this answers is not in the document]
jenhagg: IMO it's cleaner to have constants within the module unless they are used elsewhere or change based on the environment (e.g. config options).

powersimdata/network/zenodo.py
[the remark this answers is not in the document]
rouille: We can return the status code as @jenhagg suggested. It will give the user a little more context

powersimdata/network/zenodo.py
rouille: yeah

powersimdata/network/europe_tub/model.py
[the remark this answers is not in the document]
rouille: I agree with @jenhagg.

rouille:
> > > We might want to update `Pipfile.lock`. Also, looks like `wget` is only installed as part of `zenodo_get`, so we should add that explicitly (otherwise it will be removed when the lock file is regenerated). I also noticed the `wget` package hasn't been updated since 2015, but.. if it works, that's probably fine.
> > 
> > 
> > Would you recommend using `urllib` instead?
> 
> We can do this with requests too:
> 
> ```
> def _wget(url, filename, size=None):
>     with requests.get(url, stream=True) as r:
>         r.raise_for_status()
>         if size is None:
>             size = r.headers.get("Content-Length")
>         with open(filename, "wb") as f:
>             with tqdm(
>                 unit="B",
>                 unit_scale=True,
>                 unit_divisor=1024,
>                 miniters=1,
>                 total=size,
>             ) as pbar:
>                 for chunk in r.iter_content(chunk_size=8192):
>                     f.write(chunk)
>                     pbar.update(len(chunk))
> ```

Done

powersimdata/network/zenodo.py
rouille: Done

powersimdata/network/zenodo.py
rouille: Done

powersimdata/network/zenodo.py
```diff
@@ -0,0 +1,134 @@
...
+
+        :param dict f: dictionary containing information on the remote copy of a file.
+        :return: (*str*) -- checksum
+        """
+        return f["checksum"].split(":")[1]
+
+    def _get_local_checksum(self, f):
+        """Get remote copy of a file.
+
+        :param dict f: dictionary containing information on the remote copy of a file.
```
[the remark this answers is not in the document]
rouille: Done

powersimdata/network/zenodo.py
rouille: Done

powersimdata/network/europe_tub/model.py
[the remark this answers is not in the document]
BainanXia: Sure

BainanXia (approved):

powersimdata/network/zenodo.py
```diff
@@ -0,0 +1,144 @@
...
+        """Make HTTP request to zenodo API and retrieve content.
+
+        :return: (*dict*) -- content of the response in unicode.
+        :raises Exception: if connection times out.
+        :raises ValueError: if record is invalid.
+        """
+        try:
+            r = requests.get(url + self.record_id, timeout=10)
+        except requests.exceptions.ConnectTimeout:
+            raise Exception("Connection to zenodo.org timed out")
```
jenhagg, line 35: We could raise `ConnectionError` which is built-in

rouille: Done

rouille:
> We might want to update `Pipfile.lock`. Also, looks like `wget` is only installed as part of `zenodo_get`, so we should add that explicitly (otherwise it will be removed when the lock file is regenerated). I also noticed the `wget` package hasn't been updated since 2015, but.. if it works, that's probably fine.

The update of the `Pipfile.lock` will be taken care of in a separate PR as some tests fail due to an updated version of `pandas`

jenhagg (approved):
Nice

PR #698, opened 2022-11-02
Update pipenv lock file

rouille:

### Purpose
Update **Pipfile.lock** file and fix error and warnings raised by new version of `pandas`

### What the code is doing
N/A

### Testing
Existing unit tests

### Where to look
* handle a `FutureWarning` raised by `pandas` (FutureWarning: The default value of numeric_only in DataFrameGroupBy.mean is deprecated. In a future version, numeric_only will default to False. Either specify numeric_only or select only columns which should be valid for the function) in the `investment_costs` module
* Use `list` in place of `set` in the `mock_profile_input` module. That was generating several an error in several tests.

### Usage Example/Visuals
N/A

### Time estimate
5min

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "bceba796a42dc729b6102499f9d52bf4c4ecc501db76051d720de3b0f42a8639"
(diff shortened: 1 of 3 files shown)


BainanXia (approved):
Thanks

jenhagg (approved):

Issue #680, opened 2022-09-15
Make ramp rate input flexible to other grids


lanesmith:
# :beetle:

- [X] I have checked that this issue has not already been reported.

### Bug summary
Currently, generator ramp rates provided in a grid's `plant.csv` file are overridden in `REISE.jl`'s `prepare.jl` file, regardless of grid type, to equal a value according to a generator's capacity and fuel type (see [here](https://github.com/Breakthrough-Energy/REISE.jl/blob/develop/src/prepare.jl#L25-#L45) for the referenced code). This was originally implemented as a means to improve the ramp rate specifications for the generators in the TAMU U.S.A. grid. However, since we are trying to make the ecosystem more flexible to other grid types, we should make this TAMU-grid-specific adjustment only apply to the TAMU U.S.A. grid. Additionally, this adjustment probably is better suited for inclusion in `PowerSimData` than it is in `REISE.jl`.

yixingxu:
- Agree with @lanesmith and @BainanXia . Ramp rate overwrite should be moved out of `REISE.jl` and included in `PowerSimData` . It should be only used for TAMU U.S.A grid. 
- A ramp rate analysis of the most recent TAMU USA grid could inform us whether these data are reasonable. If they are, the ramp rate overwrite logic should not be applied to future TAMU USA grid scenario studies. The ramp rate overwrite logic is still needed to replicate previous scenarios. 
- When the capacities of generators are scaled up or down, their ramp rate (in MW unit) should also be scaled up or down. This is a new logic needed, once the ramp rate overwrite logic is removed.

lanesmith:
> * When the capacities of generators are scaled up or down, their ramp rate (in MW unit) should also be scaled up or down. This is a new logic needed, once the ramp rate overwrite logic is removed.

I think this point is especially important. Thank you for bringing it up!

[...]

PR #701, opened 2022-11-03
Port linearize_gencost and pmin modifications from REISE.jl

jenhagg:
### Purpose
To facilitate using a grid.pkl and remove the need for mapping between a case.mat/grid.mat/etc, we move grid modifications previously done in REISE.jl to the scenario preparation process. 

This PR should be merged at the same time as https://github.com/Breakthrough-Energy/REISE.jl/pull/187, after which we should update the existing grid.mat files. I believe we also need to update the ramp30 values in plant.csv as discussed in this issue https://github.com/Breakthrough-Energy/PowerSimData/issues/680 and release a new dataset on zenodo.

### What the code is doing
Scale plant Pmin values and linearize cost curves the same way we were doing in REISE.jl - after we apply the change table, and just before we upload the grid.pkl to the server (or, container, or wherever). I also ported the linearize_gencost tests from julia since those will be unused/deleted.

### Testing
Ran a scenario in a container and reloaded it after it finished for some cursory spot checking.

### Time estimate
20 min

### DIFF
--- a/powersimdata/design/generation/tests/test_cost_curves.py
+++ b/powersimdata/design/generation/tests/test_cost_curves.py
@@ -198,6 +198,11 @@ expected_two_segment.f2 = [
 ]
 expected_two_segment["interconnect"] = expected_two_segment.pop("interconnect")
 
+expected_all_equal = mock_grid_gc.gencost["before"].copy()
+expected_all_equal.c2 = 0
+expected_all_equal.c1 = 0
+expected_all_equal.c0 = [2707, 20508, 68409]
+
 
 def test_linearize_gencost():
     actual = linearize_gencost(mock_grid_gc)
@@ -207,6 +212,14 @@ def test_linearize_gencost():
 def test_linearize_gencost_two_segment():
     actual = linearize_gencost(mock_grid_gc, num_segments=2)
     assert_frame_equal(expected_two_segment, actual, check_dtype=False)
+
+
+def test_linearize_gencost_pmin_equal_pmax():
+    plant = mock_grid_gc.plant.copy()
+    plant.Pmin = plant.Pmax
+    grid = MockGrid({"plant": plant.reset_index().to_dict(), "gencost_before": mock_gc})
+    actual = linearize_gencost(grid, num_segments=3)
+    assert_frame_equal(expected_all_equal, actual, check_dtype=False)
 
 
 def test_get_supply_data():
--- a/powersimdata/design/generation/tests/test_cost_curves.py
+++ b/powersimdata/design/generation/tests/test_cost_curves.py
@@ -2,10 +2,10 @@ import pandas as pd
 from pandas.testing import assert_frame_equal, assert_series_equal
 
 from powersimdata.design.generation.cost_curves import (
+    _linearize_gencost,
     build_supply_curve,
     get_supply_data,
     ks_test,
-    linearize_gencost,
     lower_bound_index,
 )
 from powersimdata.tests.mock_grid import MockGrid
@@ -205,12 +205,12 @@ expected_all_equal.c0 = [2707, 20508, 68409]
 
 
 def test_linearize_gencost():
-    actual = linearize_gencost(mock_grid_gc)
+    actual = _linearize_gencost(mock_grid_gc)
     assert_frame_equal(expected_one_segment, actual, check_dtype=False)
 
 
 def test_linearize_gencost_two_segment():
-    actual = linearize_gencost(mock_grid_gc, num_segments=2)
+    actual = _linearize_gencost(mock_grid_gc, num_segments=2)
     assert_frame_equal(expected_two_segment, actual, check_dtype=False)
 
 
@@ -218,7 +218,7 @@ def test_linearize_gencost_pmin_equal_pmax():
     plant = mock_grid_gc.plant.copy()
     plant.Pmin = plant.Pmax
     grid = MockGrid({"plant": plant.reset_index().to_dict(), "gencost_before": mock_gc})
-    actual = linearize_gencost(grid, num_segments=3)
+    actual = _linearize_gencost(grid, num_segments=3)
     assert_frame_equal(expected_all_equal, actual, check_dtype=False)
 
 
--- a/powersimdata/design/generation/tests/test_cost_curves.py
+++ b/powersimdata/design/generation/tests/test_cost_curves.py
@@ -1,10 +1,11 @@
 import pandas as pd
-from pandas.testing import assert_series_equal
+from pandas.testing import assert_frame_equal, assert_series_equal
 
 from powersimdata.design.generation.cost_curves import (
     build_supply_curve,
     get_supply_data,
     ks_test,
+    linearize_gencost,
     lower_bound_index,
 )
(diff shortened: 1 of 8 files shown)


rouille:
Is **powersimdata/design/generation/** a good location for the `cost_curves` module? Should we move it in **powersimdata/scenario/** or **powersimdata/input/**? What do you think @BainanXia?

powersimdata/scenario/execute.py
```diff
@@ -203,9 +207,30 @@ def __init__(self, data_access, scenario_info, grid, ct):
...
+        plant.loc[plant.type.isin(profile_resource), "Pmin"] = 0
+        plant.loc[plant.status == 0, "Pmin"] = 0
+
     def prepare_grid(self):
         """Prepare grid for simulation."""
         print("--> Preparing grid data")
+        self._adjust_pmin()
+        storage = self.grid.storage
+        self.grid.gencost["after"] = linearize_gencost(self.grid)
+        storage["gencost"] = _linearize_gencost(storage["gencost"], storage["gen"])
```
rouille, line 232: I am not sure I understand why we need two functions. Could we do:
```
storage["gencost"] = _linearize_gencost(storage["gencost"], storage["gen"])
self.grid.gencost["after"] = _linearize_gencost(elf.grid.gencost["before"], self.grid.plant)
```
Then we could discard the `linearize_gencost` function in the `cost_curves` module and rename `_linearize_gencost` to `linearize_gencost`

BainanXia:
> Is **powersimdata/design/generation/** a good location for the `cost_curves` module? Should we move it in **powersimdata/scenario/** or **powersimdata/input/**? What do you think @BainanXia?

I agree. The original idea of `design` module is a place for functions that are optionally used in the scenario workflow. However, linearization is essential given the engine can't deal with quadratic cost curves (at least for now) and similarly for the ramp module in the future. Thus we can create a new module in `powersimdata/scenario/`. Not sure what is a good name for it, maybe `configure`?

powersimdata/scenario/execute.py
```diff
@@ -203,9 +207,30 @@ def __init__(self, data_access, scenario_info, grid, ct):
...
+        def _scale(x):
+            factor = pmin_factor.get(x.type)
+            return x.Pmin if factor is None else factor * x.Pmax
+
+        plant.Pmin = plant.apply(_scale, axis=1)
+
+        # set Pmin to 0 for generators that are off or profile based
+        profile_resource = list(mi.plants["profile_resources"])
+        plant.loc[plant.type.isin(profile_resource), "Pmin"] = 0
+        plant.loc[plant.status == 0, "Pmin"] = 0
```
BainanXia, line 224: Maybe out of the scope of this PR, two high level questions:
1. `hydro` is one of `profile_resources`, the `Pmin` of which will be set 0 at the end, however the `pmin_factor` of `hydro` is grid dependent .
2. We will need to build user interface to pass user specified `pmin_as_share_of_pmax` (either plant based or plant type based or plant type * zone based) in the future, right?

powersimdata/design/generation/tests/test_cost_curves.py
```diff
@@ -141,6 +142,85 @@
...
+    actual = linearize_gencost(mock_grid_gc, num_segments=2)
+    assert_frame_equal(expected_two_segment, actual, check_dtype=False)
+
+
+def test_linearize_gencost_pmin_equal_pmax():
+    plant = mock_grid_gc.plant.copy()
+    plant.Pmin = plant.Pmax
+    grid = MockGrid({"plant": plant.reset_index().to_dict(), "gencost_before": mock_gc})
+    actual = linearize_gencost(grid, num_segments=3)
+    assert_frame_equal(expected_all_equal, actual, check_dtype=False)
```
BainanXia, line 222: Thanks! We thought about adding such tests a while ago as in the very first version of implementation, dividing by zero wasn't avoided.

powersimdata/scenario/execute.py
BainanXia: 1. True. Just pointing out such entries are modified twice in the current process.
2. Agreed.

lanesmith:
> I agree. The original idea of `design` module is a place for functions that are optionally used in the scenario workflow. However, linearization is essential given the engine can't deal with quadratic cost curves (at least for now) and similarly for the ramp module in the future. Thus we can create a new module in `powersimdata/scenario/`. Not sure what is a good name for it, maybe `configure`?

Is the ramp module essential? Especially once (if?) `plant.csv` gets updated, I'm not sure we would need the ramp-rate correction anymore.

BainanXia:
> > I agree. The original idea of `design` module is a place for functions that are optionally used in the scenario workflow. However, linearization is essential given the engine can't deal with quadratic cost curves (at least for now) and similarly for the ramp module in the future. Thus we can create a new module in `powersimdata/scenario/`. Not sure what is a good name for it, maybe `configure`?
> 
> Is the ramp module essential? Especially once (if?) `plant.csv` gets updated, I'm not sure we would need the ramp-rate correction anymore.

We will need that when scaling existing plants as well as adding new plants without specifying ramp rates.

powersimdata/scenario/execute.py
jenhagg: done

rouille:
> > Is **powersimdata/design/generation/** a good location for the `cost_curves` module? Should we move it in **powersimdata/scenario/** or **powersimdata/input/**? What do you think @BainanXia?
> 
> I agree. The original idea of `design` module is a place for functions that are optionally used in the scenario workflow. However, linearization is essential given the engine can't deal with quadratic cost curves (at least for now) and similarly for the ramp module in the future. Thus we can create a new module in `powersimdata/scenario/`. Not sure what is a good name for it, maybe `configure`?

It can be done in a separate PR. I am fine with having the `linearize_gencost` function in a new `powersimdata.scenario.configure`. We could have the ramp rate setting there too. Or we could have a `linearize_cost_curve` module and a `assign_ramp_rate` modules in `powersimdata.input`. Hat is your opinion @jenhagg?

jenhagg:
> > > Is **powersimdata/design/generation/** a good location for the `cost_curves` module? Should we move it in **powersimdata/scenario/** or **powersimdata/input/**? What do you think @BainanXia?
> > 
> > 
> > I agree. The original idea of `design` module is a place for functions that are optionally used in the scenario workflow. However, linearization is essential given the engine can't deal with quadratic cost curves (at least for now) and similarly for the ramp module in the future. Thus we can create a new module in `powersimdata/scenario/`. Not sure what is a good name for it, maybe `configure`?
> 
> It can be done in a separate PR. I am fine with having the `linearize_gencost` function in a new `powersimdata.scenario.configure`. We could have the ramp rate setting there too. Or we could have a `linearize_cost_curve` module and a `assign_ramp_rate` modules in `powersimdata.input`. Hat is your opinion @jenhagg?

Based on the current package structure, I think having those in `powersimdata.input` makes sense, although the distinction between input and design isn't super clear to me. I'm fine with either option; there is always the possibility of reorganizing in the future.

lanesmith (approved):
Looks good to me

rouille (approved):
Thank you for tackling this!

BainanXia (approved):
Nice 🚀

PR #702, opened 2022-11-03
Generalize area type in check function

rouille:

### Purpose
Generalize area type type in check function so we can check for the european model that a given list of countries are in the grid.

### What the code is doing
Replace hard coded `state` by the `grid.model_immutables.zones["diivision"]` string that gives the division type (state or country so far) for a grid model

### Testing
Existing unit tests

### Where to look
Look at the `_check_areas_are_in_grid_and_format` functions. Other modifications are docstring improvements

### Usage Example/Visuals
N/A

### Time estimate
2min

### DIFF
--- a/powersimdata/input/check.py
+++ b/powersimdata/input/check.py
@@ -300,11 +300,11 @@ def _check_grid_type(grid):
 
 
 def _check_areas_and_format(areas, mi=None):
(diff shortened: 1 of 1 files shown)


jenhagg (approved):

BainanXia (approved):
Thanks

PR #704, opened 2022-11-14
chore: remove unused code for reading grid from matfile

jenhagg:
### Purpose
Remove code that is not used anymore.

### What the code is doing
Remove the `reise_to_grid` module and related tests since we now load the grid from a `.pkl` file once a scenario is in the analyze state.

### Testing
n/a - testing was done on the new code that replaces this

### Time estimate
5 min


rouille:
We probably want to remove the casemat_cont module too

rouille (approved):
Thanks for cleaning

PR #709, opened 2022-12-07
fix: set bus_id index name

jenhagg:
### Purpose
Fix error when we add buses, e.g. via the demo_western.py script in plug. Without the index name, the grid.pkl used in call.py cannot be exported to csv. This should address the issue from the community slack.

### What the code is doing
Add the index name for bus data frame. We already had this for the other ones but since it was tested with the demo_texas script, which doesn't transform buses, the error didn't occur.

### Testing
Ran a simulation using the western demo and didn't have the error anymore.

### Time estimate
2 min


BainanXia (approved):
Good catch. Thanks!

PR #716, opened 2022-12-20
fix: preserve inflow suffix in pypsa to grid mapping

jenhagg:
### Purpose
Fix a typo which results in the `grid.gencost` data frames having a different size than `grid.plant`, which results in an error when trying to run `add_interconnect_to_grid_data_frames`. 

### What the code is doing
Use the same variable we started with instead of reassigning it. 

### Testing
Created a grid and observed the matching dimensions. Also checked that I could add the interconnect column as described above.

### Time estimate
5 min


BainanXia (approved):
Good catch!

rouille (approved):
🙏

PR #719, opened 2023-01-03
fix: storage gencost columns

jenhagg:
### Purpose
Preserve the assignment of new columns (reassign, since that's not done in place), and remove unnecessary code.

### What the code is doing
See above.

### Testing
Manual testing, and verified that the change fixes an issue I encountered (error about number of columns in the gencost data frame, during scenario preparation).

### Time estimate
2 min


BainanXia (approved):
Good catch

PR #722, opened 2023-01-09
Support for europe_tub scenarios

jenhagg:

### Purpose
Make it possible to run a simulation with the europe grid. Note, there is more work needed to support the full network (not just the reduced network).

### What the code is doing
* Extract and upload profiles if necessary
* Add interconnect and zone columns to grid data frames
* handle possible string columns in plant profile
* other fixes as needed

### Testing
launched a scenario (in a container) and extracted results. 

### Time estimate
20 min

### DIFF
--- a/powersimdata/data_access/fs_helper.py
+++ b/powersimdata/data_access/fs_helper.py
@@ -1,8 +1,13 @@
+import logging
+
 import fs
 from fs.multifs import MultiFS
 
 from powersimdata.data_access.ssh_fs import WrapSSHFS
 from powersimdata.utility import server_setup
+
+logger = logging.getLogger("azure.core.pipeline.policies.http_logging_policy")
+logger.setLevel(logging.WARNING)
 
 
 def get_blob_fs(container):
@@ -12,7 +17,7 @@ def get_blob_fs(container):
     :return: (*fs.base.FS*) -- filesystem instance
     """
     account = "esmi"
-    sas_token = server_setup.BLOB_TOKEN_RO
+    sas_token = server_setup.get_blob_credential()
     return fs.open_fs(f"azblobv2://{account}:{sas_token}@{container}")
 
 
--- a/powersimdata/input/profile_input.py
+++ b/powersimdata/input/profile_input.py
@@ -98,6 +98,13 @@ class ProfileInput(InputBase):
         return self.data_access.get_profile_version(_callback)
 
     def upload(self, grid_model, name, profile):
+        """Upload the given profile to blob storage and local cache
+
+        :param str grid_model: the grid model
+        :param str name: the file name for the profile, without extension
+        :param pandas.DataFrame profile: profile data frame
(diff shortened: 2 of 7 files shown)


powersimdata/input/profile_input.py
```diff
@@ -35,13 +38,20 @@ def get_profile_version(_fs, grid_model, kind):
     return [f.replace(f"{kind}_", "").replace(".csv", "") for f in matching]
 
 
+def _make_fs():
+    mfs = MultiFS()
+    writeable = server_setup.BLOB_TOKEN_RW is not None
+    mfs.add_fs("profile_fs", get_blob_fs("profiles"), write=writeable)
+    return mfs
```
jenhagg, line 45: using `MultiFS` wrapper since there is code in `data_access.py` which relies on this in some cases

powersimdata/data_access/fs_helper.py
```diff
@@ -1,9 +1,14 @@
+import logging
+
 import fs
 from fs.multifs import MultiFS
 
 from powersimdata.data_access.ssh_fs import WrapSSHFS
 from powersimdata.utility import server_setup
 
+logger = logging.getLogger("azure.core.pipeline.policies.http_logging_policy")
+logger.setLevel(logging.WARNING)
```
rouille, line 10: What `logger.setLevel(logging.WARNING)` does? Just being curious.

jenhagg: It filters out a bunch of output from the azure sdk, which I guess are at the `INFO` level (as compared to `WARNING`). So not actually helpful to a user, and honestly I don't know why they would have that output by default.. probably unintentional.

powersimdata/network/europe_tub/model.py
```diff
@@ -110,35 +115,52 @@ def from_zenodo(self, zenodo_record_id, reduction):
...
+        reduction.
+
+        :return: (*str*) -- the version which is used to construct a file name
+        """
+        append = f"_{self.reduction}" if self.reduction is not None else ""
+        return f"{self.version}" + append
+
+    def _profile_exists(self, kind):
+        """Check if a profile has been uploaded for the given version
+
```
rouille, line 156: Missing entry for `kind`

powersimdata/input/profile_input.py
```diff
@@ -68,8 +78,10 @@ def _read(self, f, path):
         data = pd.read_csv(f, index_col=0, parse_dates=True)
         if "demand_flexibility" in path:
             data.columns = data.columns.astype(str)
-        else:
+        elif all(c.isdigit() for c in data.columns):
             data.columns = data.columns.astype(int)
+        else:
+            data.columns = data.columns.astype(str)
```
rouille, line 84: I am surprised that `pandas` is not able to assign the correct type to column's header

rouille (approved):
:pray:

PR #724, opened 2023-01-14
ci: prevent duplicate workflows

jenhagg:

### Purpose
Prevent duplicate workflows after opening PRs. Currently we run one for `push` and `pull_request` events, which although technically different (since PRs use the pending merge commit vs branch head), that's not relevant as far as I know. We compare the base and head repo to ensure PRs from forks still run the checks (since those won't trigger a `push` event).

### What the code is doing
Add condition to jobs which should prevent duplicates. Copy pasted from https://github.com/zopefoundation/meta/issues/145

Also removed the `skip_ci` feature since this is kinda builtin now: see https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs

### Testing
Some testing here, although that doesn't test the forked repo scenario: https://github.com/jenhagg/test-ci/pull/6

### Time estimate
5 min

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   formatting:
-    if: "!contains(github.event.head_commit.message, 'skip_ci')"
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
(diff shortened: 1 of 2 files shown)


rouille (approved):
🙏. I believe this is something we want to do in all repositories.

BainanXia (approved):
TIL!

PR #725, opened 2023-01-24
fix: include kwargs in cache key to prevent false collisions

jenhagg:
### Purpose
When we cache the grid, the key should include reduction for europe_tub networks. Otherwise all reductions will be grouped together, and reloading the grid with a different reduction will result in the initial (wrong) grid.

### What the code is doing
Add kwargs to cache key  construction, encoded as a tuple to avoid changing/adding logic. 

### Testing
Loaded multiple grids with different reductions.

### Time estimate
5 min


rouille (approved):
Good catch

BainanXia (approved):

PR #726, opened 2023-02-04
Fix storage conversion so pcm will be feasible

jenhagg:
### Purpose
Fix the infeasibility we encounter when adding storage.

### What the code is doing
Set the values during `pypsa_to_grid` conversion in a way that is consistent with `TransformGrid`. This fixes the underlying issue of initial storage (presumably) violating the JuMP constraint.

Also applied latest black formatting.

### Testing
Ran a 2 month pcm with the custom 128 node network, and got storage results files.

### Time estimate
10 min

### DIFF
--- a/powersimdata/input/converter/pypsa_to_grid.py
+++ b/powersimdata/input/converter/pypsa_to_grid.py
@@ -52,7 +52,7 @@ def _get_storage_storagedata(df, storage_type):
         storage_storagedata["OutEff"] = 1
         storage_storagedata["InEff"] = 1
     else:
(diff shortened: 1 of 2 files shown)


rouille:
@FabianHofmann this PR fixes the infeasibility we encountered

rouille (approved):
Thanks!

BainanXia (approved):
🎉

FabianHofmann:
Ah great!! Congrats, awesome debugging! Seemed to be quite difficult to find.

PR #728, opened 2023-02-08
chore: update packages, apply latest black formatting

jenhagg:

### Purpose
Fix the formatting check by applying the latest black. Regenerate Pipfile.lock which includes an update to `exceptiongroup` rather than removal, which is what was done by dependabot for some reason.

### What the code is doing
See above

### Testing
Ran tox locally

### Time estimate
3 min

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
         fail-fast: false
         matrix:
-          python-version: [3.8, 3.9, '3.10']
+          python-version: [3.9, '3.10', '3.11']
 
     name: Python ${{ matrix.python-version }}
     steps:
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -23,27 +23,27 @@
         },
         "azure-core": {
             "hashes": [
-                "sha256:223b0e90cbdd1f03c41b195b03239899843f20d00964dbb85e64386873414a2d",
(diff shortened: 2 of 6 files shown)


BainanXia (approved):
🙏

Wait: it seems some tests are failing.

jenhagg:
> 🙏
> 
> Wait: it seems some tests are failing.

Yeah it seems `xarray` is updated to a version that requires python 3.9. We might have to pin that version by adding it to the Pipfile, since I don't think we can specify a minimum python version (only exact). 

edit: alternatively, we could drop 3.8 from the checks? I think someone could still install the packages in a 3.8 environment since they aren't specified exactly in setup.cfg or requirements.txt, only in Pipfile.lock. That might, however, require a user running tox locally (thus installing from Pipfile.lock) to be using `>=3.9`.

rouille:
> > 🙏
> 
> > 
> 
> > Wait: it seems some tests are failing.
> 
> 
> 
> Yeah it seems `xarray` is updated to a version that requires python 3.9. We might have to pin that version by adding it to the Pipfile, since I don't think we can specify a minimum python version (only exact). 
> 
> 
> 
> edit: alternatively, we could drop 3.8 from the checks? I think someone could still install the packages in a 3.8 environment since they aren't specified exactly in setup.cfg or requirements.txt, only in Pipfile.lock. That might, however, require a user running tox locally (thus installing from Pipfile.lock) to be using `>=3.9`. 

Yes we can drop 3.8 and add 3.11. A final update before saying 👋

rouille (approved):
Thanks

